PR #410, opened 2015-01-22
Fix completion code for Python 3

living180:
The `filter_matches()` function in widgets/completion.py was calling the list `sort()` method with a `cmp` parameter which is not supported in Python 3.  Rework the code to only use a `key` parameter, and make a couple of other tweaks to the code.

### DIFF
--- a/cola/widgets/completion.py
+++ b/cola/widgets/completion.py
@@ -392,8 +392,8 @@ def filter_matches(match_text, candidates, case_sensitive, cmp=None):
         case_transform = lambda x: x.lower()
 
     if match_text:
-        matches = [r for r in candidates
-                    if case_transform(match_text) in case_transform(r)]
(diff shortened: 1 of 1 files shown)


cola/widgets/completion.py
```diff
@@ -317,11 +317,11 @@ def paint(self, painter, option, index):
...
-def cmp_refs(a, b):
-    """Prefer shorter refs, otherwise alphabetize them"""
-    if len(a) == len(b):
-        return cmp(a, b)
-    return cmp(len(a), len(b))
+def ref_sort_key(ref):
+    """Sort key function that causes shorter refs to sort first, but
+    alphabetizes refs of equal length (in order to make local branches sort
+    before remote ones)."""
+    return len(ref), ref
```
davvid: Brilliant!

PR #411, opened 2015-01-30
main: Refresh with 'F5' in addition to 'Ctrl+R'

cpitclaudel:
I usually press F5 a few times before I realize the correct key sequence is <kbd>Ctrl+R</kbd>.

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -1142,7 +1142,7 @@ class Rescan(Command):
 class Refresh(Command):
     """Update refs and refresh the index"""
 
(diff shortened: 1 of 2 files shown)


suside:
AFAIK you can install pyinotify and you wont have to use refresh never ever :) Details in: [cola/inotify.py](https://github.com/git-cola/git-cola/blob/HEAD/cola/inotify.py)

cpitclaudel:
That's fair; still, having F5 refresh the window would be a neat default for new users I think.

PR #421, opened 2015-03-17
Windows packaging using Pynsist

takluyver:
This is still very rough. In particular, it should be smarter about finding the git executable - it currently assumes that it's installed in Program Files (the default location). I expect that arguments in general could be handled better - I just mocked out an args object quickly to get it running.

Here is a sample installer built using current master, which works on my Windows 7 VM:
https://www.dropbox.com/s/pxo0urszizgdhcx/Git_Cola_2.1.2.exe?dl=1

To build:

1. Download PyQt4 (http://sourceforge.net/projects/pyqt/files/PyQt4/PyQt-4.11.3/PyQt4-4.11.3-gpl-Py3.4-Qt4.8.6-x32.exe ) and unpack files from Lib/site-packages into pynsist_pkgs/ in the root of the git-cola repo. Then add files from `$_OUTDIR/` into `pynsist_pkgs/PyQt4`. I hope to automate this step soon.
2. rm share/doc/git-cola/html - this is a symlink, and copying it into the build directory currently fails.
3. With pynsist and nsis installed, run `pynsist windows-installer.cfg`. The installer will be produced in build/nsis/


takluyver:
Just to be clear: I don't think this should be merged as is. I'm opening a PR for discussion about this approach.

davvid:
This looks good so far. Now if I only had a solution for the problem of the zip files taking up account disk usage on github.. I guess I can just delete old releases?

takluyver:
I don't think they have an explicit disk quota now - I think they used to, but even then it was a soft limit. Disk space is pretty cheap. So I wouldn't worry about deleting older releases until there's a problem.

https://help.github.com/articles/what-is-my-disk-quota/
https://help.github.com/articles/distributing-large-binaries/

takluyver:
Step 1 is now automated - run `./fetch_pyqt_windows.sh` to download and unpack PyQt4 Windows binaries into pynsist_pkgs.

If I come up with a general scheme for this, it will probably use Python code, but for one-off scripts, doing it in bash is convenient.

davvid:
Ya, I have no problem with bash/sh for simple stuff.

I just released v2.1.2. I'd like to make it a goal to use pynist for the next Windows release so I'll try and get my win32 environment setup soon. Is there anything else that needs changing before this is ready to merge?

takluyver:
A couple of things that we should think about:

1. What would you like to do about the symlink at share/doc/git-cola/html? Modern versions of Windows do have symlinks, but they're not often used, and I don't think NSIS has a way to create them. For now I'm just deleting it.
2. I've mocked out the parsed command line arguments in a fairly simple manual way, but it would be easy to forget to update this if you add more command line options. This is necessary because argument parsing is done in the script `bin/git-cola`, which is not an importable module. If we moved the argument parsing code into the `cola` package, we wouldn't need the DummyArgs class.
3. However, we would still need something to find where git was installed. It looks like your current installer prompts the user to locate it; Pynsist doesn't offer an easy way of prompting the user - it's possible with NSIS, but you'd need a custom template for the .nsi file. Do you think it's enough to assume the user has installed Git for Windows in one of the default locations? Should there be either a config option to the location of git, or a prompt that comes up if it can't find git when it starts?

davvid:
I eliminated the html symlink, so that should fix item 1.

Moving stuff from `bin/git-cola` into some importable module e.g. `cola/main.py` sounds like a good idea for item 2.

For item 3, it would be nice if there were some way to do this.  Prompting could be okay, or maybe having a way to configure it would be a workaround, but I don't have such a setup so I wouldn't be able to test it.

One idea is that if we check the registry (or windows apis) for the values for Program Files, etc., and then check for the existence of "Git" folders in those and other common directories then that should be good enough.

I'm not sure if users rely on the ability to install Git in arbitrary locations, or whether Windows will name ProgramFiles as e.g. C:/Programfiler/ when installed in a different language.  But, apparently windows lets you change Program Files by twiddling some registry bits, so it seems to make sense that if we read those same bits and check for `Git/` there and other common places, e.g. `C:/Git/` then we should be ok.

If we break any windows users we can look into other ways of prompting the user.  The installer could theoretically bootstrap itself by installing PyQt first, and then prompting for paths would be possible. That might be overkill if there aren't any users relying on that feature, though, so maybe we won't need it.

fetch_pyqt_windows.sh
```diff
@@ -0,0 +1,31 @@
...
+7z x -opyqt4-windows "$INSTALLER_FILE"
+
+rm -rf pynsist_pkgs
+mkdir pynsist_pkgs
+
+echo "Rearranging files into pynsist_pkgs..."
+mv 'pyqt4-windows/Lib/site-packages'/* pynsist_pkgs/
+rm pynsist_pkgs/PyQt4/assistant.exe pynsist_pkgs/PyQt4/designer.exe
+mv 'pyqt4-windows/$_OUTDIR/'*.pyd pynsist_pkgs/PyQt4/
+# These may not be necessary:
```
davvid: Right... `qsci` is QScintilla, which we don't use, so it should be okay to omit it.

The `$_OUTDIR` thing had me scratching my head -- I couldn't see where that variable was being set. Upon realizing that it's a literal dollar-OUTDIR, then I figured it must be a `7z`-ism ;-)

I'm curious why `sip` went to `PyQt4/sip`.. `sip` needs to be importable as `import sip` from the top-level, and I'm assuming that this is already handled in the above site-packages `mv` command. What's in the sip directory?  If it's just small readmes/licenses then we can keep it and maybe keep it at `pynsist_pkgs/extras/sip` so that it's tucked away separately from the importable python modules.

living180:
> I'm not sure if users rely on the ability to install Git in arbitrary locations, or whether Windows will name ProgramFiles as e.g. C:/Programfiler/ when installed in a different language. But, apparently windows lets you change Program Files by twiddling some registry bits, so it seems to make sense that if we read those same bits and check for Git/ there and other common places, e.g. C:/Git/ then we should be ok.

This would not directly affect me, as I use git-cola from a source checkout rather than using the Windows installer, but I do have Git installed in a custom location (C:\tools\git).  Just thought I would add that as a data point.  Perhaps it would make sense to search for the git executable on the PATH, first?  (`distutils.spawn.find_executable` provides this functionality:  http://stackoverflow.com/a/12611523/3573385)

kkofler:
> If we break any windows users we can look into other ways of prompting the user. The installer could 
> theoretically bootstrap itself by installing PyQt first, and then prompting for paths would be possible.

It is possible to prompt from within an NSIS script. If you're going to go through the trouble of installing PyQt and running a GUI in it, you will have to handwrite NSIS script code anyway, so you may as well just use the dialog facilities offered by NSIS.

> Upon realizing that it's a literal dollar-OUTDIR, then I figured it must be a `7z`-ism ;-)

It's actually an NSIS-ism. :-)

takluyver:
Sorry, this got buried in my inbox while I was at PyCon :-). I will work, hopefully tomorrow, on moving code from `bin/git-cola` to an importable module.

I suspect that for now it makes most sense to find git using Python code at runtime. We should certainly check on PATH first, then the default locations, possibly the registry (does git add registry entries? I couldn't see any when I looked), and fall back on asking the user.

> If you're going to go through the trouble of installing PyQt and running a GUI in it, you will have to handwrite NSIS script code

I hope not! ;-) Pynsist can generate an installer for a PyQt app without the developer having to write NSIS code by hand. We could probably even bundle git if it can easily be put in a relocatable folder.

davvid:
That sounds good.  By "registry" I think I meant the ProgramFilesDir setting in the registry at the time, but that's actually irrelevant since microsoft doesn't support changing it anyways ~  https://support.microsoft.com/en-us/kb/933700 ~ Git itself doesn't record any registry options as far as I know.

There is a relocatable version of Git, but it's not the mainstream "Git for Windows" release.  It's probably better if we track the official one.  We can probably punt on making it relocatable in the installer itself and keep it simple.

The core issue is that we cannot rely on `git.exe` being in the path on Windows.  So we could do something like e.g. allow the user to provide a `~/.config/git-cola/git-bindir` file with the location of Git's `bin/` directory on the filesystem.  The installer can write a hard-coded path to the this file for `git-cola` to read.

That way if an advanced user wants to use a relocatable version of Git then they can read our docs and point `git-cola` at it by editing this file.  The `process_args()` startup code can check for the existence of this file (on win32 only to save our home servers a `stat()`) and do the same path setup that it would do for the `--git-path` option sans the `dirname`, since we can just use the value as-is.

At that point we really don't need the `-g`, or `--git-path` options at all.  Unix users don't need that option because they already have `git` in their `$PATH`.  It's only there for Windows users, so having this silent `git-bindir` thing seems good enough without needing to make the installer too complication.

I also like that it has the potential to eliminate the command-line flag, thus simplifying the command line interface.

That said, I'm assuming that writing a file and getting the same homedir that we'd get by doing `os.path.expanduser('~/.config/git-cola')` is easy to do in the installer, if it's not I'm always open to other ideas. Let me know what you think.

davvid:
@living180 if Git can be found on the path via distutils then you won't need the path setup code done by git-cola anyways, since all it does is it places the `bin/` directory in `$PATH` so that our later code succeeds when it execs `git`.

If you have `C:/tools/git/bin` in your `$PATH` then there's nothing to do and you won't need a `~/.config/git-cola/git-bindir` file either ~ it'll just work.

davvid:
My Git for Windows ignorance is showing ;-) @living180 if you have `C:/tools/git` then surely it's relocatable, which means bundling maybe be possible too.. is there a way to bundle it indirectly so that we point to a URL and the installer optionally brings in their installer?  We can always explore bundling ideas later once we've switched to Pynsist if the bundling part is too complex for the initial version.

The "optionally" feature would be for users who already have an existing (and possibly newer) Git installed and just want to upgrade the cola part.

living180:
> My Git for Windows ignorance is showing ;-) @living180 if you have C:/tools/git then surely it's relocatable, which means bundling maybe be possible too..

I'm not 100% certain, but I do think Git for Windows is relocatable, in the sense that if you install it, and then manually move the directory where it was installed on the filesystem, it will still work just fine (though it would probably break Start Menu shortcuts and also the uninstaller).

fetch_pyqt_windows.sh
takluyver: Yep, there's a `sip.pyd` in site-packages. This was some .sip files - I suspect they're only needed at compilation time, but I copied them in just to be on the safe side. I'll try without them.

As someone else has mentioned, `$_OUTDIR` is an NSIS thing.

takluyver:
OK, I have refined the script that fetches PyQt4 a bit, and moved the command line argument parsing to a new module `cola.main`. For now, I've left it just checking the default install locations for `git` - we can add a config file or something later, but I think that probably doesn't belong amidst the launcher code.

Building the Windows installer on Linux is now two simple steps:

1. Run `./fetch_pyqt_windows.sh` to download and unpack PyQt4.
2. `pynsist windows-installer.cfg` will build the installer in `build/nsis/`

cola/main.py
```diff
@@ -0,0 +1,480 @@
...
+def shortcut_launch():
+    """Launch from a shortcut
+
+    Prompt for the repository by default, and try to find git.
+    """
+    argv = ['cola', '--prompt']
+    git_path = windows_find_git()
+    if git_path is not None:
+        argv += ['--git-path', git_path]
+    main(argv)
```
davvid: This file is missing a newline at the end of the file (sublime text?)

cola/main.py
```diff
@@ -0,0 +1,480 @@
...
+# Windows shortcut launch features:
+
+def windows_find_git():
+    """Return the path of git.exe, or None if we can't find it."""
+    # For now, we assume that git is installed in one of the typical
+    # locations. This should be smarter.
+
+    pf = os.environ.get('ProgramFiles', 'C:\\Program Files')
+    pf32 = os.environ.get('ProgramFiles(x86)', 'C:\\Program Files (x86)')
+    for p in [pf32, pf]:
```
davvid: It might not hurt to have `C:\\` in here too, so that `C:\Git\bin\git.exe` is found (a pretty common path).

cola/main.py
```diff
@@ -0,0 +1,480 @@
...
+
+
+def cmd_tag(args):
+    context = application_init(args)
+    from cola.widgets.createtag import new_create_tag
+    view = new_create_tag(name=args.name, ref=args.ref, sign=args.sign)
+    return application_start(context, view)
+
+
+# Windows shortcut launch features:
```
davvid: Would it be possible to have the `windows_find_git()` logic live in the installer so that `git-cola` itself doesn't need to contain these platform-specific details?

I was thinking that the installer could do these checks and then write bin dir it finds to `~/.config/git-cola/git-bindir` (or some other path) so that this code can simply checks for the existence of that file when running on windows and then update `$PATH` accordingly.  That way we don't have to have `C:\ProgramFiles` installation details in the code itself, if at all possible.

cola/main.py
takluyver: Pycharm. Fixed, and I've found the setting to add a newline by default, because that's been annoying me too.

cola/main.py
takluyver: Done.

cola/main.py
takluyver: It's not especially easy to add logic like that to the installer. It could be done, but even if it were easier, I'd advise against it, because:

- As a general design principle, I think the installer should be as stupid as possible. It's harder to test and debug installer code. If someone checks out the git-cola repo on Windows and runs it from source, what is going on should be as similar as practical to what happens when it's installed. The more the installer does, the greater the differences between a source checkout and an installed copy.
- Especially if we will want to prompt the user to manually tell us where git is, I imagine we can implement and maintain that much better using PyQt than using NSIS.
- Finding git at install time means the user needs to install git before git-cola. By finding it at runtime, they can do it in either order, so the GUI could prompt them to go and install it.

I do think we should ultimately eliminate the `--git-path` option in favour of storing the location in a file, but that's a significant enough change that I think it's better done as a separate PR.

fetch_pyqt_windows.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
davvid: I try to minimize the number of files in the top-level. Would you mind moving this and the other new top-level files into `./contrib/win32`?  I think the last thing would then be updating the readme to list the new instructions for creating the installer (since I don't actually know those myself). I'll try and test it soon too in my windows env. and see how it works.

Is possible for `windows-installer.cfg` to live in the subdirectory too? I'm assuming the pynsist invocation probably has an option for specifying it, so I'll wait and see if that's reflected in the readme.

I agree the `--git-path` stuff can wait for later ~ I can always tweak that stuff once this is merged. Thanks for helping with this, I'm looking forward to using it.

takluyver: I can certainly move this file.

The .cfg file can live in a subdirectory, but in that case it would package the *installed* cola module, not the one from the source tree - Pynsist looks in the directory where the .cfg file is for packages. We could have a bit of extra control, like a Makefile that copies `cola/` into `contrib/win32/` and then invokes pynsist. But I'm a bit wary of a situation that would make it easy to run Pynsist and successfully build an installer that contains a different version of the code than you expected. What do you think?

davvid: I see. In that case let's leave the .cfg in-place for now since that sounds like a good reason to keep it there.

takluyver:
OK, I've moved `fetch_pyqt_windows.sh`. Should I remove the other files in `contrib/win32/` as part of this pull request, or do you want them handy in case you need to fall back on that way of building installers.

davvid:
Let's keep them there for now. I'll try and merge this by this weekend. Any other tweaks before then? It looks good to me ~ it think it just needs the readme instructions explaining how to create the installer.

takluyver:
I've added a note at the bottom of the README. I don't think there are any tweaks I want to make.

Vdragon:
Well done :3

albertdev:
I just found this pull request after researching for somebody else how git-cola needs to be installed on Windows.

Is this new installer already available? If so, the [installation instructions for Windows](https://github.com/git-cola/git-cola/blob/master/README.md#windows-installation) are still talking about installing everything by hand.

nocnokneo:
I have the same question as @albertdev ... just trying to figure out what the state of things is here. I'm available to help build and/or install.

davvid:
I could use some help getting automation setup to build the windows installers.  Some CI thing would be ideal.  I don't have any windows machines or licenses so I'm kinda unable to help.

We did have some issues[1] with the pynsist installer in v2.3 (?) so I've been using the older innosetup infrastructure until someone has a chance to sort them out.

[1] https://github.com/git-cola/git-cola/issues/471

nocnokneo:
> I could use some help getting automation setup to build the windows installers.

I have access to a Windows machine, but I thought pynsist allows building Windows packages from Linux, no?

> Some CI thing would be ideal. I don't have any windows machines or licenses so I'm kinda unable to help.

It looks travis-ci is being used right now so the idea would be to add pynsist and whatever else to the install dependencies in travis-ci, is that right?

In the meantime, maybe the windows download link on the git-cola landing page can at least be fixed.

Vdragon:
It would be needed to use GitHub API to upload the built assets to the release page.

nocnokneo:
Okay, so like this:

https://docs.travis-ci.com/user/deployment/releases

davvid:
yes, exactly what I was thinking. In the "todo" branch (I keep a separate
clone in a Meta/ subdir with "todo" checked out) there's the release script
that I currently run to prep the release and upload the installer to
github. Replacing some (or all ;-)) of it with travis would be really
sweet.

I currently run it after pushing the branches and tag as "./Meta/release
--all". it assumes there's a windows vm with ssh running so that the script
can login and build over there. It also requires manually updating the
win32 cola sandbox before running the script. for shame!

I hadn't realized the win32 download link was busted. I'll try and get that
fixed tomorrow after I merge your latest PR ;-)

cheers,
David

takluyver:
If we do need to do the builds on Windows, [appveyor](http://www.appveyor.com/) is a Windows CI service with a free tier for open source projects.

Pywin32 is a really awkward package, but I have a ~working example of building an application with it from Linux at takluyver/pynsist#57. I suspect it depends on which bits of pywin32 are used, though - as discussed in #471, there are some modules that pywin32 generates at runtime, which make life trickier.

Issue #438, opened 2015-04-24
Consider removing "Revert Uncommited Edits..." command?


living180:
For modified files, there are two similar commands, "Revert Unstaged Edits..." and "Revert Uncommitted Edits...".  Personally, I would like to have only the first of these.  This way, the unstaged changes can go two directions, either revert (discarded) or stage (to index).  Having a command that also affects the staged changes is to me confusing and error prone:  more than once I'll stage the changes that I intend to commit, and want to throw away the rest (e.g. perhaps they were for debugging).  I always have to stop and think carefully to make sure I chose the correct "Revert" menu item to avoid throwing away the staged changes as well.  This does require two steps if wanting to revert staged changes (unstage, then revert), but for me personally at least,  I'm willing to deal with that to avoid the risk of accidentally throwing away changes I wanted to keep.

If this would be a desirable change, I'm happy to implement it and make a PR.

ugtar:
I actually +1 this

davvid:
I'm convinced too :+1:

living180:
Great, then I'll work on getting a PR up sometime next week.

living180:
Closed by PR #454.

PR #454, opened 2015-05-12
widgets.status: remove Revert Uncommitted Edits...

living180:
Fix issue #438 by removing the Revert Uncommitted Edits... item from the Status widget pop-up menu.  There is still a Revert Uncommitted Edits... item in the File Browser pop-up menu.  If that should go away as well, I can update the PR to remove it (I have no opinion on that, as I do not use the file browser).


davvid:
Thanks, @ugtar and @living180 !

PR #458, opened 2015-05-23
Use system theme icons

benedictleejh:
With KDE's new Plasma 5 Workspace, some of the icons are looking very much out of place with the new Breeze icon theme. So I've basically made git-cola take icons from the system whenever possible, instead of using its own. It does still fallback on its own icons if it can't find a system icon with the provided name.

This should help it blend better into Plasma 5's looks, with the nice side effect of probably also blending in better with other desktop environments as well.

Signed-off-by: Benedict Lee benedictleejh@gmail.com


Vdragon:
Would you like to make a screenshot of Git Cola using system theme icons?  I wonder what It'll look like after the patch?

benedictleejh:
Here are a couple of screenshots. Some of it is hard to take screenshots for, because they're in the menus. And a lot of the icons already use Qt to get the right icons in the first place, so, by and large, it already fit in reasonably well. It's just that it becomes very obvious when it doesn't, because Breeze is visually very different from Oxygen, which the original icons did a decent job of fitting well in.

Not all icons are changed though, some because they have no facsimile in the system, others because I'm unsure about them. The changed icons mostly lie in the menus and buttons, because those, I find, should be more integrated into the system look.

recently-modified
edit-remotes-dialog
main

cola/qtutils.py
```diff
@@ -515,25 +515,25 @@ def help_icon():
...
 
 def open_file_icon():
-    return icon('open.svg')
+    return QtGui.QIcon.fromTheme("document-open", icon('open.svg'))
 
 
 def options_icon():
     """Return a standard open directory icon"""
-    return icon('options.svg')
+    return QtGui.QIcon.fromTheme("configure", icon('options.svg'))
```
davvid: Sorry for the style nit, but can you please use single-quotes here so that it's consistent with the rest of the line?

In a follow-up commit, I might be tempted into refactoring theme_icon() to allow for this kind of invocation by allowing for an optional 2nd argument.  That way these calls can all use theme_icon().  Don't worry about that now, though, this'll be fine once the quote thing is fixed-up.

davvid:
Merged, thanks. I made some small tweaks to accommodate older versions of Qt in aaa3543edcf2b3cbfb788b4ea96ae9478f8b1be0 as a follow-up.

PR #465, opened 2015-06-14
Spelling fixes

scop:
Signed-off-by: Ville Skyttä <ville.skytta@iki.fi>


Vdragon:
@scop 
I'm concern about changing the `*.po` translation, it should be automatically re-generated by GNU Gettext's command (before/after a release?) than changing manually.

scop:
If that fits your workflow better, fine with me. But changing them here shouldn't cause any problems either. Would you like an updated pull req without those changes?

davvid:
Keep that note in mind for future submissions. I'll merge it as-is and re-run the po update. Thanks for the review, @Vdragon

PR #466, opened 2015-06-15
Rewrite diff syntax highlighter

### LINKED ISSUE #462: Diff containing a line "-- something" not displayed correctly
Using Git Cola 2.1.1

Example diff:
```
$ git diff 
diff --git a/test.sql b/test.sql
index a82ee34..f7bddf1 100644
--- a/test.sql
+++ b/test.sql
@@ -1 +1 @@
--- SQL comment
+-- change that comment
```

Then the diff view doesn't display the removed line with red background, but in gray, as if pertaining to the diff preamble.

To reproduce:
`mkdir bam && cd bam && git init && (echo "-- SQL comment" > test.sql) && git add -A && git commit -m "initial commit" && (echo "-- change that comment" > test.sql)`

living180:
The previous `DiffSyntaxHighlighter` worked adequately, but because it was stateless, it needed a lot of complicated regular expressions to try to match all the possible diff header lines.  It could also be thrown off, if for example a line was removed starting with `'-- '`, the resulting diff line would start with `'--- '`, causing the syntax highlighter to interpret it as a header line instead of a removal.  To address these shortcomings, completely rewrite the syntax highlighter to be stateful. This greatly simplifies the logic and drastically reduces the number of regular expressions that need to be matched.  Because `DiffSyntaxHighlighter` was the only user of `GenericSyntaxHighlighter`, and `GenericSyntaxHighlighter` does not fit the stateful approach, eliminate `GenericSyntaxHighlighter` (moving a few necessary bits to `DiffSyntaxHighlighter`) and move `DiffSyntaxHighlighter` from `qtutils` to `widgets.diff`, where it is actually used.

Closes #462.

### DIFF
--- a/cola/qtutils.py
+++ b/cola/qtutils.py
@@ -5,7 +5,6 @@ from __future__ import division, absolute_import, unicode_literals
 
 import mimetypes
 import os
-import re
 import subprocess
 
 from PyQt4 import QtGui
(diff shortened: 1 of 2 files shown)


davvid:
Nicely done. :clap:

AndiDog:
Changed/new commits in submodules are now gray text instead of black. Intended?

living180:
No, I intended to preserve pre-existing behavior (modulo an additional bug I found in the old code). I'll try to look at what is going on this week to see about getting a fix in. Thanks for testing!

Issue #459, opened 2015-05-31
"Rename branch" menu item should be disabled before initial commit


Vdragon:
Hi Git Cola developers!

It seems that an empty repo has none branches and is not possible to "rename" them, the "Rename branch" menu item should be disabled at this state since every rename attempt will error with following console output:

`````
error: refname refs/heads/master not found
fatal: Branch rename failed

程式以 128 狀態碼結束
`````

Vdragon:
@mmargoliono Thanks!

PR #475, opened 2015-07-15
Menu: Disable "Branch" menu in empty repositories

mmargoliono:
When current working repository is empty (eg: just initialised or cloned
an empty repository), it doesn't make sense to rename or delete local
branch.

For checking whether repository is empty or not, local branches is used.
Alternative considered was:
1. Git log/rev-list and check exit code (Fail when there is remote
branches, bot not local branches)
2. Git count-objects (probably has similar issue with above)

Issue: #459
Reported-By: Vdragon
Signed-off-by: Minarto Margoliono <lie.r.min.g@gmail.com>

### DIFF
--- a/cola/models/main.py
+++ b/cola/models/main.py
@@ -216,6 +216,9 @@ class MainModel(Observable):
         return not(bool(self.staged or self.modified or
                         self.unmerged or self.untracked))
 
+    def is_empty_repository(self):
+        return not self.local_branches
+
     def _update_remotes(self):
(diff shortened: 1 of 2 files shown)


cola/models/main.py
```diff
@@ -216,6 +216,9 @@ def is_empty(self):
         return not(bool(self.staged or self.modified or
                         self.unmerged or self.untracked))
 
+    def is_empty_repository(self):
+        return not self.local_branches
+
```
davvid: Beautiful!

PR #486, opened 2015-08-11
ensure Git subprocess spawns no window on win32

### LINKED ISSUE #477: git window appears despite Popen hidden flag
Using this distribution <<https://github.com/madhur/msysgit-2.0.0>>, Git window popups on every invocation. That's despite the Popen flag to specifically hide the window.

sthalik:
There's an additional Popen flag that needs specifying, as per commit.

Fixes #477

### DIFF
--- a/cola/core.py
+++ b/cola/core.py
@@ -133,6 +133,10 @@ def start_command(cmd, cwd=None, add_env=None,
         # the subprocess.
         cwd = None
 
+    if WIN32:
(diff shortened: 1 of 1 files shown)


living180:
+1

sthalik:
@davvid I agree conditional "creationflags" makes the code less messy, even though it's used only for win32. I've adjusted the wraparound with $EDITOR. Maybe long single-words could be always not broken?

PR #487, opened 2015-08-11
widgets.main: update status when main window is activated

AndiDog:
Updates the Git status (e.g. current branch in window title) when main window is activated. Helpful e.g. if you switch branches from the command line and then go back to git-cola.

### DIFF
--- a/cola/widgets/main.py
+++ b/cola/widgets/main.py
@@ -479,6 +479,8 @@ class MainView(standard.MainWindow):
         self.connect(self, SIGNAL('update()'),
                      self._update_callback, Qt.QueuedConnection)
 
+        self.installEventFilter(self)
+
         self.connect(self, SIGNAL('install_cfg_actions(PyQt_PyObject)'),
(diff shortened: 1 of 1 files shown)


davvid:
If you're on Linux then another option is using the inotify support.  Otherwise, this looks okay, but I think it might be better to make it optional.  Before installing the event filter, how about checking `gitcfg.current().get('cola.refreshonfocus', False)` ?  If it returns True then install the filter.

What do you think?

AndiDog:
I will check both options. inotify could work when watching the Git directory, however I'm currently trying to make [out-of-repo-worktree repos](http://jasonkarns.com/blog/versioning-dotfiles-with-git/) work in Git Cola, so we must ensure the file .git/HEAD is really watched. The setting is also a good way.

AndiDog:
Since git-cola already watches the working tree files, I want to watch some Git files instead (namely "HEAD") to detect branch switches or checkouts. Will open a new pull request, forget about this one ;)

PR #491, opened 2015-08-21
fix history browser default value on win32

sthalik:
Default history browser fails to start due to backslashes in Windows path. Converting backslashes to slashes beforehand makes it work for all cases.


sthalik:
To avoid doubt, slash-pathnames are a first-class citizen in win32 APIs, it's not merely Python converting them into backslashes. Even to the point that in, say, cmd.exe, the Windows command prompt, typing "cd c:/users" will succeed. Some of the userspace win32 utilities have problems with parsing slashes in their argv, but from pathname operations' perspective they're perfectly fine.

PR #497, opened 2015-09-14
Center the startup dialog on the screen and make it larger

djhaskin987:
The start-up screen's size has always been a pain point with me, because I could not see at a glance which repositories I wanted to look at with git cola. The path names of the repository were always longer than the window itself. Like this:

smallstartupgitcola

This PR widens the window to half the screen width/length, and centers the window on the screen:

bigstartupgitcola

I have found this modification very useful and wanted to share it. Please consider merging it :), if you don't like how I fixed the small window problem I can change the code to your satisfaction.


djhaskin987:
Yay!

PR #505, opened 2015-10-02
inotify: update status when checkout is performed

AndiDog:
Replaces PR #487.

The directory watcher module was changed to detect modification of the .git/HEAD file, which is touched e.g. on branch switch or other kind of checkout. This makes the GUI update automatically for these changes (including branch name in window title).

Note that for changes to files in the working tree, the GUI already updated automatically before this commit, so I don't think a preference option is necessary for this new behavior.

Tested and working on Ubuntu 15.04 and Windows 7.


davvid:
Very slick

cola/inotify.py
```diff
@@ -199,6 +204,10 @@ def _add_watch_failed_warning(self, directory, e):
         core.stderr('    sudo tee -a /etc/sysctl.conf &&')
         core.stderr('    sudo sysctl -p\n')
 
+    def _is_ignored(self, path):
+        # Ignore all Git paths except "HEAD" (notify on branch switch or other type of checkout)
+        return (path.startswith('.git') or '/.git/' in path) and not path.endswith('HEAD')
```
davvid: The `path.startswith('.git')` check, will that make it blind to .gitignore changes?  Instead of `endswith('HEAD')`, maybe `os.path.basename(path) == 'HEAD'` instead?

AndiDog:
Right, wanted to take `.git/` over from the previous win32 ignores, missed the trailing slash. Now fixed.

living180:
@davvid I don't want in any way to negate the excellent work that @AndiDog has done in this PR, but just wanted to let you know that I have a branch I have been using personally for several months that overhauls the way the change monitoring works.  The major changes that it provides are eliminating all polling for FS changes and instead moving to a fully event-driven model; resetting the refresh timer on any FS change (with the current code, the timer can expire while Git is is still running a long operation, such as a rebase on Windows, causing a conflict on Git's filesystem lock); monitoring changes in newly created directories on Linux; and most significantly refreshing on changes to the Git repo in addition to changes to tracked files (I do most of my committing/branching/checking out from the command line, so this helps ensure that Git Cola stays in sync).  It is a much more substantial rework than this PR, and the Linux side could probably stand to have some more testing, because my day-to-day work is currently in Windows.  I don't know that I can get a PR up this weekend, but if you are interested in reviewing the code, I can try to get it up this week sometime.

davvid:
I'd be happy to see a rework of the inotify code and can help test the linux side. We can take some time to let this cook -- I think I've merged the last topic for v2.4 and will be tagging it soon (probably next weekend) so there's plenty of time before the subsequent release to get this in.  Thanks for the heads-up..

davvid:
Thanks @AndiDog I think this issue should have been handled by @living180 's PR that was just merged.  Please test it out and let me know if you run into any issues.  I'm testing (in production) now too.

PR #507, opened 2015-10-05
Use Ctrl+O as shortcut for opening Git repository

AndiDog:


davvid:
Thanks.  I'd like to merge this but I vaguely recall that Ctrl-O is the platform-specific hotkey for opening the preferences page on Mac OS X.  I may be mis-remembering, though, so I'll double check now before merging.

AndiDog:
I think I also checked before (in code only, don't have a Mac) that there are no conflicts. Typically Cmd+O is "Open file" and Cmd+"," is for app preferences, if I remember correctly from my Mac experience.

Thanks for merging!

PR #510, opened 2015-10-13
Re-architect filesystem change monitoring

living180:
This is a quite substantial PR which significantly re-architects the filesystem change monitoring infrastructure.  The three primary goals were:

* Move to a completely event-driven architecture instead of requiring polling.
* Refresh on changes to the `.git` directory made outside of git-cola, e.g. after using git from the command line.
* Restart the refresh timer on file system changes.  This reduces the chance of interfering with a long-running git operation outside of git-cola (e.g. a rebase on Windows) by delaying the refresh until the filesystem changes have stopped.

In addition, a few minor fixes and cleanups were made - I tried to make separate commits for each of these.

The major side-effect of this PR is that pyinotify is no longer required on Linux.  Instead, git-cola now ships with a small ctypes-based wrapper around the inotify functions in libc.  The reasons for moving away from pyinotify are:
* It does not have support for event-driven operation.
* It does not provide access to the low-level error codes from the inotify functions, which is required for proper use of the inotify API, e.g. for handling certain race conditions on the filesystem, and for properly detecting out-of-watch-descriptor errors.
* Its API has not been stable, requiring compatibility code in git-cola.

Otherwise, I've tried to make each commit self-describing, so I won't duplicate that information here.  If there are any questions or suggestions for improvement, they are welcome.  Additionally, if any of my changes has undesired behavior, let me know and I can revert that.

I have been developing and using this branch in my daily work for several months on Windows, so I feel fairly confident of the code on that platform.  I've done some testing on Linux as well, but not nearly to the same degree, so help flushing out any bugs that might be lurking in the Linux/inotify code would be welcome.


davvid:
Very cool. I'll pull from your branch and install it at work so that we can get some quick widespread testing on the linux inotify side (in addition to my manual testing).

living180:
> I don't recall testing inotify while switching repos interactively using e.g. the "Recent" or "Favorites" widget. I wonder if we need a way to reset self._worktree, or recycle the singleton.

Yes, that should be do-able.  Especially since there is an actual class now instantiated as a singleton, it should be relatively straightforward to destroy and re-instantiate the class when switching repos, especially if there is a good place to hang the instance where it can be easily reached by everything that needs it (which is actually in very few places).  I can put this on my to-do list, but feel free to look at it also if you like.

living180:
> Please avoid set/dict-comprehensions since we still have users on python2.6. slower, but more compatible list comprehensions with a typecast should be good enough for now.

Ah, yes, I missed that - I don't use 2.6 at all, so occasionally I forget what features are exclusive to 2.7.  Will fix those up.

living180:
OK, just updated to remove all set literals and comprehensions.  If you run into anything else, please let me know!

davvid:
I tested this on el6 and debian/sid and it's looking good so far. I'm going to go ahead and merge this and deploy this to production here at work now.  Thanks again!

living180:
Most welcome.  I use git-cola virtually every time I code, so I'm happy to do what I can to contribute back.

PR #519, opened 2015-11-05
Update fedora instructions on README

fidalgo:
Fedora replaced YUM with DNF package manager.


davvid:
Thanks @fidalgo much appreciated!

PR #521, opened 2015-11-07
status: support blame GUI

AndiDog:
Adds menu option "Blame..." to status items, defaults to "git gui blame <path>".


davvid:
Nicely done.  Thanks again!

Issue #516, opened 2015-11-02
pywin32 notify during build makes dag refresh all the time


sthalik:
With pywin32 for notify support, `make -j4` makes the `git dag` blink all the time. `dag` feels the need to refresh commit list on changes to gitignored build directory.

living180:
Can you report which git-cola version you are using?  My unreleased changes to the filesystem notification code should hopefully help alleviate this problem, as it now avoids refreshing until it has not detected any filesystem activity for at least 888ms (previously upon detecting filesystem activity, it would wait 888ms and then refresh, even if there had been more filesystem activity in the interim).  However, as I just commented in issue #517, an additional improvement would be to not refresh at all if the file in question is ignored by git.

sthalik:
For the case of DAG, unless dot.git gets refreshed should be more precise, yes? DAG doesn't care about uncommitted changes.

living180:
Ah, yes.  I don't really use `git dag` but it makes sense that it doesn't need to refresh the way the status browser does.  I'll need to look at a way to make sure filesystem monitoring doesn't affect the DAG.  Thanks for reporting.

living180:
@davvid, @sthalik  Do you think it makes sense to have `git dag` auto-refresh on new commits?  Trying to decide whether to use more limited filesystem monitoring for `git dag` or to disable it altogether.  Since I don't really use `git dag`, user input would be helpful.

sthalik:
@living180 it could be refreshed on any new commit, branch, rebase, or tag change. These changes are rare enough to warrant no additional processing. Note the commit-ish selector in dag that can point at anything.

sthalik:
@living180 a simpler option would be to use Nagle's algorithm for notify refreshes in `git dag`.

davvid:
@living180 refreeshing on new commits or branches would be pretty cool, but maybe there's some things that can be tweaked project-side to make it more amenable.

If "make" is triggering refreshes then that suggests that the directories where file creation/modification is happening is inside a directory that is not ignored.

If you don't want git-dag refreshing then one way to make that happen is to add that directory to your .gitignore.  That's the first thing to try to do.  Possibly re-arranging the directory structure to place non-source files in a .gitignore-able subdir would help since we would no longer be watching items inside that directory.

[...]

PR #539, opened 2016-01-07
Fewer dag refreshes

living180:
This PR should fix issue #516.  It first of all limits the scope of `.git` dir monitoring in general to only refresh on changes to `index` and `HEAD` (in additional to changes to files under `refs/`).  Then, for `git-dag`, if further ignores changes to `.git/index` as well as the worktree.  As a result, `git-dag` should only refresh on changes to `HEAD` or `.git/refs/`.

I would like some feedback on how `git-dag` actually enables this behavior however.  I added an optional parameter to `application_start()`, `monitor_refs_only`.  That parameter gets passed in turn to `fsmonitor.instance().start()`.  However, I don't know if this the best mechanism for controlling that behavior.

### DIFF
--- a/bin/git-dag
+++ b/bin/git-dag
@@ -84,7 +84,7 @@ def parse_args():
 def cmd_dag(args):
     context = application_init(args)
     view = git_dag(context.model, args=args, settings=args.settings)
-    return application_start(context, view)
+    return application_start(context, view, monitor_refs_only=True)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


davvid:
Thanks @living180, this is nicely done.  I think the `application_start()` changes seem good enough for now.  If we later add other entry points, or the behavior diverges more, then we might want to allow the callers to supply their own setup function/object that we could invoke on their behalf.

PR #546, opened 2016-01-27
fsmonitor: use the resolved git repository path

### LINKED ISSUE #545: Exceptions in file monitor module when running in a git worktree
In a worktree, ".git" is not a directory but a file. This causes an exception in the file monitor module because _refresh_watches() collects only directories.

  File "/opt/git-cola/share/git-cola/lib/cola/fsmonitor.py", line 152, in run
    self.refresh()
  File "/opt/git-cola/share/git-cola/lib/cola/fsmonitor.py", line 207, in refresh
    self._git_dir_wd = self._git_dir_wd_map[self._git_dir]
KeyError: u'/home/weo/Projects/MyProject/.git'

davvid:
git_dir() is the worktree's .git path, but it is not necessarily
the path where refs/ objects/ and friends are found.  When using
submodules, for example, .git is a file that contains the path
to the actual git repository.

The fsmonitor registers the git path so that it can notice
changes.  When doing so it ignores ENOTDIR errors to avoid a
race condition, but it also silently ignores the request to
watch the .git-file.

This later causes an exception when asking for the git path:

	File "cola/fsmonitor.py", line 207, in refresh
	self._git_dir_wd = self._git_dir_wd_map[self._git_dir]
	KeyError: u'/home/david/src/git/.git'

Fix the logic by using git_path() when setting up the monitor so
that the real git repository path is used.

Closes #545
Reported-by: Wolfgang Ocker <weo@reccoware.de>

### DIFF
--- a/cola/fsmonitor.py
+++ b/cola/fsmonitor.py
@@ -116,7 +116,7 @@ if AVAILABLE == 'inotify':
         def __init__(self, monitor, refs_only):
             _BaseThread.__init__(self, monitor, refs_only)
             self._worktree = core.abspath(git.worktree())
-            self._git_dir = git.git_dir()
+            self._git_dir = git.git_path()
             self._lock = Lock()
             self._inotify_fd = None
             self._pipe_r = None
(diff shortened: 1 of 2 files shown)


davvid:
@living180 would you mind reviewing this change?  It seems to do the right thing for me, and the logic seems sound, but wouldn't mind a second set of eyes.

living180:
@davvid :+1: looks good to me.

PR #564, opened 2016-04-12
Better filesystem monitoring for bare repositories

living180:
Instead of using `git.getcwd()` to get the Git worktree for monitoring for filesystem changes, use `git.worktree()`.  If that function returns `None`, the repository is bare.  In that case, simply skip monitoring the (non-existent) worktree.


davvid:
Totally makes sense, thanks @living180 !

PR #565, opened 2016-04-13
Add Ctrl+K shortcut to focus status window

hvm2hvm:
Hi, I didn't see anything about your policy on pull requests but since you have a how-to fork line I thought I might just send it directly.

Using CTRL+L to focus the commit message box came naturally but going back to the status window was not as intuitive. I figured CTRL+K is a good choice since CTRL+L came to me from browser use and CTRL+K is sometimes used for the search box in browsers (e.g. firefox)

If you don't want to merge it that's fine though.

### DIFF
--- a/cola/hotkeys.py
+++ b/cola/hotkeys.py
@@ -32,6 +32,7 @@ MOVE_UP_SECONDARY = Qt.AltModifier + Qt.Key_K
 MOVE_UP_TERTIARY = Qt.ShiftModifier + Qt.Key_K
 MOVE_RIGHT = Qt.Key_L
 FOCUS = Qt.ControlModifier + Qt.Key_L
+FOCUS_STATUS = Qt.ControlModifier + Qt.Key_K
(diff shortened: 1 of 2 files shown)


davvid:
I think we already have shortcuts to focus the various windows.

To focus the commit message editor use `Ctrl+1`.
To focus the status window use `Ctrl+2`
To focus the diff window use `Ctrl+3`
etc

hvm2hvm:
I know about those but they're not consistent with the Ctrl+L one. I'm currently using it like this, i.e. I ctrl+k to the status window and add all the files I want after which I ctrl+L to go to the commit message, and then ctrl+enter. All of this is very convenient. Ctrl+1,2,3 are a bit clumsy to use - I have to move my hands away from the center of the keyboard.

davvid:
Okay, you sold me on the "I have to move my hands away from the center of the keyboard" argument. I think I'd probably start using Ctrl+K now that it's available. Thanks for the PR!

PR #567, opened 2016-04-20
dag: add menu to reset to commit

AndiDog:
Add menu for reset (soft, mixed, hard)

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -1266,6 +1266,30 @@ class Refresh(Command):
         fsmonitor.current().refresh()
 
 
+class ResetToCommit(ConfirmAction):
+
+    def __init__(self, ref, reset_type):
+        ConfirmAction.__init__(self)
+        self.model = main.model()
+        self.icon = icons.undo()
+        self.ref = ref # the commit
+        self.reset_type = reset_type
+
+    def confirm(self):
+        title = N_('Reset to commit?')
+        text = N_('This will reset to the selected commit.')
+        info = N_('Reset to commit?')
+        ok_text = N_('Reset to commit')
+        return Interaction.confirm(title, text, info, ok_text,
+                                   default=True, icon=self.icon)
+
+    def action(self):
+        return self.model.git.reset('--%s' % self.reset_type, self.ref)
+
+    def success(self):
+        self.model.update_file_status()
+
+
 class RevertEditsCommand(ConfirmAction):
 
     def __init__(self):
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -111,6 +111,12 @@ class ViewerMixin(object):
     def reset_worktree(self):
         self.with_oid(lambda oid: cmds.do(cmds.ResetWorktree, ref=oid))
 
+    def reset_to_commit(self, reset_type):
+        sha1 = self.selected_sha1()
+        if sha1 is None:
+            return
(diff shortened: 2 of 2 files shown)


cola/widgets/dag.py
```diff
@@ -131,6 +137,15 @@ def context_menu_actions(self):
...
                                self.create_tag),
+        'reset_soft':
+            qtutils.add_action(self, N_('Reset (soft)...'),
+                               lambda: self.reset_to_commit(reset_type='soft')),
+        'reset_mixed':
+            qtutils.add_action(self, N_('Reset (mixed)...'),
+                               lambda: self.reset_to_commit(reset_type='mixed')),
+        'reset_hard':
+            qtutils.add_action(self, N_('Reset (hard)...'),
+                               lambda: self.reset_to_commit(reset_type='hard')),
```
davvid: Instead of `--hard`, consider using `--merge` instead.  It's a safer choice.

cola/cmds.py
```diff
@@ -1266,6 +1266,30 @@ def do(self):
...
+    def confirm(self):
+        title = N_('Reset to commit?')
+        text = N_('This will reset to the selected commit.')
+        info = N_('Reset to commit?')
+        ok_text = N_('Reset to commit')
+        return Interaction.confirm(title, text, info, ok_text,
+                                   default=True, icon=self.icon)
+
+    def action(self):
+        return self.model.git.reset('--%s' % self.reset_type, self.ref)
```
davvid: We can reuse the double-dash logic in the git command wrapper by writing this as,
```python
kwargs = {self.reset_type: True}
return self.model.git.reset(self.ref, **kwargs)
```

cola/widgets/dag.py
```diff
@@ -111,6 +111,12 @@ def reset_branch_head(self):
     def reset_worktree(self):
         self.with_oid(lambda oid: cmds.do(cmds.ResetWorktree, ref=oid))
 
+    def reset_to_commit(self, reset_type):
+        sha1 = self.selected_sha1()
+        if sha1 is None:
+            return
+        cmds.do(cmds.ResetToCommit, reset_type=reset_type, ref=sha1)
```
davvid: The constructor requires these two arguments (they don't have default values) so it probably makes sense to reorder these and pass them as `cmds.do(cmds.ResetToCommit, sha1, reset_type)`

Also, the logic is pretty similar to the other "needs an object id" actions here, so it probably makes sense to reuse the `if sha1 is None` logic in the `with_oid()` helper.

```python
self.with_oid(lambda  oid: cmds.do(cmds.ResetToCommit, oid, reset_type))
```

cola/widgets/dag.py
```diff
@@ -191,6 +209,11 @@ def context_menu_event(self, event):
         menu.addSeparator()
         menu.addAction(self.menu_actions['create_branch'])
         menu.addAction(self.menu_actions['create_tag'])
+        reset_menu = menu.addMenu(N_('Reset'))
+        reset_menu.setEnabled(self.menu_actions['reset_soft'].isEnabled())
```
davvid: The menu disabling is a nice touch here. Thanks for the careful attention to usability details.

AndiDog:
I just saw you already implemented that in January (413d445b5e2e2f604cd63633951eecb826be9625). Totally overlooked that - probably because it was on my TODO list for longer than that. Guess we can ignore this PR completely.

PR #570, opened 2016-05-03
Fix main Read the Docs links

living180:
The main Read the Docs site is still at `readthedocs.org`, so switch those links back from `readthedocs.io`.


davvid:
Good catch, thanks

PR #589, opened 2016-08-21
fix early crash on python3 + PyQt5

sthalik:
Must pass "bytes" data type to QByteArray in PyQt5.

### DIFF
--- a/cola/compat.py
+++ b/cola/compat.py
@@ -14,7 +14,12 @@ WIN32 = sys.platform == 'win32' or sys.platform == 'cygwin'
 
 try:
     # pylint: disable=bytes-builtin
-    bstr = bytes
+    bytes()
+    if PY3:
+        def bstr(x):
+            return bytes(x, encoding='latin1')
+    else:
+        bstr = bytes
 except NameError:
     # Python <= 2.5
     bstr = str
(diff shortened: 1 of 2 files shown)


cola/compat.py
```diff
@@ -14,7 +14,12 @@
 
 try:
     # pylint: disable=bytes-builtin
-    bstr = bytes
+    bytes()
```
davvid: this should work sans the `()`.... could even do `b = bytes`

cola/compat.py
```diff
@@ -14,7 +14,12 @@
...
+    if PY3:
+        def bstr(x):
+            return bytes(x, encoding='latin1')
```
davvid: this is starting to look suspiciously like `core.encode()`.  Shouldn't it be `utf8`?

cola/widgets/standard.py
```diff
@@ -128,9 +129,12 @@ def apply_state(self, state):
         if windowstate is None:
             result = False
         else:
+            windowstate = str(windowstate)
+            if qtpy.PYQT5:
+                windowstate = str.encode(windowstate, encoding='us-ascii')
```
davvid: this should probably use `core.encode(...)`

davvid:
Notes aside, i think I'll take this PR as-is and see what I can tweak on my end.  Thanks for the fix!

cola/compat.py
sthalik: Is it for human-readable strings? If so, definitely.

PR #590, opened 2016-08-22
Python3 PyQt5 compatibility tweaks

davvid:
Simplify the fixes from #589 related to Python3 / PyQt5 compatibility.


davvid:
@sthalik here's the super tiny tweaks I ended up making to your PR.  Thanks again!  Please try this on your setup and let me know if I missed anything.  I tested PyQt4/5 on Python2.7 and PyQt5 on Python3, but since we arranged to only have this affect Python3 I think it should be fine elsewhere.

PR #595, opened 2016-09-06
Allow auto-reset of "Bypass Commit Hooks" setting

AndiDog:
This feature is typically used in exceptional cases (e.g. commit something that I will not push unchanged), so my usual workflow is to reset the flag immediately after using it. Add a configuration for auto-resetting that flag.

### DIFF
--- a/cola/widgets/commitmsg.py
+++ b/cola/widgets/commitmsg.py
@@ -460,6 +460,7 @@ class CommitMessageEditor(QtWidgets.QWidget):
         sign = self.sign_action.isChecked()
         status, out, err = cmds.do(cmds.Commit, amend, msg, sign,
                                    no_verify=no_verify)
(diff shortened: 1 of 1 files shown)


davvid:
The `Bypass Commit Hooks` setting definitely does seem like a one-off use setting, similar to the `Amend` mode that clears itself as soon as a commit is made.

I'm slightly in favor of making this the default behavior, and not have a configuration option for it at all.

IMO, if someone needs to keep disabling their commit hook then they can just as easily disable it completely at the repository level.  From that perspective, the code that disables the `Bypass Commit Hooks` option should always run.

What do you think?

AndiDog:
I'm actually of the same opinion, but just made it backward compatible since I thought that's the usual way. Ok, then please wait until I've adapted to make it the default without configuration.

AndiDog:
Done, now it's a really simple one line change ;)

Issue #599, opened 2016-09-14
Assing new name and rename items in Favorites list


IgorGalarraga:
[whislist]

I usually clone git repositories to ./git directory inside projects directories:
projects/projectAA/example
projects/projectAA/git
projects/projectBB/foo
projects/projectBB/git

In this example, both repositories on favorites will be added as 'git'

It would be very usefull to assign custom names on item addition.
Also, it would be very interesting the item rename functionality.

This, (renaming) could be extended to recent list. Displayed items, have the described problem.

PR #601, opened 2016-09-16
Added custom name for favorites item

IgorGalarraga:
* Refs: #599
* Custom name for favorites item.
* Recent item name is now full path.
* Very initial implementation. My first contact with Python.
* TODO:
- [ ] Old settings file (~/.config/git-cola/settings) must be removed. Make compatibility.
- [ ] Unify 2 prompt windows when add favorite on one window.
- [ ]  Traductions for new strings.
- [ ] Add rename line on contextual menu.
- [ ]  Extend renaming to recent repo list.

### DIFF
--- a/cola/settings.py
+++ b/cola/settings.py
@@ -63,7 +63,7 @@ class Settings(object):
         missing_recent = []
 
         for bookmark in self.bookmarks:
-            if not self.verify(bookmark):
+            if not self.verify(bookmark['path']):
                 missing_bookmarks.append(bookmark)
 
         for bookmark in missing_bookmarks:
@@ -82,13 +82,15 @@ class Settings(object):
             except:
                 pass
 
-    def add_bookmark(self, bookmark):
+    def add_bookmark(self, path, name):
         """Adds a bookmark to the saved settings"""
+        bookmark = { 'path': path, 'name' : name }
         if bookmark not in self.bookmarks:
             self.bookmarks.append(bookmark)
 
-    def remove_bookmark(self, bookmark):
+    def remove_bookmark(self, path, name):
         """Remove a bookmark"""
+        bookmark = { 'path': path, 'name' : name }
         if bookmark in self.bookmarks:
             self.bookmarks.remove(bookmark)
 
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -569,10 +569,11 @@ class RemoteRename(RemoteCommand):
 
 class RemoveFromSettings(ConfirmAction):
 
-    def __init__(self, settings, repo, icon=None):
+    def __init__(self, settings, repo, name, icon=None):
         ConfirmAction.__init__(self)
         self.settings = settings
         self.repo = repo
+        self.name = name
         self.icon = icon
 
     def success(self):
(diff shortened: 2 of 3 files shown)


davvid:
Thanks for this first version.  It looks like we're on the right track. The backwards-compatibility thing is probably essential to iron out before rolling this in.  I'll see if I can lend a hand in the coming weeks.

Please run the test suite via `make test`.  You may have noticed that the Travis CI build failed because the tests were broken by these changes, so we'll have to get that in order too.

Let me know if you have any questions.  I'll do a thorough review of this first version over the weekend.

Lin-Buo-Ren:
I haven't tried the patch, but I wonder if it's sane to rather use the `<repo>/.git/name` traditional(?) file for repo description instead.

IgorGalarraga:
@Lin-Buo-Ren Is the first time that I ear about .git/description file for GitWeb application. Thanks!.
After evaluate this a bit, I think that the best way to add this file usage could be:
* Mantain repo path and name on git-cola settings. Like now.
* On Favorite add action, check this file: if it has a name, propouse this name. If it is default (as they are in all my repos! ;) ) propouse repo base name like now.
* On settings create a new check box to select or not this file sincronization with git-cola renaming system.

IgorGalarraga:
@davvid I know that the backwards-compatibility is very important, but it seem difficult for me and for now with my limited knowledge on Python, anyway, I will try to do it.
After pull-request I check Travis CI errors, I will try to pass them: some functions has changed their number of parameters.

cola/settings.py
```diff
@@ -60,18 +60,34 @@ def __init__(self, verify=git.is_git_worktree):
...
+            try:
+                if not self.verify(bookmark['path']):
+                    missing_bookmarks.append(bookmark)
+            except:
+                """Old version bookmark format """
+                new_bookmark = { 'path' : bookmark, 'name' : os.path.basename(bookmark) }
+                if not self.verify(bookmark):
+                    missing_bookmarks.append(bookmark)
+                else:
+                    update_bookmarks.append(new_bookmark)
```
IgorGalarraga: Two bookmarks version are a list (of different components). `self.remove_missing()` works well on both settings file versions  (on my tests), because it deals with whole unit on list not with content, after this parse, the bookmark will be updated.
Yes, `try/except` not is more elegant, the use of `if isinstance(bookmark, dict):` is the correct choice for this code.
In my first try, I wanted do a first upgrade parse, but I did not know to find the solution in Phyton... thanks for your example.

cola/settings.py
```diff
@@ -60,18 +60,34 @@ def __init__(self, verify=git.is_git_worktree):
...
 
         for bookmark in missing_bookmarks:
             try:
                 self.bookmarks.remove(bookmark)
             except:
                 pass
 
+        """In case of old settings version, clear old bookmars list and fill with new list"""
```
IgorGalarraga: Thanks for the clarification.

cola/settings.py
IgorGalarraga: My comment:

> Two bookmarks version are a list (of different components). self.remove_missing() works well on both settings file versions (on my tests), because it deals with whole unit on list not with content, after this parse, the bookmark will be updated.

was an explanation of behavior of `self.bookmarks.remove(bookmark)` into `def remove_missing(self)`, not for `self.remove_missing()`...

IgorGalarraga:
When add `def upgrade(self):` I have 2 problems:
* I can't directly update self.bookmarks like your example. I must use other list and use `self.bookmarks.remove` and `self.bookmarks.append`. BUT:
* It seems that changes are overwritten (with file contents in old version) again in `remove_missing`

Therefore I have changed `try/except` using `isinstance`.

Also code style has been updated.

I am not able to make a prompt window with two text boxes... any help (url, reference, example)?

Thanks!

PR #611, opened 2016-10-20
Rebase win32 fix and window size improvements

sthalik:
Rebase win32 fix is commented with regard to its logic. As for creationflags and STARTUPINFO I moved it to the same place. There was no need to put them in random files during the initial fix.

The same behavior occured on both msys2 (Cygwin fork) git and official git apparently built with mingw-w64. Perhaps git has some flaky logic in interactive rebase.

As for widget size changes - I can submit them separately if you wish. Or just merge `FETCH_HEAD^^` 👍

### DIFF
--- a/cola/widgets/standard.py
+++ b/cola/widgets/standard.py
@@ -59,6 +59,13 @@ class WidgetMixin(object):
         if gitcfg.current().get('cola.savewindowsettings', True):
             settings.save_gui_state(self)
 
+    def resizeEvent(self, event):
+        state = self.windowState()
+        maximized = bool(state & Qt.WindowMaximized)
+        if not maximized:
+            self._unmaximized_size = event.size()
+        super(WidgetMixin, self).resizeEvent(event)
+
     def restore_state(self, settings=None):
         if settings is None:
             settings = Settings()
@@ -89,11 +96,21 @@ class WidgetMixin(object):
         """Exports data for view save/restore"""
         state = self.windowState()
         maximized = bool(state & Qt.WindowMaximized)
+        width, height = self.width(), self.height()
+
+        # when maximized we don't want to overwrite saved width/height with
+        # desktop dimensions.
+        if maximized:
+            try:
+                old_size = self._unmaximized_size
+                width, height = old_size.width(), old_size.height()
+            except:
+                width, height = None, None
         return {
             'x': self.x(),
             'y': self.y(),
-            'width': self.width(),
-            'height': self.height(),
+            'width': width,
+            'height': height,
             'maximized': maximized,
         }
 


cola/widgets/standard.py
```diff
@@ -89,13 +89,26 @@ def export_state(self):
...
-            'width': self.width(),
-            'height': self.height(),
+            'width': width,
+            'height': height,
             'maximized': maximized,
         }
+        if ret['width'] is None or ret['height'] is None:
+            del ret['width']
+            del ret['height']
+        return ret
```
davvid: Makes sense.  It's unfortunate that we have to read from disk to get the old size.  It'd be nice if we could store the pre-maximized state in memory and then just ask the widgets.

One small sug.. maybe instead of setting them to `None` and then checking afterwards, how about we remove them from the initial `ret = {...}` statement and then do:

```
    if width and height:
        ret['width'] = width
        ret['height'] = height
    return ret
```

at the end so that it avoids adding it in the first place..  Alternatively, is there any downside in not setting them to `None`?  It seems like that'd still have the desired effect since the common code path is that it'll have a previous size.

cola/core.py
```diff
@@ -143,8 +144,18 @@ def start_command(cmd, cwd=None, add_env=None,
...
+        # git-cola invokes git, which is very annoying.  The code below
+        # prevents this by ensuring that any window will be hidden.
+        startupinfo = subprocess.STARTUPINFO()
+        startupinfo.dwFlags = subprocess.STARTF_USESHOWWINDOW
+        startupinfo.wShowWindow = subprocess.SW_HIDE
+        extra['startupinfo'] = startupinfo
+
+        if WIN32 and not no_win32_startupinfo:
+            CREATE_NO_WINDOW = 0x08000000
+            extra['creationflags'] = CREATE_NO_WINDOW
```
davvid: This is a great solution.  I'll merge the early part first and we can follow-up with the later parts since they have some nice usability improvements.

davvid:
I've merged the early part now, please rebase

sthalik:
@davvid it's probably more of what the user wants to save last unmaximized position, not the last unmaximized session's position. Working on it.

cola/widgets/standard.py
```diff
@@ -89,11 +96,21 @@ def export_state(self):
...
+        width, height = self.width(), self.height()
+
+        # when maximized we don't want to overwrite saved width/height with
+        # desktop dimensions.
+        if maximized:
+            try:
+                old_size = self._unmaximized_size
+                width, height = old_size.width(), old_size.height()
+            except:
+                width, height = None, None
```
davvid: Wow, nice use of the `resizeEvent()` to capture the original size.  Is there any concern about `None` getting into the settings file?  It seems like that can happen since we unconditionally add `width` and `height` to the returned dictionary.  Is that ok?

PR #614, opened 2016-10-21
Unified prompt window adding favourites

IgorGalarraga:
Implemented promptDouble that permits edit two edit box in same dialogue. Used for adding favourites and prompts favourites name and path.

Started with: #599


IgorGalarraga:
@davvid thanks for your advices.

IgorGalarraga:
Thanks for your advices, again! :)

PR #618, opened 2016-10-25
Require remote branch to be specified in Push dialog for non-fast forward mode since implementation only does force-push in that case

AndiDog:
Even if "Fast Forward Only" was deselected, a branch wouldn't be force-pushed unless the remote branch is selected in the dialog. That's because of the implementation

```
def refspec(src, dst, ffwd, push=False):
    if push and src == dst:
        spec = src
    else:
        spec = '%s:%s' % (src, dst)
    if not ffwd:
        spec = '+' + spec # <-----------
    return spec


def refspec_arg(local_branch, remote_branch, ffwd, pull, push):
    """Return the refspec for a fetch or pull command"""
    if not pull and local_branch and remote_branch: # <-------------
        what = refspec(remote_branch, local_branch, ffwd, push=push)
    else:
        what = local_branch or remote_branch or None
    return what
```

This change therefore requires the remote branch to be selected. I could have changed the implementation, but I think since force-push is so dangerous, it's a safe choice to require the user to select the target.

### DIFF
--- a/cola/widgets/remote.py
+++ b/cola/widgets/remote.py
@@ -438,6 +438,14 @@ class RemoteActionDialog(standard.Dialog):
                 info_txt = N_('Force fetching from %s?') % remote
                 ok_text = N_('Force Fetch')
             elif action == PUSH:
+                if not remote_branch:
+                    # Push implementation (see function `refspec_arg`) only
+                    # does force-push if remote branch name is given
+                    title = N_('Force Push')
+                    msg = N_('You must first select the remote branch for a force-push.')
+                    qtutils.critical(title, msg)
+                    return
+
                 title = N_('Force Push?')
                 msg = N_('Non-fast-forward push overwrites published '
                          'history!\n(Did you pull first?)')


kkofler:
I think the code should just be fixed instead.

AndiDog:
I see the following options:

1. Require selection in dialog (my solution)
2. Just force-push to upstream branch if no remote branch was selected by the user. Don't know by heart how the command would look like.
3. Have the implementation return an error in that case, and from that cause show the dialog (as in my solution, but error coming from implementation, instead of UI making assumptions about implementation).

I would prefer 1 or 2. If we opt for 2, I will research what safe options there are for the command line, or use `--force` otherwise.

davvid:
Hmm I do like that this makes it a little bit safer, but since we already prompt the user about the force-push then it probably makes sense to try and make it just work.

The syntax for force-pushing would be e.g. `git push origin +branch-name` so it's probably a small oversight that is preventing this from working.

While it does repeat the logic a little.. this might work for the `else` block in `refspec_arg()`:

```python

what = local_branch or remote_branch or None
if what and not ffwd:
    what = '+' + what

```

kkofler:
Why not just use the `-f` flag (which is separate from the branch and can be given even without one) instead of the `+` shortcut (which has to be prepended to the branch)?

davvid:
I've always wanted to refactor the code so that fetch, push, and pull are better handled.  Passing `force=<bool>` to the underlying command object would probably make things simpler too.

Also, taking a look at the UI, Fetch and Push should probably say "Force" in the UI.

Only Pull should say "Fast-forward Only", and that checkbox is hidden for Pull (that's a mistake).  unless someone beats me to it I'll take a look at fixing this shortly.

davvid:
As a short-term safety measure, this is good so I'll merge this for now before I start tweaking it.

PR #654, opened 2017-01-31
Simplify column assignment algorithm

laerreal:
A graph configuration is found that is not correctly handled using current way to maintain occupied column set. If all children of a fork commit are merges and they are assigned a column before the commit processing then no one of children will be assigned same column. Also, no one of them will left parent's column because of the condition in the main loop. In other configuration the column is left when a non-fork ancestor at the column merges. But, in current configuration, no child remains on the column. As a result, the column 'leaks'. This causes the graph spreading.

Such configuration is not found in Git-Cola graph. I've found one in Qemu graph. After selected fork the 7-th column leaks.

example

Investigating the bug I have realized that fixing up the predicate is possible but it will result in new nested loop. Instead, another way was found. It is described in comments and commit messages.

### DIFF
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -1665,18 +1665,6 @@ coordinates based on its row and column multiplied by the coefficient.
                 # columns for such nodes.
                 node.column = self.alloc_column()
 
-            if node.is_merge():
-                for parent in node.parents:
-                    if parent.is_fork():
-                        continue
-                    if parent.column == node.column:
-                        continue
-                    if parent.column is None:
-                        # Parent is in not among commits being laid out. Hence, it
-                        # have no column.
-                        continue
-                    self.leave_column(parent.column)
-
             node.row = self.alloc_cell(node.column, node.tags)
 
             # Allocate columns for children which are still without one. Also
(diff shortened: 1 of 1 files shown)


davvid:
This is great.  I wonder if it would be resilient to octopus merges, and octopus over octopus.  I don't think I've ever seen those in the wild, though ;-)

One thing I noticed is that the first-parent chain seems to get lost on the right side, and this tiny tweak seems to make keep things left-aligned

```
diff --git a/cola/widgets/dag.py b/cola/widgets/dag.py
index 3562226a..15ecdbeb 100644
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -1112,7 +1112,7 @@ class GraphView(QtWidgets.QGraphicsView, ViewerMixin):
     x_adjust = Commit.commit_radius*4/3
     y_adjust = Commit.commit_radius*4/3
 
-    x_off = 18
+    x_off = -18
     y_off = -24
 
     def __init__(self, notifier, parent):
```

I'm going to merge this as-is, but might apply that on top later.  What do you think?

The negative offset seems to work ok on the qemu repo too.  In the cola repo the first-parent history chain become easier to see with a negative x offset.  Any downsides?

There also seems to be a rendering artifact from overlapping tags with the commit circle (`git dag v2.7.0` in qemu) when jumping up and down between the top two commits.  The items might be overlapping. I'll see if I can find what's causing that.

laerreal:
The problem is detection of 'first' ('main') chain. (I think, 'first-chain' is the one that was 'master' before the merge. Correct me if it is wrong.) If 'master' branch consists of merge commits only then the heuristic in column distribution assigns same column to those merge commits. Because generation of each other is gather than generation of any other child of the fork. But, in real life, 'master' contains regular commits too. It is not clear how to distinguish what branch is 'main'. In fact, same generation commits is assigned random columns. As a result, the first-parent chain is shifted to side.

PR #656, opened 2017-02-01
Symmetric DAG

laerreal:
Current column allocation algorithm is too straightforward. It results in following issues.

Graph inclines towards one direction. I think symmetric DAG would appear more pretty.

Graph is spread enought after merging of many parallel branches. In such situation new branches may jump far enough from its forks because the fork column is far from 0 column and there is free column near 0.

The patch series targets mostly this issues.

Changing of graph alignment in commit 02818e2 breaks tag label overlapping prevention algorithm. The patch dd164d177294e0c7f1bb94629c48564badf52364 improves the algorithm to be adaptive to current graph alignment.

### DIFF
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -1605,18 +1605,31 @@ step 2. Hence, it must be propagated for children on side columns.
             # First commit must be assigned 0 row.
             self.frontier[column] = 0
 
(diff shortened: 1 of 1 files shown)


davvid:
Awesome, thanks!

Issue #652, opened 2017-01-28
git-dag hangs indefinitely on start


Coacher:
Hello.

I have git-cola @ 9c5473babfcb2e170f4d0c6dcce53abdb24a19a5.
git-dag never launches whatever I do.

Steps to reproduce (any of the following):
- Run `git-dag` from a terminal inside git repository.
- Run `git-dag` from a terminal outside git repository and choose one in the repo select dialogue.
- From a running git-cola window click View->DAG.

Expected results:
git-dag is started.

Actual results:
Nothing happens. There are no errors or crashes. When started from a terminal `git-dag` just silently sits there indefinitely. Removing git-cola configuration from ~/.config doesn't help either.

davvid:
Interesting -- the underlying widgetse being used have not changed so here's a few questions that might help you find a solution.

- Did it work in a previous commit?
  - If so, can the problem be bisected down to a single commit?
- Are you using PyQt5?
- Does it work in PyQt4?
  - You can do `env QT_API=pyqt4 git cola` to force PyQt4 to be used when both are installed
- We use QGraphicsScene and QGraphicsView widgets for rendering.  Do other Qt apps that use these work?  There are various examples [here](http://www.programcreek.com/python/example/52415/PyQt4.QtGui.QGraphicsScene) that you can try. If QGraphicsScene is not working in general then there may be something lower-level going on (e.g. graphics related)
- Are you on Linux, Windows, or Mac?
- Does it get as far as running the `git log` commands?  You can do `env GIT_COLA_TRACE=1 git dag` to make it print the git commands it runs
- "Never launches" literally means that the window is never shown, correct?  If I'm understanding that correctly what it means is that the window never appears, not that the window appears but it stays empty.

Coacher:
I'm using PyQt4. PyQt5 is not installed. All other Qt apps work fine, both Qt4 and Qt5. OS is Gentoo Linux amd64. Raw git commands in terminal work fine, git is 2.11.0. Yes, the git-dag window is never shown. However, repo selection dialogue is shown when git-dag is started outside of git repo.

Here's what happens if I run `env GIT_COLA_TRACE=1 git dag` and choose a local clone of git-cola repo @ 9c5473ba:
```
$ env GIT_COLA_TRACE=1 git dag
git -c diff.suppressBlankEmpty=false config -z --get-all cola.icontheme
git -c diff.suppressBlankEmpty=false config --null --file /home/coacher/.gitconfig --list
[...]

PR #657, opened 2017-02-03
Fix up infinity loop bug while multiroot graph drawing

laerreal:
I just faced a DAG window freezing. When a column is allocated after all columns were left, declare_column method falls into infinity loop searching for an occupied column. The situation may occurs during multiroot graph drawing.

### DIFF
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -1586,7 +1586,11 @@ step 2. Hence, it must be propagated for children on side columns.
 
     def declare_column(self, column):
         if self.frontier:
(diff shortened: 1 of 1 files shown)


davvid:
@Coacher I wonder if this PR addresses the issue you've been seeing where the DAG window never appears?  This topic is about an infinite loop, so maybe it's possible that the repo you were viewing had a similar configuration.  Figured it's worth a shot ;-)

Coacher:
@davvid, this patch on top of 830dad0b54566cf6e2d81cb2029d3eba3246a7e6 doesn't solve #652.

PR #658, opened 2017-02-03
Fix up edge disappearing

laerreal:
Ascending order of generations may be broken during loading of a big graph. This results in repositioning of commit nodes. On the other hand, every edge is created once during commit addition and edge bounding rectangle is defined during the edge creation. This confuses Qt causing it skip edge redrawing.

before

This patch series introduces edge invalidation and optimizes edge path update.

after

### DIFF
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -954,6 +954,8 @@ class Commit(QtWidgets.QGraphicsItem):
         self.pressed = False
         self.dragged = False
 
+        self.edges = {}
+
     def blockSignals(self, blocked):
         self.notifier.notification_enabled = not blocked
 
@@ -1501,7 +1503,14 @@ class GraphView(QtWidgets.QGraphicsView, ViewerMixin):
                 except KeyError:
                     # TODO - Handle truncated history viewing
                     continue
-                edge = Edge(parent_item, commit_item)
+                try:
+                    edge = parent_item.edges[commit.oid]
+                except KeyError:
+                    edge = Edge(parent_item, commit_item)
+                else:
+                    continue
+                parent_item.edges[commit.oid] = edge
+                commit_item.edges[parent.oid] = edge
                 scene.addItem(edge)
 
     def layout_commits(self):
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -1519,9 +1519,23 @@ class GraphView(QtWidgets.QGraphicsView, ViewerMixin):
 
     def layout_commits(self):
         positions = self.position_nodes()
+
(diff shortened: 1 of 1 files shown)


davvid:
I've merged both 657 and 658 into a local branch and I'm seeing an error when visualizing the git.git repository.

To reproduce:

```sh
$ git clone git://github.com/git/git.git
$ cd git
$ git checkout 6e3a7b3398559305c7a239a42e447c21a8f39ff8
$ ../git-cola/bin/git-dag --count 10000
  File "/home/david/src/git-cola/cola/widgets/dag.py", line 620, in add_commits
    self.graphview.add_commits(commits)
  File "/home/david/src/git-cola/cola/widgets/dag.py", line 1501, in add_commits   
    self.layout_commits()
  File "/home/david/src/git-cola/cola/widgets/dag.py", line 1530, in layout_commits
    positions = self.position_nodes()
  File "/home/david/src/git-cola/cola/widgets/dag.py", line 1793, in position_nodes
    self.recompute_grid()
  File "/home/david/src/git-cola/cola/widgets/dag.py", line 1775, in recompute_grid
    self.propagate_frontier(child.column, node.row + 1)
  File "/home/david/src/git-cola/cola/widgets/dag.py", line 1726, in propagate_frontier
    current = self.frontier[column]
KeyError: -5
error: git-dag died of signal 6
```

This tweak seems to fix that:

```
@@ -1723,7 +1724,10 @@ step 2. Hence, it must be propagated for children on sid
         return cell_row
 
     def propagate_frontier(self, column, value):
-        current = self.frontier[column]
+        try:
+            current = self.frontier[column]
+        except KeyError:
+            current = self.frontier[column] = max(self.frontier.values())
         if current < value:
             self.frontier[column] = value
 
```

One thing that might be worth fixing is the scene rect calculation.  `git.git` shows the nodes going off to the left but the scrollbar doesn't allow moving over there.  I'm going to try adjusting the scene rect by tracking the minimum x position.

laerreal:
Yes, the tweak does. But something went wrong earlier. The declare_column had not defined frontier for column -5 during allocation. I will investigate this tomorrow.

davvid:
I merged this in and did a few fixups so please rebase to the latest master.  Notably you'll see that the `boundingRect()` implementation for `Label` is better now, `shape()` was removed so that we can rely on the default implementation, and we no longer call `setSceneRect()`, to also allow the default behavior to kick in.

This fixes the scene bounds issues I was seeing in git.git, and rendering artifacts from the incorrect `boundingRect()`.  It includes the above fixup, but please do take a look to see if there's anything further that should be tweaked.

PR #659, opened 2017-02-06
Depth-first traversal column reset

laerreal:
I've investigated the issue about KeyError in propagate_frontier. That was 52-th column in may case. Sometimes the error had not appeared. Hence, I guess the error is related to non-determinism of incremental graph loading. All children of currently added nodes are assigned a column. But, some of them might not be reset because they are not added yet. During consequent iteration the algorithm assumes that corresponding column is allocated because it was not reset. This time the error raises.

I suggest to use depth-first traversal of graph during reset. Complexity of the algorithm seems to remain O(N) while this guaranties that all nodes are reset.

I had not got an error during 10 runs. However, it would be nice if you check it at your side.

### DIFF
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -1652,8 +1652,21 @@ step 2. Hence, it must be propagated for children on side columns.
     """
 
     def reset_columns(self):
+        # Some children of displayed commits might not be accounted in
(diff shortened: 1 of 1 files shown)


davvid:
Thanks for the detailed descriptions. Makes perfect sense

PR #662, opened 2017-02-09
fix radios action

javierrodriguezcuevas:
Hi,
This solves a strange behavior. When you clicked on the radio buttons in 'rev_radio_group' to list the branches it did not get the clicked element correctly and it was necessary to click the element twice to get correctly.


javierrodriguezcuevas:
Hi again, sorry for not signing before. I followed your instructions and I hope it is now correct.
If it is not I will make a new pull request with "sign-off" commit.

Thanks you!

davvid:
Thanks again!

PR #664, opened 2017-02-19
Feature/branch widget

javierrodriguezcuevas:
Hi David,

With this pull request I will like to introduce a new widget for the main window, a visual tree to navigate through the branches with common actions in context menu. 
As the nomenclature I use for branch names is `type/description` The branches are grouped by their names separated by character '/' and children are orderer by their names.
Ex:
```
    branch/feature/foo
    branch/feature/bar
    branch/doe
```
Will produce:
```
    branch
        - doe
        + feature
            - bar
            - foo
```

In action:
captura de pantalla de 2017-02-19 13-22-28

I also added some documentation to `git-cola.rst`

Tested and working on:
1. Ubuntu 16.04


cola/widgets/branch.py
```diff
@@ -0,0 +1,520 @@
...
+        remote_branch = gitcmds.tracked_branch(branch)
+
+        if remote_branch is not None:
+            rgx = re.compile(r'^(?P<remote>[^/]+)/(?P<branch>.+)$')
+            match = rgx.match(remote_branch)
+
+            if match:
+                remote = match.group('remote')
+                branch_name = match.group('branch')
+                self.git_action_async('push', [remote, branch_name])
```
davvid: Just want to make sure I'm following -- the assumption in the "Push" action here is that, if the selected branch is `origin/master` and we select "Push" then we want to push the local "master" branch to the remote.  That probably warrants an in-code comment.

cola/widgets/branch.py
```diff
@@ -0,0 +1,520 @@
...
+
+        if branch != self.current_branch and qtutils.confirm(title, question,
+                                                             info, ok_btn):
+            remote = False
+            root = self.tree_helper.get_root(self.selected_item())
+            if NAME_REMOTE_BRANCH == root.name:
+                remote = True
+
+            if remote is False:
+                self.git_action_async('delete_local', [branch])
```
davvid: In the conditional above, instead of saying `if remote is False:`, in python it's more idiomatic to write `if not remote:`.  Not necessary, but flipping the flow so that it's a positive test is preferred, e.g. do the `if remote:` ... part first, then do `not remote` part in the `else:`.

cola/widgets/branch.py
```diff
@@ -0,0 +1,520 @@
...
+            for key in grouped_branches.keys():
+                item = BranchTreeWidgetItem(key, child_icon)
+                item.addChildren(create_children(grouped_branches[key]))
+
+                if item.childCount() > 0:
+                    item.setIcon(0, icons.ellipsis())
+
+                result.append(item)
+
+            result.sort(key=lambda x: x)
```
davvid: The `key=lambda x: x` might not do anything.. maybe the `key=...` argument can be removed?

cola/widgets/branch.py
```diff
@@ -0,0 +1,520 @@
...
+
+        return result
+
+
+class GitHelper(object):
+    def __init__(self, git):
+        self.git = git
+
+    def log(self, origin, args):
+        return self.git.log(origin, *args)
```
davvid: `git log` tends to work better if you stick to `git log [options] [revargs] -- [pathspec]` but we're passing the `--oneline` option before the `branch..branch` part.  Please adjust it so that the options come before the ref arguments.

Typically when using `git log` we try to always keep the double-dash at the end, even when we're not specifying any pathspecs, since it serves to disambiguate refs from paths.

cola/widgets/branch.py
```diff
@@ -0,0 +1,520 @@
...
+        return self.git.pull(remote, branch, no_ff=True, verbose=True)
+
+    def delete_remote(self, remote, branch):
+        return self.git.push(remote, branch, delete=True)
+
+    def delete_local(self, branch):
+        return self.git.branch(branch, D=True)
+
+    def merge(self, branch):
+        return self.git.merge(branch, no_ff=False, no_commit=True)
```
davvid: If you say `no_ff=False` then it effectively does nothing -- no flag is every passed to git.  If you want the `--no-ff` flag to make it through, set it to `True`.

cola/widgets/branch.py
```diff
@@ -0,0 +1,520 @@
...
+
+    def rename(self, branch, new_branch):
+        return self.git.branch(branch, new_branch, M=True)
```
davvid: Do we really want `-M` instead of `-m`?  I tend to prefer the safer form.

test/branch_test.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python
```
davvid: Thanks for these unit tests!

cola/widgets/branch.py
```diff
@@ -0,0 +1,520 @@
...
+        remote_branch = gitcmds.tracked_branch(branch)
+
+        if remote_branch is not None:
+            rgx = re.compile(r'^(?P<remote>[^/]+)/(?P<branch>.+)$')
+            match = rgx.match(remote_branch)
+
+            if match:
+                remote = match.group('remote')
+                branch_name = match.group('branch')
+                self.git_action_async('fetch', [remote, branch_name])
```
javierrodriguezcuevas: With this command I want to update the selected branch with his ref only if its has a tracked branch

javierrodriguezcuevas:
Thanks for your comments!
I'm happy you like the widget, I will work on them to merge as soon as possible.

cola/widgets/branch.py
davvid: I forgot to mention about the argument passing.. if you allow passing through both `*args` and `**kwargs` then you can let the Git class handle the option vs. argument stuff for you.

By passing the `--oneline` flag as `oneline=True` to the `log()` method lets the class handle the option vs. argument stuff.  That's probably easier to handle from the caller rather than trying to rearrange how the flags are passed.

javierrodriguezcuevas:
Hi David!

I made the changes you suggested to me, but when I tried to rebase and squash the commits I fail, I've never used this command before.
I created a new branch in my repository with the name 'feature/branches_widget' with only one commit. Could be possible to close this pull request and I make another from the new branch?

Thanks!

davvid:
I merged the `feature/branches_widget` branch and made a few small fixups.  Thanks again!

javierrodriguezcuevas:
I was looking at them right now and I find the optional icons very good idea!! Thanks for the merge!

PR #669, opened 2017-03-06
fix \\ No newline at end of file

xieofxie:
It happens when git diffs file without newline at end of file

### DIFF
--- a/cola/diffparse.py
+++ b/cola/diffparse.py
@@ -130,6 +130,7 @@ class DiffLines(object):
         DIFF_STATE = 1
         state = INITIAL_STATE
         merge = self.merge = False
+        NO_NEWLINE = '\\ No newline at end of file'
 
         old = self.old.reset()
         new = self.new.reset()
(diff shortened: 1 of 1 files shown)


cola/diffparse.py
```diff
@@ -155,7 +155,7 @@ def parse(self, diff_text):
                     new.parse(parts[3][1:])
                     lines.append((self.DASH, self.DASH, self.DASH))
                     continue
-            if state == INITIAL_STATE:
+            if state == INITIAL_STATE or text == '\\ No newline at end of file':
```
davvid: Move the `\\ No newline ...` string contents into a "constant" variable called `NO_NEWLINE`, which would be consistent with another similar usage in this file.  That'll make this line a little easier to read as `if state == INITIAL_STATE or text == NO_NEWLINE:`

cola/widgets/diff.py
```diff
@@ -195,6 +195,7 @@ def __init__(self, parent):
 
     def set_diff(self, diff):
         parser = self.parser
+        parser.valid = True
```
davvid: When reading this diff my first reaction was that this seemed wrong.  But, it actually is the correct state change, but this `parser.valid = True` should actually be done at the very beginning of `def parse()` as `self.valid = True`.  That way the caller here does not need to deal with it.

xieofxie:
A more aggressive fix.. By not setting it to false, Modified and Untracked buttons in Status can also show line numbers now. But Unmerged button is incorrect because the input is using @@ not @@@.

PR #670, opened 2017-03-12
Add Ukrainian translation file

kopach:
Signed-off-by: Igor Kopach <ikopach@ukr.net>


kopach:
Hi guys. I'm not familiar with the python, but I use git-cola daily basis, so decided to translate it into Ukrainian. If adding `po/uk.po` is not enough to do so - please let me know.

davvid:
I tested it by running `env LANG=UK_ua ./bin/git-cola` and it looked to be working.  Awesome, Thank you!

PR #672, opened 2017-03-15
Update ukrainian translation

kopach:
Signed-off-by: Igor Kopach <ikopach@ukr.net>


kopach:
After some more time of usage - updated translation

PR #675, opened 2017-03-25
return abspath in read_git_file for submodule

xieofxie:
should return an absolute path if a relative path is in the .git file of a submodule

### DIFF
--- a/cola/git.py
+++ b/cola/git.py
@@ -67,6 +67,9 @@ def read_git_file(path):
         data = core.read(path).strip()
         if data.startswith(header):
             result = data[len(header):]
+    if result and not os.path.isabs(result):
+        path_folder = os.path.dirname(path)
+        result = os.path.join(path_folder, result)
     return result
 
 


cola/git.py
```diff
@@ -67,6 +67,9 @@ def read_git_file(path):
         data = core.read(path).strip()
         if data.startswith(header):
             result = data[len(header):]
+    if result and not os.path.isabs(result):
+        path_folder = os.path.dirname(path)
+        result = os.path.join(path_folder, result)
```
davvid: I think we'd only want to do this inside the same `if data.startswith(header)` block, since that's the only one which ever sets `result`, so this seems like it should be indented in two levels as a starting point.

Do you happen to have a reproduction recipe showing the problem?  I'd like to test this on a test repo locally, maybe as a starting point for a test case.

The commit message should explain why we would want to return an absolute path rather than echoing what the code is already saying; it's better to say something that the code cannot.

The `Ctrl+m` amend mode in cola's commit editor can be used to revise the commit.

Issue #586, opened 2016-08-07
Toolbar for often used functions


guoyunhe:
I found I use fetch, pull, push functions very often. But using menu is not very convenient. I know here is a widget of action buttons, but the position is not good, button size is small in small window, no icon:

![spectacle hfb938](https://cloud.githubusercontent.com/assets/5836790/17463209/d11e6268-5cc9-11e6-9734-6fdeb2b7f4ca.png)

It is better to add an toolbar like normal KDE and GNOME applications.

davvid:
I just realized that this feature has already been implemented ~ thanks @javierrodriguezcuevas !

PR #678, opened 2017-04-06
Feature/toolbar

javierrodriguezcuevas:
Hi again!

I was thinking about how to make the actions widget more configurable and saw the issue #586 so I created a customizable toolbar.
As I did not know how many actions to add, I added all of the menubar, but not all actions have icon.

In action:
peek 06-04-2017 22-34


davvid:
This look amazing.  I'll tag the next stable version this weekend and start taking a look at this since we just finished a stabilization round.  This looks really cool.

The small additional things we'd probably want (not sure if it's already there) would be ability to toggle the toolbar and persist that state, since some users will have limited screen space and might prefer the minimalist view.

The toolbar seems like it should be a full-on replacement for the "Actions" widget, as it has a much nicer interaction overall.  Can the toolbar be docked left, right, top, or bottom?

I removed the Actions widget briefly at one point in the past.  Several folks quickly asked that we restore it, but my sense is that this toolbar would be a fine replacement for the Actions widget as long as it has some flexibility in how it can be displayed.  `QMainWIndow::addToolBar(...)` allows setting the area (top, right, bottom, left), so that's probably sufficient control for our needs.

I would probably have the toolbar at the bottom or right.  Using icons will probably let it be slimmer, which was always a pet peeve of mine about the Actions widget.  It never had a nice small profile.  The toolbar does, which is really good.

cola/widgets/toolbar.py
```diff
@@ -0,0 +1,527 @@
...
+ACTION_REBASE = 100
+ACTION_REBASE_EDIT = 101
+ACTION_REBASE_CONT = 102
+ACTION_REBASE_IGNORE = 103
+ACTION_REBASE_ABORT = 104
+
+
+SEPARATOR_ACTION = {'title': SEPARATOR_TEXT, 'icon': None, 'action': None}
+# We assume that a child key defined in TOOLBAR_ACTIONS has a key defined
+# with the same name in TOOLBAR_TEXTS
```
davvid: It would be good to avoid all of these globals, and we're doing too much during module import.

cola/widgets/toolbar.py
```diff
@@ -0,0 +1,527 @@
...
+ACTION_BROWSE_CURRENT = 86
+ACTION_BROWSE_OTHER = 87
+ACTION_VISUALIZE_CURRENT = 88
+ACTION_VISUALIZE_ALL = 89
+
+ACTION_REBASE = 100
+ACTION_REBASE_EDIT = 101
+ACTION_REBASE_CONT = 102
+ACTION_REBASE_IGNORE = 103
+ACTION_REBASE_ABORT = 104
```
davvid: If these various IDs were moved into a class then we could have a helper function or generator that increments a counter, so that if we add/remove actions we won't have to worry about renumbering them.

Or, do these numbers need to be persistent and never change, because they appear in the config?  It might be good to mention whether these are purely internal.

cola/widgets/toolbar.py
```diff
@@ -0,0 +1,527 @@
...
+    },
+    N_('Rebase'): {
+        ACTION_REBASE: lambda: app().activeWindow().rebase_start(),
+        ACTION_REBASE_EDIT: lambda: app().activeWindow().rebase_edit_todo(),
+        ACTION_REBASE_CONT: lambda: app().activeWindow().rebase_continue(),
+        ACTION_REBASE_IGNORE: lambda: app().activeWindow().rebase_skip(),
+        ACTION_REBASE_ABORT: lambda: app().activeWindow().rebase_abort()
+    }
+}
+TOOLBAR_TEXTS = {
```
davvid: All of the `N_(...)` calls here happen during module import, which is bad because we may or may have not yet setup the i18n stuff, so we should avoid anything related to that here.

The icons circular import dependency also needs to be eliminated.  The inline imports in `cola.main` are a sign that we need to refactor.

More importantly, it seems like the translated strings get written to the config, which would not be good. Is it instead that the integers get written?

cola/widgets/main.py
```diff
@@ -76,6 +76,10 @@ def __init__(self, model, parent=None, settings=None):
         # Runs asynchronous tasks
         self.runtask = qtutils.RunTask()
 
+        from cola.widgets.toolbar import create_toolbar
```
davvid: Let's figure out why we have to defer this import.  If you pull it out to where the others are, what traceback do you see?

cola/widgets/toolbar.py
```diff
@@ -0,0 +1,527 @@
...
+        ACTION_OPEN_NEW_WIN: lambda: guicmds.open_repo_in_new_window(),
+        ACTION_CLONE: lambda: app().activeWindow().clone_repo(),
+        ACTION_REFRESH: cmds.run(cmds.Refresh),
+        ACTION_FIND_FILES: lambda: finder.finder(),
+        ACTION_EDIT_REMOTE_REPO: lambda: editremotes.remote_editor().exec_(),
+        ACTION_RECENT_EDIT: lambda: recent.browse_recent_files(),
+        ACTION_APPLY_PATCHS: lambda: patch.apply_patches(),
+        ACTION_EXPORT_PATCHS: lambda: guicmds.export_patches(),
+        ACTION_SAVE_TAR_ZIP: lambda: app().activeWindow().save_archive(),
+        ACTION_PREFERENCES: lambda: app().activeWindow().preferences()
```
davvid: I think any time we see `app().activeWindow().XXX()` it means we should refactor.  We should not have that dependency here.

cola/widgets/toolbar.py
```diff
@@ -0,0 +1,527 @@
...
+
+def load_toolbar_config():
+    prefs_model = prefs.PreferencesModel()
+    return json.loads(prefs_model.get_config(SOURCE_CONFIG, PREFS_TOOLBAR))
+
+
+def save_toolbar_config(actions_dict):
+    prefs_model = prefs.PreferencesModel()
+    cmds.do(prefs.SetConfig, prefs_model, SOURCE_CONFIG, PREFS_TOOLBAR,
+            json.dumps(actions_dict))
```
davvid: Loading and saving the toolbar config here is slow because we have to touch the filesystem (again).  We shouldn't do that because we already have code paths that save and restore "settings", and we should leverage that.

Also, this assumes that we only have a single toolbar, which is not necessarily true.  `git-dag` might have its own.  It's better to tie the save/restore to the main widget's save/restore, and have it push values to the toolbar instances that it owns.

The functions in the main widget related to saving state are `export_state()` and `apply_state()`.

cola/widgets/main.py
```diff
@@ -76,6 +76,10 @@ def __init__(self, model, parent=None, settings=None):
...
+        self.toolbar = self.addToolBar('Main Toolbar')
+        create_toolbar(self.toolbar)
```
davvid: Instead of a `Main Toolbar` it might be simplest to make the code reusable and then have `Top Toolbar`, `Bottom Toolbar`, `Left Toolbar`, and `Right Toolbar`, and reuse as much as possible.

They would all default to hidden.

cola/widgets/toolbar.py
javierrodriguezcuevas: These IDs should not be changed since they are used to save the toolbar config. I'm thinking now to validate that they exists in TOOLBAR_ACTIONS before inserting the action to prevent errors if we remove one that a user had configured. I also will add a comment about not changing their values.

cola/widgets/toolbar.py
javierrodriguezcuevas: I will take a look to refactor these functions to a class or helper.

cola/widgets/toolbar.py
javierrodriguezcuevas: I will try to remove inline imports. Yes integers are written in configuration instead translated strings.

cola/widgets/main.py
javierrodriguezcuevas: If I put with the others it gives me the following error with all the icons.

`Cannot open file '/home/ubuntu/python/git-cola/bin/icons:folder-new.svg', because: No such file or directory`

javierrodriguezcuevas:
About position, the toolbar can be docked to left, right, top, and bottom. The user can drop it wherever he wants.

davvid:
This is looking good.  I'll start testing it tonight and try to merge shortly now that we're open for v3.0 alpha changes.  I didn't realize the single toolbar was draggable to any location, that's great.

javierrodriguezcuevas:
The refactoring of `app().activeWindow().XXX()` is still pending. 
About constant 'MAIN_ACTIONS' now it can be converted to an object easily, but it has to be passed to `ColaToolBar` as an argument and I don't know how to make it dynamic when loading configuration.

cola/widgets/main.py
```diff
@@ -692,6 +717,13 @@ def apply_state(self, state):
         diff_numbers = state.get('show_diff_line_numbers', False)
         self.diffeditor.enable_line_numbers(diff_numbers)
 
+        toolbar_data = state.get('toolbar', '[]')
```
davvid: The default value here must be `[]` instead of the `'[]'` string.. just ran into a traceback when switching into this branch from the state where the settings file does not have this data.

cola/widgets/main.py
```diff
@@ -679,6 +694,16 @@ def export_state(self):
         show_status_filter = self.statuswidget.filter_widget.isVisible()
         state['show_status_filter'] = show_status_filter
         state['show_diff_line_numbers'] = self.diffeditor.show_line_numbers()
+
+        saved_toolbars = []
+        toolbars = self.findChildren(ColaToolBar)
+        for toolbar in toolbars:
+            saved_toolbars.append({'name': toolbar.windowTitle(),
+                                   'area': self.toolBarArea(toolbar),
+                                   'config': json.dumps(toolbar.get_config())})
```
davvid: I think it would be better to not `json..dumps()` here.. this ends up with this in the final settings file:

```js
            "toolbar": [
                {
                    "config": "{\"data\": [{\"parent_id\": 122, \"child_id\": 40}, {\"parent_id\": 122, \"child_id\": 44}, {\"parent_id\": 123, \"child_id\": 60}, {\"parent_id\": 123, \"child_id\": 61}, {\"parent_id\": null, \"child_id\": 130}, {\"parent_id\": 121, \"child_id\": 20}, {\"parent_id\": 121, \"child_id\": 22}, {\"parent_id\": 121, \"child_id\": 21}], \"show_icons\": true}", 
                    "name": "Main", 
                    "area": 4
                }
            ]
```

note how there's embedded json inside that string.  It would be better to let this be the real raw value instead so that we don't have to compensate for it when reading and writing.  Down below we do `json.loads(...)` and we won't need to do that anymore.

If you make those changes you'll probably need to delete the toolbars from your settings, as it'll make the format incompatible for your current settings.

cola/widgets/toolbar.py
```diff
@@ -0,0 +1,670 @@
...
+from cola.widgets import merge
+from cola.widgets import patch
+from cola.widgets import recent
+from cola.widgets import remote
+from cola.widgets import search
+from cola.widgets import standard
+from cola.widgets import stash
+from qtpy import QtGui
+from qtpy.QtCore import Qt
+from qtpy import QtWidgets
```
davvid: qtpy imports should come before cola here (separated by a newline) and we should probably prefer relative imports these days.  We've eliminated all of the non-essential absolute cola imports so new code should follow that style

cola/widgets/toolbar.py
```diff
@@ -0,0 +1,670 @@
...
+
+        if len(items) > 0:
+            top = items[0]
+            child = self.insert_child(top.data(Qt.UserRole), child_id, title,
+                                      icon)
+            top.appendRow(child)
+            top.sortChildren(0, Qt.AscendingOrder)
+
+
+MAIN_ACTIONS = {
```
davvid: It would be nice if there was an action that ran the `Edit` command, so that the selected files are opened in the editor (same as the Ctrl+e hotkey).  I went looking for it and didn't see an entry for it, though that's something I use pretty often (I might be an atypical user).

Another notable missing item was "Difftool" from the Diff section (same as the Ctrl-d hotkey) so that we can diff the selected files.

I would also recommend making it so that the "Configure toolbar" dialog remembers its window geometry.  I keep having to resizing it everytime I open it so reusing the apply_state() code can do that for us so that it'll remember its size and position.

davvid:
(apologies for the multiple comments.. I should have used the "start review" button -- last note for now)

I noticed that if you drag a toolbar off of the main UI it'll be free-floating, which is fine.  But if you save and then restart cola it'll print this error:  `QMainWindow::addToolBar: invalid 'area' argument` and the free-floating state is not saved, which might be unexpected.

`invalid 'area' argument` error is not 100% reproducible, but I was able to trigger it a few times by trying a few different things involving having two toolbars with the same name in the same area, and then dragging it in and out of the UI, in case that helps you reproduce it.

Since there are only a few different areas it might not hurt to make avoid exposing Qt's raw value for the toolbar area and intern it into a string that we know about in our code.  That way we can handle the out-of-band values in one place and never have to worry about handing over an invalid `area` value when constructing the UI.

The UI construction and settings code should be especially paranoid, as it should be prepared to deal with completely broken and invalid data, so adding a layer of indirection between the `toolbarArea(..)` value and what we actually store/restore would be a good idea.

My last meta-note is that the resulting settings file is a little bit terse and hard to understand, notably because of the magic numbers that are in the file.

What do you think about having a separate, dependency-free file (similar to the cola/widgets/defs.py flie) that contains just constants that we can use to "name" the various actions.

For example, the `Rebase->Edit` command can be called "Rebase::Edit" in the settings file, instead of `101` with parent ID `125`.

Right now the `MAIN_ACTIONS` dict in the toolbars is doubling as both the definition for the UI hierarchy and as a way to find the recipe for running the commands.

Right now we're storing the ID and parent ID so that we can index into the actions dict.  It'd be better to have the UI definition be its own thing, and bind them together by having unique leaf names for each command.  e.g.

```python

COMMANDS = {
   'Rebase::Edit': {
        'title': ...,
        'action': ...,
        'icon': ...,
   },
   ...
}

TREE_LAYOUT = {
    'File': [ ... ],
    'Rebase': [
        'Rebase::Edit',
        'Rebase::Continue',
        ...
    ],
   ...
}
   ...
}
```

The benefit of doing this is that the settings file only says `Rebase::Edit`, and it doesn't contain integers, and it's generally kinda still readable and human-editable.

Also, it can isolate the `COMMANDS` dependency, which could then be refactored to a separate file, etc.

cola/widgets/toolbar.py
```diff
@@ -0,0 +1,670 @@
...
+
+        self.setModel(QtGui.QStandardItemModel())
+
+    def create_item(self, name, data):
+        item = QtGui.QStandardItem()
+
+        item.setEditable(False)
+        item.setDragEnabled(True)
+        item.setText(N_(name))
+        item.setToolTip(N_(name))
```
davvid: After using this for a few days, the tooltip here is pretty distracting, specifically because the tooltip does not give any extra information that is not already present in the name itself.  Let's remove this tooltip.

cola/widgets/main.py
javierrodriguezcuevas: :+1:

cola/widgets/toolbar.py
javierrodriguezcuevas: I will add those commands. What do you think if we remove the commands that need refactoring `app().activeWindow().XXX()` while not working on it?

cola/widgets/toolbar.py
javierrodriguezcuevas: I think the same. I will work on the refactoring of the tree and try to reproduce the issues.

javierrodriguezcuevas:
Hi, I worked on the previous comments. 
The problem with `invalid 'area' argument` I found when removing a toolbar, it is assigned `Qt.NoToolBarArea` and later when is used in `self.addToolBar()` throws the error. 
I also worked on the imports, but I'm not sure if it's ok now. 
The object `ToolbarManager` is what you meant about adding a layer of indirection? I can add a restore value method, but the `addToolBar` and `addToolBarBreak` method is in the `MainView` and float must be set after insert the toolbar, so the only solution I find is return a dict or object with the toolbar, break, float and positions to be used in the `MainView`.
I updated my SO and now when an element from left list is dropped on right list the element is not removed from left list and then when is removed from right list is added on the left, so I will prevent duplicated elements in tree, but I will try with `DraggableTreeWidget` or find another solution to remove the element when dropped.

PR #681, opened 2017-04-14
for submodule's folder,if stage with /,stage all files in it in windows

xieofxie:
After adding a submodule, its folder will be shown in untracked. However, git add folder/ will add all files in folder in windows and git add folder will only add the submodule's folder. Or one can do this check in status.py, but I think in gui, adding a folder will never happen because it always shows untracked or modified files one by one

### DIFF
--- a/cola/models/main.py
+++ b/cola/models/main.py
@@ -467,8 +467,8 @@ class MainModel(Observable):
 
         for path in set(paths):
             if core.exists(path):
-                if path[-1:] == '/':
-                    path = path[:-1]
+                if path.endswith('/'):
+                    path = path.rstrip('/')
                 add.append(path)
             else:
                 remove.append(path)
--- a/cola/models/main.py
+++ b/cola/models/main.py
@@ -467,6 +467,8 @@ class MainModel(Observable):
 
         for path in set(paths):
             if core.exists(path):
+                if path[-1:] == '/':
+                    path = path[:-1]
                 add.append(path)
             else:
                 remove.append(path)


cola/models/main.py
```diff
@@ -467,6 +467,8 @@ def stage_paths(self, paths):
 
         for path in set(paths):
             if core.exists(path):
+                if path[-1:] == '/':
+                    path = path[:-1]
```
davvid: A slightly more idiomatic python way to write this might be,

```python

if path.endswith('/'):
    path = path.rstrip('/')
```

They're both just about equivalent, except the `rstrip()` will consume multiple `/`, but that doesn't matter since we never see those in practice.

cola/models/main.py
```diff
@@ -467,6 +467,8 @@ def stage_paths(self, paths):
```
davvid: When creating commits please sign off your work, per https://github.com/git-cola/git-cola/blob/master/CONTRIBUTING.md#sign-your-work

In the commit message editor we can use the `Ctrl+i` hotkey to add the Signed-off-by line.  You may need to adjust your config to include your full name.

davvid: I forgot to mention that it would be good to include some of the details that were mentioned in the pull request comment in the commit message's `Extended description` since the actual commit didn't include any of these details.

I tend to merge PRs via cola or git itself rather than github, so it's better to include those details in the commits rather than in the github comments.

xieofxie: add sign-off & example

PR #688, opened 2017-04-21
Revert unstaged/uncommited edits message

4ndrej:
The revert unstaged/uncommited edits message clarification.

The original message was not specific about the reverted files, it was not clear whether it's related to selected file(s) or all changed files.


davvid:
The precision of the new language is good, but weighed against the i18n effort,  I'm not sure it's worth changing.  @Vdragon do you  have a preference?

PR #702, opened 2017-05-27
get submodule list via git submodule

xieofxie:
Maybe GitKraken uses .gitsubmodule file as well.
Is this the correct way to get the list?
By the way, I was wondering if it is possible to add a feature like GitKraken's commit tree to dag. It seems that it combines all branches & remotes together to make it more visually clear.


davvid:
Consulting .gitmodules does seem to be the core way to get at the submodule data, while the `git submodule` command can give you state that .gitmodules cannot.

I don't think `git submodule`'s output is necessarily considered part of Git's `plumbing` interface, but it should be stable enough for our needs.

Issue #691, opened 2017-05-01
Allow autocompletion in the Name field of "Create Tag" window


Vdragon:
It would be useful to select the latest tag from the autocomplete result, then increment the digit you wish.  
Currently you'll have to close the "Create Tag" window, check out the latest tagged version from DAG or other method, and open the "Create Tag" window again.

Another possible solution might be allow moving focus to the main window while "Create Tag" window is opened, so I can open DAG window without closing the "Create Tag" window

PR #706, opened 2017-06-15
createtag: autocompletition on tag_name field

judaszszymon:
Add completer to tag_name in createtag.py so user is informed about
existing tags.

Reported-by: @Vdragon #691
Signed-off-by: Szymon Judasz szymon.judasz@student.uj.edu.pl

### DIFF
--- a/cola/widgets/createtag.py
+++ b/cola/widgets/createtag.py
@@ -12,6 +12,7 @@ from . import defs
 from . import completion
 from . import standard
 from . import text
+from ..models import main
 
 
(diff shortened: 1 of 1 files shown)


davvid:
Thanks, looks good, I'll merge this shortly.

If we wanted to take it even further, creating a sibling to `GitRefLineEdit` and `GitRefCompletionModel` in `cola/widgets/completion.py` that implements `def matches()` and returns only tags in the `GitTagCompletionModel`.  Then swap the `HintedLineEdit` for the new `GitTagLineEdit`.

That'll get us the nice icons for free, but this works for now.

PR #713, opened 2017-07-19
branch: add filter widget

javierrodriguezcuevas:
Hi!
This is a simple filter that hightlight the branches containing the text on it.

Signed-off-by: javierrodriguezcuevas <javierrodriguezcuevas@gmail.com>


javierrodriguezcuevas:
I discarded the use of 'Highlight' because in my case it is the same color as when I select a branch. Making the name of the branch bold seems even better than using the color palette for me. I Will make the change this weekend.

davvid:
Awesome

PR #714, opened 2017-07-27
qtutils: use the proper clicked signal instead of pressed

### LINKED ISSUE #604: TypeError: __init__() takes 1 positional argument but 2 were given
```
  File "/usr/share/git-cola/lib/cola/cmds.py", line 1835, in runner
    do(cls, *local_args, **local_opts)
  File "/usr/share/git-cola/lib/cola/cmds.py", line 1856, in do
    return do_cmd(cls(*args, **opts))
TypeError: __init__() takes 1 positional argument but 2 were given
error: git-cola died of signal 6
```

When clicking `Refresh` action button

jeff-dagenais:
Reacting to the pushed event makes the interface behave unlike any other
GUI out there. Restore the normal clicked behaviour. Fix #604 using a
bounce function.

This reverts d9a3f3f8f84d15723dd10ded2adeb6d9ffeb1fcd while still fixing

### DIFF
--- a/cola/qtutils.py
+++ b/cola/qtutils.py
@@ -46,7 +46,7 @@ def connect_action_bool(action, fn):
 
 def connect_button(button, fn):
     """Connect a button to a function"""
-    button.pressed.connect(fn)
+    button.clicked.connect(fn)
 
 
 def connect_released(button, fn):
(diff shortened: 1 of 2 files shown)


jeff-dagenais:
Sorry, I have to rework this a bit. Will resubmit later.

jeff-dagenais:
Ok, I only had a problem in my commit message. My reference to #604 was being wrapped and happened to fall at the start of the line, and thus disappeared since git thought it was a commented line.

PR #715, opened 2017-07-27
qtutils: use the proper clicked signal instead of pressed

### LINKED ISSUE #604: TypeError: __init__() takes 1 positional argument but 2 were given

jeff-dagenais:
Reacting to the pushed event makes the interface behave unlike any other
GUI out there. Restore the normal clicked behaviour. Fix #604 using a
bounce function.

This reverts d9a3f3f8f84d15723dd10ded2adeb6d9ffeb1fcd while still
fixing #604

### DIFF
--- a/cola/qtutils.py
+++ b/cola/qtutils.py
@@ -46,7 +46,7 @@ def connect_action_bool(action, fn):
 
 def connect_button(button, fn):
     """Connect a button to a function"""
-    button.pressed.connect(fn)
+    button.clicked.connect(fn)
 
 
 def connect_released(button, fn):
--- a/cola/widgets/action.py
+++ b/cola/widgets/action.py
@@ -50,6 +50,9 @@ def tooltip_button(text, layout):
     button.setToolTip(text)
     return button
 
+""" Takes care of argument mismatch for connect_button event"""
+def RefreshButtonAction():
+    return cmds.run(cmds.Refresh)
 
 class ActionButtons(QFlowLayoutWidget):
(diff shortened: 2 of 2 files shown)


jeff-dagenais:
Sorry for the naive patch.
1. I am no python champion yet,
2. Although I have been using cola for 6-7 years, I have very little knowledge in the source code.

What I do know is that using GUI apps which react immediately on button pressed event hurts the brain. Feels like my mouse is broken... or is it my brain... 

So with the provided suggestion about the `def wrapper...` I can give it a whirl and re-submit.

davvid:
Sweet, thanks for the careful attention to these usability issues, I believe these changes will be well received all around, and wholly agree that user experience is more important than code simplicity.

In case it helps, here's roughly what I was describing (I guess this is where I should have added a comment about the previous choice)

```
def connect_button(self, button, fn):
    def wrapper(*args, **kwargs):
        """Consume clicked's callback arguments to handle differences in behavior between Qt 4 and 5"""
        fn()
    button.clicked.connect(wrapper)
```

Looking forward to it, and thanks again!

PR #738, opened 2017-10-20
Freezing without git

laerreal:
git-cola freezes when `git` is not found in `PATH`. The cause is a deadlock on `INDEX_LOCK`. The mutex is not released after an exception. I faced with this under Windows but the problem is probably actual for all the OSes.

This patch request elimenates the deadlock and also fixes checking for git existence under Windows.

### DIFF
--- a/cola/git.py
+++ b/cola/git.py
@@ -234,13 +234,15 @@ class Git(object):
         # Guard against thread-unsafe .git/index.lock files
         if not _readonly:
             INDEX_LOCK.acquire()
-        status, out, err = core.run_command(
-                command, cwd=_cwd, encoding=_encoding,
-                stdin=_stdin, stdout=_stdout, stderr=_stderr,
-                no_win32_startupinfo=_no_win32_startupinfo, **extra)
-        # Let the next thread in
-        if not _readonly:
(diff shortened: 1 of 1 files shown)


laerreal:
The problem seems to be quite old. I was tried to use 2.10 before debugging of master. Git Cola 2.10 behaves similarly.

cola/git.py
```diff
@@ -330,9 +332,14 @@ def git(self, cmd, *args, **kwargs):
                 # case of argv overflow. we should be safe from that but use
                 # defensive coding for the worst-case scenario. on other OS-en
                 # we have ENAMETOOLONG which doesn't exist in with32 API.
-                status, out, err = self.execute(['git', '--version'])
-                if status == 0:
-                    raise e
+                try:
+                    status, out, err = self.execute(['git', '--version'])
+                except FileNotFoundError:
+                    pass # git command is not available
```
davvid: `FileNotFound` is a python3-ism; it's preferable to catch `OSError` to not break python2 users.  I'll merge this shortly with a fixup on top.  Thanks for the fix, indeed, this is a problem on all OSes.

Issue #96, opened 2011-10-28
recognize .gitattributes, at least "encoding" field


socketpair:
I have .gitattributes with this:

> *.sql encoding=cp1251

So, in git GUI, "interactive add" woks out of the box, but not in git cola, as I have russian lines.
Also, git cola displays wrong symbols in these lines.

PR #745, opened 2017-11-09
Fixup non-UTF-8 line staging

laerreal:
Recently I have faced the problem, I cannot stage selected lines in several files. This related to lines those cannot be decoded using default `"utf-8"` coder. The core of the problem can be easily seen by data flow description presented below.

1. Git-Cola uses `git` utility to obtain diff text. This utility is launched using `subprocess.Popen` (see `cola.core.start_command` function).

2. Then, diff text is taken from `stdout` returned by `communicate` method (see `cola.core.run_command` function). `run_command` uses `cola.core.decode` to decode raw data returned by `git` to unicode (`str` under Py3 or `unicode` under Py2).

The problem begins at this step. `decode` is given an `encoding` argument. This encoding is used to decode those raw data. If it failed then default encodings are probed one by one (see `cola.core._encoding_tests` list).

In my case, `encoding` argument is set to `"utf-8"` because of it is default and no encoding has been explicitly set for the file. But real encoding of the file is cp1251 (Cyrillic). A file encoded in cp1251 frequently contains bytes those cannot be decoded using `"utf-8"`. As a result, first "successful" decoding happens for `'iso-8859-15'`. To be continued...

3. After, those data are passed unchanged through several wrappers towards a `DiffParser` object. Given line indices, the parser formats a patch changing selected lines. See `do` method of `cola.cmds.ApplyDiffSelection`

4. Next, the patch is written to a temp file. In course of it, the patch is encoded using either explicitly set encoding or default `'utf-8'` (for me).

Now, data originally decoded in `'iso-8859-15'` are encoded in `'utf-8'`. Non-ANSII characters are encoded to different bytes.

5. Finally, Git Cola attempts to apply the patch using `git apply` command. It failed because removed lines in the patch is not equal to themselves in the file.

I suggest to use `"raw_unicode_escape"` for both diff text and patch files. This coder seems to be able to decode any byte in raw data and encode it back unchanged. Non-UTF-8 lines in diff widget are displayed differently but remain unreadable.

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -250,11 +250,12 @@ class ApplyDiffSelection(Command):
         if patch is None:
             return
 
-        cfg = gitcfg.current()
-        encoding = cfg.file_encoding(self.model.filename)
         tmp_file = utils.tmp_filename('patch')
         try:
-            core.write(tmp_file, patch, encoding=encoding)
+            # Patch encoding is always "raw_unicode_escape" because
+            # model.diff_text is always same. See method `git` of
+            # class `cola.git.Git`.
+            core.write(tmp_file, patch, encoding="raw_unicode_escape")
             if self.apply_to_worktree:
                 status, out, err = self.model.apply_diff_to_worktree(tmp_file)
             else:
--- a/cola/git.py
+++ b/cola/git.py
@@ -312,6 +312,10 @@ class Git(object):
                 '_no_win32_startupinfo',
                 )
 
+        if cmd == "diff":
+            # use invariant encoding for diff
+            kwargs["_encoding"] = "raw_unicode_escape"
+
         for kwarg in execute_kwargs:
             if kwarg in kwargs:
                 _kwargs[kwarg] = kwargs.pop(kwarg)
--- a/cola/gitcmds.py
+++ b/cola/gitcmds.py
@@ -359,18 +359,14 @@ def diff_helper(commit=None,
     elif head and amending and cached:
         argv.append(head)
 
-    encoding = None
     if filename:
         argv.append('--')
         if type(filename) is list:
             argv.extend(filename)
         else:
             argv.append(filename)
-            cfg = gitcfg.current()
-            encoding = cfg.file_encoding(filename)
 
     status, out, err = git.diff(R=reverse, M=True, cached=cached,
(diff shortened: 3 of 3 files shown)


laerreal:
Oops, it breaks appearance non-ASCII characters in UTF-8 encoded files too.
A custom coder is probably needed. A one that pass UTF-8 characters to standard utf-8 encoder and escapes only non-UTF-8 characters.

```
# bytes to unicode
b'\'      ->    \\
UTF-8     -> "utf-8"
non-UTF-8 -> \xXX

# unicode to bytes
\\        -> b'\'
\xXX      -> b'\xXX'
other     -> "utf-8"
```

davvid:
Thanks for the heads-up.

We should support this via `.gitattributes` and the `gui.encoding` variable[1].

If this is broken when using those features please let me know.

Nonetheless, we might be able to use the `surrogateescape` encoder, but that's a guess, and might not work at all.  Being able to preserve bytes is a noble goal.

Would it be possible to share a public test repo that demonstrates the problem in case the staging still doesn't work when using those features?  I'll see if I can dig up the test data from #96 (a long time ago ;-))

[1] https://www.git-scm.com/docs/gitattributes#_viewing_files_in_gui_tools

laerreal:
2017-11-15 8:55 GMT+03:00 David Aguilar <notifications@github.com>:
>
> Thanks for the heads-up.
>
> We should support this via .gitattributes and the gui.encoding variable[1].
>
> If this is broken when using those features please let me know.
>
> Nonetheless, we might be able to use the surrogateescape encoder, but that's a guess, and might not work at all. Being able to preserve bytes is a noble goal.

Yes. I think, original data have to be preserved internally as long as
possible. Ideally, a decoding attempt should be made for only
displaying in a widget.
Normally, I don`t need to be able to read lines I want to stage.
Because, I can identify the changes. I just want to stage them
technically.
Using gui.encoding seems to be helpful. But what if the intent of a
user is to change the encoding...

laerreal:
I just push the branch [encodings](https://github.com/laerreal/git-cola/commits/encodings) to my clone of git-cola. There are two files with a Cyrillic text in utf-8 and cp1251 encoding. You may just delete some lines and try to stage those lines. This reproduces the problem for me.

Issue #711, opened 2017-07-04
Cannot choose directory in File->Export patches


wojnilowicz:
It would be nice if git cola had feature to configure directory into which it exports patches. Right now it's <REPOSITORY>/patches

davvid:
This would be good.  It's currently hard-coded to my workflow ;-)

Most of the plumbing is already in place to make it configurable -- the top-level `export_patcheset()` function already takes an output parameter, but we just haven't hooked anything up to it yet.  It would be nice to have it default to `patches` and allow the user to enter anything.

davvid:
Another thing that would be really cool is to add a "split patch" feature that lets you write patch files directly.

From the right-click menu:

```
Export (Hunk | Selection) to Patch
- New Patch
- Append Patch
  - Choose Patch File...
  - <list-of-existing-patches>
```

This can work with the selection and let you interactively create a patch file from your current edits.

Basically, the same idea as https://github.com/aleclearmind/split-patch but integrated into the diff editor.

This feature would also need to know about a designated directory for patches, so these are related.

PR #747, opened 2017-11-14
selectcommits: add output dir selector

javierrodriguezcuevas:
Hi! 

I was playing with 'Export Patches' option and I found that I could not choose the output directory as it is says in the issue #711, so I added a button in the 'Export Patches' dialog which opens a directory selector prompt.

git-cola_export-patches


davvid:
This has already been merged.  It may not have been reflected here because I may have rebased the commit to fix a 'Rebase stoppped' vs 'Rebase stopped' typo.

PR #751, opened 2017-11-15
RFC: Fixup non utf8 line staging

laerreal:
There is another approach to solution of problem I previously touched in #745. What do you think about it?

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -239,7 +239,9 @@ class ApplyDiffSelection(Command):
         self.apply_to_worktree = apply_to_worktree
 
     def do(self):
-        parser = DiffParser(self.model.filename, self.model.diff_text)
+        diff_text = self.model.diff_text
+
+        parser = DiffParser(self.model.filename, diff_text)
         if self.has_selection:
             patch = parser.generate_patch(self.first_line_idx,
                                           self.last_line_idx,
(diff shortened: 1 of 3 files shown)


laerreal:
There is a problem with non-ANSII characters in utf-8 files again... I will investigate it soon.

laerreal:
But git-cola confuses encoding if there are two different encodings in one diff. E.g. user changed encoding of its file. This is actual for current master too.

davvid:
This is a brilliant solution and seems to work pretty nicely.

I'll make a few tweaks on top of your branch and try to merge this in soon.  I don't think we should worry too much about files with multiple encodings -- that's a rare edge case.  We can always use the "Ctrl-S" hotkey (or double-click) to stage the entire file when an encoding change is performed over the whole file.

In my branch I'm also making it so that the `cola.fileattributes` config option is automatically enabled when a `.gitattributes` file is present.  I realized that I suggested using a  `.gitattributes` file to specify the encoding, but I forgot to mention that you also have to `git config cola.fileattributes true` to enable `.gitattribute` support.

`.gitattributes` is the most robust solution since the user tells us what encoding to use.  Example file:

```
test/cyrillic-cp1251.txt encoding=cp1251
```

With that in place I was able to stage the test file.  Just wanted to mention in case it wasn't working in your testing.  Nonetheless, this PR should still be merged since it's better all around.

PR #758, opened 2017-12-12
Spelling fixes

scop:


davvid:
FYI the `qtpy/` directory is from https://github.com/spyder-ide/qtpy in case you'd like to fix those typos upstream.  Thanks!

PR #793, opened 2018-04-08
 Update Czech translation

pavreh:


davvid:
Please avoid including "Merge" commits in topic branches.  I merged the new commit directly so this topic is all set, but please rebase or recreate the `czpo` branch for future submissions.

We recently changed the `po` files so that they're sorted and no longer include line numbers.  That should help minimize any merge conflicts in the future.  Hopefully that change didn't make things too tricky if updating from an old tree.

Thanks again!

PR #800, opened 2018-04-12
app: change date/time format in log widget

alex-gulyas:
The previous format displayed a weird date/time when the active locale
didn't have AM/PM (%p): `10:45  57s 2018-04-12`.

### DIFF
--- a/cola/widgets/log.py
+++ b/cola/widgets/log.py
@@ -53,7 +53,7 @@ class LogWidget(QtWidgets.QWidget):
         cursor.movePosition(cursor.End)
         text = self.output_text
         # NOTE: the ':  ' colon-SP-SP suffix in used by the syntax highlighter
-        prefix = core.decode(time.strftime('%I:%M %p %Ss %Y-%m-%d:  '))
+        prefix = core.decode(time.strftime('%x %X:  '))
         for line in msg.splitlines():
             cursor.insertText(prefix + line + '\n')
         cursor.movePosition(cursor.End)


alex-gulyas:
> I dunno, this changes the format string for folks that do have AM/PM, so I can't merge this as-is.

This is not correct, if your locale uses AM/PM so will this format:
- `LC_TIME=en_US.UTF-8`: 04/13/2018 04:15:05 PM
- `LC_TIME=hu_HU.UTF-8`: 2018-04-13 16:18:01

> That said, I don't really see what's so weird about the format string you pasted.

By weird I actually meant this (I should've explained it more in the PR, sorry):
- Because `%p` is empty for me, there will be two spaces between the minutes and seconds (markdown ate the 2nd space in my PR message)
- Even though `%p` is empty, the time is still showed as 12 hour based, so 16:20 is confusingly printed as 04:20

Sample with `hu_HU` locale at 16:20:38

    04:20  38s 2018-04-13
         ^^

> The logic behind it is that we want to show the most volatile parts first -- the time and seconds. The date is less volatile, so there's no sense in showing it first.

Actually I didn't mean to swap the order, I did it instinctively. I can live with the reverse order, but I have to say, I cannot remeber ever seeing the time before the date in any kind of log messages.

davvid:
Okay, you've convinced me.  It's true that the format string we're using is peculiar from the global perspective.  Thanks!

kakra:
The format string could be passed through I18n/gettext... That way, it's possible to show the most volatile first correctly in every locale...

alex-gulyas:
I don't think passing the time format through gettext is the correct thing to do, because then the time format will depend on the current language, and not the current locale. Locale and language settings are not the same. `%x` and `%X` are formats which correctly use the current locales date and time format, which for example might be different for US and UK locales, event though they both use English.

> Locale settings control how numbers, dates, and times display for your region – which may be a country, or a portion of country or may not even honor country boundaries. They are independent of your language.

-- [The Differences between Locales and Languages (blogs.msdn.microsoft.com)](https://blogs.msdn.microsoft.com/heaths/2005/02/17/the-differences-between-locales-and-languages/)

PR #801, opened 2018-04-13
gitcmds: Don't force patience diff algorithm

### LINKED ISSUE #799: git-cola does not respect diff.algorithm git setting
I've set `git config diff.algorithm histogram` and git-cola doesn't seem to respect this. In one of my projects, histogram diffs generate better patches.

kakra:
If the user specified a different preference, Git-Cola should honor that
and not use the patience algorithm. If no preference is set, Git-Cola
continues defaulting to the patience algorithm.

Fixes #799

Signed-off-by: Kai Krakow <kai@kaishome.de>


davvid:
Sweet, thank you!

PR #802, opened 2018-04-14
Improve German translation

kakra:
This improves the German translation by bringing it more in-line with
the wording of the official Git Book and removes some ambiguous
translations. It also replaces most occurrences of Git technical terms
with their English original as the German Git Book does. It should be
more clear now what each action really does.

Further work will follow while I'm using the software but the current
stage should be suitable for merge.

The po file was edited with poedit2 so it may result in more diff lines
than a hand-crafted edit. But at least the file now follows poedit2
format.

Signed-off-by: Kai Krakow <kai@kaishome.de>

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -9,25 +9,971 @@ msgid ""
 msgstr ""
 "Project-Id-Version: git-cola VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-10-17 01:06-0700\n"
-"PO-Revision-Date: 2013-05-16 19:54+0100\n"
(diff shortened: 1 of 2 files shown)


davvid:
Thanks!  Just a note -- one thing that might help with merge conflicts later is using `make po` after making all your edits.  That should take care of sorting the output.

I'm going to run `make po` after the merge to see if it picks up any edits, so if there's a subsequent commit on this file from me please rebase to pick up the latest.

davvid:
FYI `make po` did end up editing the file, so please update to pickup these edits before the next batch.

Issue #788, opened 2018-03-27
PNG Image not displayed


futureweb:
Hey there,

nothing serious ... but some PNG Images won't get displayed within the new Git Cola Image Viewer.
Sample: 

icon_arrow_up.png - http://temp.in.futureweb.at/icon_arrow_up.png
bannerVideoPlay.png - http://temp.in.futureweb.at/bannerVideoPlay.png

Maybe a Problem with the Alpha Channel?

bye from Austria
Andreas Schnederle-Wagner

davvid:
PR #803 (thanks @xieofxie) should have fixed this issue.  Thanks for the report.

PR #803, opened 2018-04-15
Draw checkboard for transparent image

xieofxie:
Should use static brush or not?
See issue #788.
Same as Eye of GNOME, see eog-scroll-view.c.
Signed-off-by: xhl <xieofxie@gmail.com>


davvid:
This really makes the various diff modes much more useful for svgs and other transparent images.  Nicely done.

PR #805, opened 2018-04-18
Add Alt+R hotkeys for both rebase Start and Continue

laerreal:
git's rebase feature is frequently used while preparing commits for publication.
Currently, rebase controls are placed deep in the menu.
Assigning hotkeys for them would be very handy.

I propose to use same hotkeys (`Alt` + `R`) for both Start and Continue
because they are never active simultaneously.


davvid:
Hmm, as long as that works then this seems good.  I would have expected that maybe Qt would complain about an ambiguous Shortcut overload, but I guess it makes sense that it allows it since they're only ever active one at a time.

PR #823, opened 2018-05-15
Flake8 fixes

scop:


test/diffparse_test.py
```diff
@@ -30,8 +30,8 @@ def test_diff(self):
         self.assertEqual(hunks[0].lines[4], '+class DiffSource(object):\n')
         self.assertEqual(
                 hunks[0].lines[-1],
-                "         self._header_start_re = re.compile('^@@ -(\d+)"
-                " \+(\d+),(\d+) @@.*')\n")
+                "         self._header_start_re = re.compile('^@@ -(\\d+)"
+                " \\+(\\d+),(\\d+) @@.*')\n")
```
davvid: Thanks for these fixes.  I usually ignore the sphinxtogithub since it's an external project but no need to undo that, but it might get clobbered in the future when we update the vendored copy.

For this line in the test, we might prefer a `r'aw string'` with an "r" prefix instead since we want the string to match exactly what's in the file.  I'll tweak this locally, thanks!

Issue #494, opened 2015-09-03
[enhancement] status: preserve horizontal scroll across updates


StanAngeloff:
This is related to #178. I tried to solve it again, but failed to do so. This has really been annoying me on projects with long directory paths. If the status widget is not wide enough to fit a filename, a horizontal scroll bar will appear. Scrolling to the right to reveal the filename and then selecting a different file from the list resets the horizontal scroll position.

![Demo](http://i.imgur.com/pG09L5y.gif)

davvid:
Ugh, I tried making this work but it looks like Qt is doing something automatically when the item is clicked.  I'd like to get this working if possible.  I hope there's a way to do it..

StanAngeloff:
Same here, I thought I nailed it in #178 and then gave it another go a month ago, still not found what's causing it. Bummer.

davvid:
I think I found a crucial missing ingredient.. `self.setAutoScroll(False)` and then a few extra bits to store/restore the horizontol scroll.  I think I should have something shortly...

StanAngeloff:
Works great! :star2: All I need to do now is unlearn to use the scroll wheel after each click.

PetrGlad:
There is a problem with this behavior: if you scroll right to see ends of long paths and then scroll vertically to files closer to project root then items will look empty.

PR #828, opened 2018-05-23
status: Scroll current item into view when selection changes

PetrGlad:
When moving selection up/down with keyboard in a long file list the
selected item goes out of view and there's no way to tell which file the
current diff corresponds to.
This change ensures that latest selected item is always visible.
The horizontal scrolling is preserved as it was implemented before.

See also issue #494 and pull request #178 for related discussion.

Signed-off-by: Petr Gladkikh <petrglad@gmail.com>

### DIFF
--- a/cola/widgets/status.py
+++ b/cola/widgets/status.py
@@ -822,7 +822,7 @@ class StatusTreeWidget(QtWidgets.QTreeWidget):
                 return
             idx -= len(content)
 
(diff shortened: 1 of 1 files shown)


PetrGlad:
The build fails on 2.6, should I fix this? How?

davvid:
Nicely done.  Thanks!

Issue #833, opened 2018-05-27
Changing diff viewer options results in an exception


living180:
Changing a diff viewer option, for example, toggling `Ignore all whitespace` results in the following exception:

    Traceback (most recent call last):
      File "/home/dharding/code/git-cola/cola/qtutils.py", line 39, in <lambda>
        action.triggered[bool].connect(lambda x: fn())
      File "/home/dharding/code/git-cola/cola/widgets/diff.py", line 597, in update_options
        self.widget.update_options()
    AttributeError: 'Viewer' object has no attribute 'update_options'

PR #834, opened 2018-05-27
diff: fix changing diff viewer options

living180:
In commit 649de5bb248c3a33e6778c44f75ab4457e13508c the diff viewer was refactored to be able to display images diffs.  As a result of that change the `parent` passed to the `Options` object is now a `Viewer` rather than a `DiffEditor`.  In the `Options.update_options()` method, it calls the `update_options()` method of the `parent` provided when it was constructed.  However, unlike `DiffEditor`, `Viewer` did not have an `update_options()` method, which would cause an exception when `Options.update_options()` was executed.

To address this add an `update_options()` method to `Viewer` which delegates to the `update_options()` method of the contained `DiffEditor`.

Fixes: #833

### DIFF
--- a/cola/widgets/diff.py
+++ b/cola/widgets/diff.py
@@ -382,6 +382,9 @@ class Viewer(QtWidgets.QWidget):
         else:
             self.stack.setCurrentWidget(self.text)
 
+    def update_options(self):
+        self.text.update_options()
+
     def reset(self):
         self.image.pixmap = QtGui.QPixmap()
         self.cleanup()


davvid:
I think I may have papered over this in 68399e35c0bac12b89ae68c67444a87efbc545d9 by setting the options.widget to the text widget, but I like your solution much better.  Can you throw a revert of 68399e35c0bac12b89ae68c67444a87efbc545d9 into your branch and verify that it's working correctly?

I'll merge this later tonight once I'm back home.  Thanks!

davvid:
Just tested this out -- yup, reverting my commit was the right thing to do. Thanks!

living180:
Thanks for the merge!  I hit the issue while running 3.1, and tried to test `master` but didn't realize it was a couple of weeks behind so I never saw your change from 68399e35c0bac12b89ae68c67444a87efbc545d9.

PR #867, opened 2018-08-03
improve and complete spanish translation (WIP)

victorhck:
more spanish translation to git-cola


victorhck:
finished spanish translation to git-cola

PR #877, opened 2018-08-25
Fix logo

Lin-Buo-Ren:
The current logo has some issues:

* No known practice to keep the formatting when modified
* Contains attributes that are not relevant to the graphics, but will make plaintext diffing difficult
* Aspect ratio not 1:1
* Dimension(side length) not in base 2

This PR addresses most of the problems and resizes the logo to 512x512px, most graphical details are kept *approximately* intact.

screenshot_20180826_004105


Lin-Buo-Ren:
The markup formatting tool, Tidy, can be installed on most GNU+Linux distributions via the snap:  
https://snapcraft.io/tidy

PR #878, opened 2018-08-29
import os in pynsist-preamble.py

cclauss:
__os__ is used 5 times but is never imported.


davvid:
Sweet, thanks for the fix.  FWIW, I re-released the v3.2 Windows installer to pickup this fix.

PR #879, opened 2018-08-30
Makefile: 'make flake8' tests the entire codebase

cclauss:
It is safer and just as fast to have flake8 test the entire codebase.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -271,7 +271,7 @@ app-tarball: git-cola.app
 	$(MARKDOWN) $< >$@
 
 flake8:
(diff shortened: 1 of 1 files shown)


davvid:
Good idea.  For the `pynsist-preamble.py` you can split the expression up into multiple lines and  then on the lines that fail append `  # noqa` to the end of the lines and flake8 should skip them.

Those are variables provided by the pynsist installer framework, so they're invisible.

davvid:
Merged, but I ended up reverting the part the makes it use `.` in the flake8 invocation.  See de29ab2 for details.

PR #903, opened 2018-11-10
use travis apt installer

stephengroat:
should run a bit faster


davvid:
sweet, thanks for the PR.

Issue #17, opened 2009-08-15
Better submodule support


davvid:
Interact with submodules as if they were part of the main repo could be really slick.

PR #911, opened 2019-01-10
Update git submodules support

jm4R:
Add context menu, action menu and toolbar items that allows specific/all submodules update. Related-to #17.


jm4R:
Looks like build failure is environmental problem, not new code problem.

PR #916, opened 2019-01-15
Submodule

xieofxie:
A simple widget for navigating in submodules. Maybe it is better to set up a dev branch. I also notice that there are many modifications after 'make pot'.

### DIFF
--- a/cola/gitcmds.py
+++ b/cola/gitcmds.py
@@ -658,6 +658,27 @@ def diff_upstream(context, head):
     return diff_filenames(context, base, tracked)
 
 
+def list_submodule(context):
+    """Return submodules in the format(state, sha1, path, describe)"""
+    git = context.git
+    status, data, _ = git.submodule('status')
+    ret = []
+    if status == 0 and data:
+        data = data.splitlines()
+        # see git submodule status
+        # TODO better separation
+        for line in data:
+            state = line[0].strip()
+            sha1 = line[1:OID_LENGTH+1]
+            left_bracket = line.find('(', OID_LENGTH + 3)
+            if left_bracket == -1:
+                left_bracket = len(line) + 1
+            path = line[OID_LENGTH+2:left_bracket-1]
+            describe = line[left_bracket+1:-1]
+            ret.append((state, sha1, path, describe))
+    return ret
+
+
 def merge_base(context, head, ref):
     """Return the merge-base of head and ref"""
     git = context.git
--- a/cola/widgets/submodules.py
+++ b/cola/widgets/submodules.py
@@ -1,0 +1,126 @@
+"""Provides widgets related to submodules"""
+from __future__ import absolute_import
+
+from qtpy import QtWidgets
+from qtpy.QtCore import Qt
+from qtpy.QtCore import Signal
+
+from .. import cmds
+from .. import core
+from .. import qtutils
+from .. import icons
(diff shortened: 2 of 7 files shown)


cola/gitcmds.py
```diff
@@ -658,6 +658,28 @@ def diff_upstream(context, head):
     return diff_filenames(context, base, tracked)
 
 
+def list_submodule(context):
+    """Return submodules in the format(state, sha1, path, describe)"""
+    git = context.git
+    SHA1_LENGTH = 40
```
davvid: Instead of `SHA1_LENGTH`, please use the existing `OID_LENGTH` which is already imported in this module.

cola/gitcmds.py
```diff
@@ -658,6 +658,28 @@ def diff_upstream(context, head):
...
+    status, data, err = git.submodule()
```
davvid: this might be slightly more explicit by saying `git.submodule('status')` instead

cola/widgets/submodules.py
```diff
@@ -0,0 +1,131 @@
...
+        if not self._active:
+            self._active = True
+            self.refresh()
+        return super(SubmodulesTreeWidget, self).showEvent(event)
+
+    def back_to_parent_command(self):
+        # https://stackoverflow.com/questions/7359204/git-command-line-know-if-in-submodule
+        # TODO if stable
+        status, path, _ = self.context.git.git(
+            'rev-parse', '--show-superproject-working-tree')
```
davvid: You should be able to do `self.context.git.rev_parse(show_superproject_working_tree=True)`.  What version introduced `--show-superproject-working-tree`?  We should probably add a `version.check_git(...)` key for it so that we can only use that flag when Git is new enough.

cola/widgets/submodules.py
```diff
@@ -0,0 +1,131 @@
...
+        self.main_layout = qtutils.vbox(
+            defs.no_margin, defs.spacing, self.tree)
+        self.setLayout(self.main_layout)
+        # button
+        # TODO better icons
+        self.refresh_button = qtutils.create_action_button(
+                tooltip=N_('Refresh'), icon=icons.sync())
+        qtutils.connect_button(self.refresh_button, self.tree.refresh_command)
+        self.back_to_parent_button = qtutils.create_action_button(
+                tooltip=N_('GoToParent'), icon=icons.push())
```
davvid: `GoToParent` isn't a real sentence or word.  How about, "Open Parent Repository" instead?

davvid:
Last note (I promise) -- please run `make check` before submitting new PRs to catch style issues.  It looks like the pylint checks found some long lines; we stick to 80 chars.  Please reflow the long line..

cola/gitcmds.py
xieofxie: Done. I am wondering whether the outputs of git is stable across different versions.

cola/widgets/submodules.py
xieofxie: Now the button will be hidden when it is unavailable. 
For submodules in submodules, its .git file contains the top parent instead of its direct parent which makes it a little troublesome.
And I am afraid I need some help about how to monitor the changes of submodules.

xieofxie:
Thanks! 'make check' is really helpful. Maybe it is better to design some new icons?

PR #930, opened 2019-02-15
Branch sort order based on committer time

laerreal:
Currently, branches are alphabet ordered.
But it's also handy to sort them by committer time.
So, recently edited branches are to the top of the list.

This patch series adds a toggle button to branch widget, allowing a user to select desired order.
Current order is also preserved in the state.

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -586,6 +586,12 @@ class Commit(ResetMode):
         return msg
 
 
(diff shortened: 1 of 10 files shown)


laerreal:
Added icons for dark version.

laerreal:
It is look like the problems reported by travis-ci, are with the commits those are already in `master`.

Issue #902, opened 2018-11-07
The section: "Status", could be rendered the information like a tree


yferras:
Now all the files are classified as: Staged, Modified, Untracked. But the list of files for each category has the same margin as its own category entry. Sometimes I have to look many times to differentiate which is which. But if the margin is different between the category and the list of files, I think that each user will be able to identify the category and the file in question more quickly. 
The follow image is for reference:
![ref](https://user-images.githubusercontent.com/11029945/48148800-a23aeb00-e288-11e8-9fd9-ef57aa13719a.png)

davvid:
Green is always staged, but besides that, we now have an option to enable this behavior thanks to @javierrodriguezcuevas

PR #931, opened 2019-02-25
Feature/status widget options

javierrodriguezcuevas:
Based on issue #902 added the option to leave/remove the indentation in status widget. Also added an option to show the number of files in each category.
The options can be found in Preferences/Settings

### DIFF
--- a/cola/widgets/status.py
+++ b/cola/widgets/status.py
@@ -123,7 +123,7 @@ class StatusTreeWidget(QtWidgets.QTreeWidget):
         self.setDragEnabled(True)
         self.setAutoScroll(False)
 
-        if not prefs.status_child_indentation(context):
+        if prefs.status_remove_indentation(context):
             self.setIndentation(0)
 
         ok = icons.ok()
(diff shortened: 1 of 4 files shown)


cola/widgets/status.py
```diff
@@ -524,6 +526,9 @@ def _set_subtree(self, items, idx,
             parent.addChild(treeitem)
         self._expand_items(idx, items)
         self.blockSignals(False)
+        if prefs.status_show_totals(self.context):
+            parent_title = parent.text(0).split()[0]
+            parent.setText(0, '%s (%s)' % (parent_title, len(items)))
```
davvid: What happens if this block of code runs more than once?  It avoids problems by calling `split()` to always take the first word only, but that seems a little too complicated.

All of the `_set_subtree` callers "know" what the correct title should be, so it'd be better to just pass down an explicit title instead of getting it from the `parent.text()`.  We can still do the `'%s (%s)'` thing here, but it'll at least separate the logic from being tied to details of what the message looks like.

For example, maybe the internationalized text will have more than one word, so it's better to make it generic.

PR #932, opened 2019-03-02
Make icons sharp in HiDPI screen

guoyunhe:


cola/app.py
```diff
@@ -178,6 +178,8 @@ def __init__(self, context, argv, locale=None,
 
         self.context = context
         self._install_hidpi_config()
+        # Make icons sharp in HiDPI screen
+        self._app.setAttribute(Qt.AA_UseHighDpiPixmaps, True)
```
davvid: This makes sense to be in the ColaQApplication constructor instead.

We still support PyQt4, so please use check `hasattr(Qt, 'AA_UseHighDpiPixmaps')` before using it here.

This code probably doesn't work as written, and should be hitting a traceback -- self._app hasn't been assigned yet -- it's getting assigned on the line below.

guoyunhe:
I updated the code as you commented.

Before:

screenshot_20190304_092413

After:
screenshot_20190304_092512

System icons, like the settings icon (from Breeze theme) look pretty nice. But app's SVG icons are still not clear.

guoyunhe:
I think the SVG icons need some changes. `x.svg` is using 768x1024px sizes, which caused some scaling issue. If I change it to 22x22px, it becomes sharp:

image

Could I also modify all icons? Thanks!

davvid:
Yes, please, that would be awesome.

davvid:
I've merged the current state as-is and speculatively updated the release notes to mention the upcoming updates to the icons.  I'll merge make sure to merge that in asap.  Thanks again

davvid:
Can you share your updated `x.svg`?  I tried doing the same by hand and it still looked blurry.

guoyunhe:
@davvid You can try this branch https://github.com/guoyunhe/git-cola/tree/make-icons-sharp

What I did is to change the size and view port of SVG files to 22x22px. Tested with @1.5 scale factor.

davvid:
Thanks, I've now done this to all the icons, including the ones in the `dark/` sub-directory.  I couldn't find an easy way to get stuff in inkscape to align along integer px grid boundaries but if you know of any tips I'm all ears.  Let me know if the icon updates look good to you.  Thanks again.

PR #962, opened 2019-07-07
README.md: add slackware instructions

richarson:
Add reference to SlackBuilds.org for Slackware


richarson:
A belated thank you!

PR #980, opened 2019-08-20
display the real import error for easier debug

jedie:
In my case i see this error:

`qtpy.PythonQtError: No Qt bindings could be found (No module named 'PySide')`

instead of only:

`qtpy.PythonQtError: No Qt bindings could be found`


davvid:
Please submit this change to https://github.com/spyder-ide/qtpy and once it's available upstream I'll pick it up.

PR #981, opened 2019-08-20
dag: add set_revtext_gather_paths

xieofxie:
In case project has many files.

I'm sorry that I'm working on Windows so I didn't run make check and make pot.

### DIFF
--- a/cola/models/prefs.py
+++ b/cola/models/prefs.py
@@ -38,6 +38,7 @@ TEXTWIDTH = 'cola.textwidth'
 USER_EMAIL = 'user.email'
 USER_NAME = 'user.name'
 SAFE_MODE = 'cola.safemode'
+AUTOCOMPLETE_PATHS = 'cola.autocompletepaths'
 SHOW_PATH = 'cola.showpath'
 SPELL_CHECK = 'cola.spellcheck'
 THEME = 'cola.theme'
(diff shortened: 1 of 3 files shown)


cola/widgets/dag.py
```diff
@@ -586,6 +587,10 @@ def __init__(self, context, params, parent=None, settings=None):
         self.refresh_action = qtutils.add_action(
             self, N_('Refresh'), self.refresh, hotkeys.REFRESH)
 
+        self.revtext_gather_paths_action = qtutils.add_action_bool(
+            self, N_('Gather Paths in Log'), self.set_revtext_gather_paths,
```
davvid: The user-visible label here might make more sense as simply `N_('Autocomplete file paths')` that way we don't mention `Log` or "gathering", which might not make sense in some contexts.

I'm kinda curious -- what was the original motivation for this change?  Is the autocomplete for paths getting too slow when working in large repos?  If so thank you for improving performance for folks that have large repos.

xieofxie:
Thanks for the comments! I would like to redesign this.

Yes, this switch should be a local setting. But I didn't explore the code carefully before, so I simply follow the 'Lock Layout'.. 

Right, it is quite slow for projects with many files..

xieofxie:
Now it totally depends on model.cfg.get(prefs.AUTOCOMPLETE_PATHS)

PR #982, opened 2019-08-21
stash: add author date to tool tip

xieofxie:
Signed-off-by: xhl <xieofxie@126.com>

### DIFF
--- a/cola/models/stash.py
+++ b/cola/models/stash.py
@@ -20,8 +20,8 @@ class StashModel(observable.Observable):
         if not model.initialized:
             model.update_status()
 
-    def stash_list(self):
-        return self.git.stash('list')[STDOUT].splitlines()
+    def stash_list(self, *args):
+        return self.git.stash('list', *args)[STDOUT].splitlines()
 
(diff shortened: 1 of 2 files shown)


cola/models/stash.py
```diff
@@ -32,11 +32,14 @@ def is_changed(self):
 
     def stash_info(self, revids=False, names=False):
         """Parses "git stash list" and returns a list of stashes."""
-        stashes = self.stash_list()
-        revids = [s[:s.index(':')] for s in stashes]
-        names = [s.split(': ', 2)[-1] for s in stashes]
-
-        return stashes, revids, names
+        stashes = self.stash_list('--format=%gd/%aD/%s')
```
davvid: Small sug: it might make sense to use a python `r'aw string'` here (prefix the string literal with `r`) so that it's clear that these are all literal values and we do not expect python to do any interpolation over the `%xxx` fields.  It doesn't make a difference in practice, but it makes it clear to the reader that we want to pass the string in exactly as-is.

PR #1001, opened 2019-10-20
recent: update recent repo list when opening repo

### LINKED ISSUE #937: Recent list does not gets updated
Recent list does not gets updated after switching repository (for example from favorites list).

stimut:
Currently, the recent repo list is only updated when the application is
exited, due to the list being stored in the settings, which are saved on
exit. As well as newly opened repos not showing on the recent list, it
means that if you start on repo A, open repo B, then open repo C, and
finally exit, repo B will not have been added to the recent list.

With this change, when a repo is opened it will be added to the recent
list immediately, and will therefore show in the recent repo list in the
menu. In addition to that, a signal was created for the changing of the
worktree, so that the bookmarks widget for the recent repo list can be
updated.

Closes #937
Signed-off-by: Tim Brown stimut@gmail.com

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -27,6 +27,7 @@ from .git import MISSING_BLOB_OID
 from .i18n import N_
 from .interaction import Interaction
 from .models import prefs
+from .settings import Settings
 
 
 class UsageError(Exception):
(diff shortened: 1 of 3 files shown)


stimut:
I'm open to feedback on where the best place to put the functionality is. In particular, where the code should go that saves the newly opened repo to the recent list.

I put it in `cmds.py` as I wanted it to trigger no matter how a repo was opened.

PR #1013, opened 2019-11-17
prefs: option to automatically load template commit

### LINKED ISSUE #735: Option to automatically add commit message template?
It would be great to have an option to automatically _always_ load the commit message template, when preparing a commit message. What do you think?

vnepveu:
- Add cola.commit.autotemplate constant in preferences
- Add checkbox in both "All repositories" and "Current repository"
- Load commit template if needed when opening new repository

Closes #735

### DIFF
--- a/cola/models/prefs.py
+++ b/cola/models/prefs.py
@@ -9,6 +9,7 @@ from .. import utils
 from ..cmd import Command
 
 
+AUTOTEMPLATE = "cola.commit.autotemplate"
 BACKGROUND_EDITOR = 'cola.backgroundeditor'
 BLAME_VIEWER = 'cola.blameviewer'
 BOLD_HEADERS = 'cola.boldheaders'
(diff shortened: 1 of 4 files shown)


cola/models/prefs.py
```diff
@@ -9,6 +9,7 @@
 from ..cmd import Command
 
 
+AUTOTEMPLATE = "cola.commit.autotemplate"
```
davvid: Single quotes for consistency, please.  This seems reasonable.    One small note -- everywhere else we use just a single flat namespace for config values.  e.g. `cola.autoloadcommittemplate` is more in-line with the rest of the config keys we use.  In documentation we spell it as `cola.autoloadCommitTemplate` which helps the readability.

Please make sure this feature is documented in `doc/git-cola.rst` in the `CONFIGURATION VARIABLES` section.

vnepveu:
```suggestion
AUTOTEMPLATE = "cola.commit.autotemplate"
```
I used this format because of your comment (see https://github.com/git-cola/git-cola/issues/735#issuecomment-336750453)

davvid: Ah, sorry for not remembering.  Thanks for the updates.

PR #1016, opened 2019-11-20
widgets: show server response with clickable links

### LINKED ISSUE #951: Feature request: Show server response messages with clickable URLs when pushing
It's common for git server response messages to include messages with a convenient URL to e.g. create a PR/MR for a newly pushed branch. It would be nice if git-cola showed these messages with any URLs as clickable links.

bensmrs:
This commit allows to open a dialog when the remote server sends messages. Several changes are made:
- The `link` function is moved from `widgets.about` to `qtutils` to ease code reuse;
- The `self.runtask.start` function calls in `widgets.branch` are now given the `result` argument to trigger the dialog if a remote message is found;
- A new widget for the response dialog allows to show remote messages if lines starting with 'remote: ' are found in the output of the git command.

Closes #951
Signed-off-by: Benjamin Somers <benjamin.somers@telecom-bretagne.eu>

### DIFF
--- a/cola/qtutils.py
+++ b/cola/qtutils.py
@@ -274,12 +274,8 @@ def link(url, text, palette=None):
     rgb_values = 'rgb(%s, %s, %s)' % (color.red(), color.green(), color.blue())
     scope = dict(rgb=rgb_values, text=text, url=url)
 
-    return """
-        <a style="font-style: italic; text-decoration: none; color: %(rgb)s;"
-            href="%(url)s">
-            %(text)s
-        </a>
-    """ % scope
+    return ('<a style="font-style: italic; text-decoration: none; '
+            'color: %(rgb)s;" href="%(url)s">%(text)s</a>' % scope)
 
 
 def add_completer(widget, items):
@@ -622,6 +618,8 @@ def default_size(parent, width, height, use_parent_height=True):
 def default_monospace_font():
     if utils.is_darwin():
         family = 'Monaco'
+    elif utils.is_win32():
+        family = 'Courier'
     else:
         family = 'Monospace'
     mfont = QtGui.QFont()
--- a/cola/qtutils.py
+++ b/cola/qtutils.py
@@ -266,6 +266,22 @@ def textbrowser(text=None):
     return widget
 
 
+def link(url, text, palette=None):
+    if palette is None:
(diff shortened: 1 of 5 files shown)


davvid:
Hmm, do you have a sug on how to best test this?  I tested it by pointing to an invalid gitlab.com URL and I got `remote: ...` messages back.

On the downside, though, I noticed that it opened three dialogs.

The 1st dialog is the regular dialog we get when `git fetch` returns a non-zero exit status.  It includes the remote messages in a text box at the bottom of that dialog.

Then there were two different RemoteMessages dialogs with just the remote messages.  One was on top of the other one, which caused weird modality issues (I'm no Linux) and the one was kind of in the background.  It seems like there's something funky about that.

In Qt, we have to hold onto widgets otherwise they'll get deleted.  So creating a widget in a function and then calling show() might not behavior correctly in all cases.

I haven't tested the use case where `git fetch` or `git push` return exit status zero, but also have remote messages.  Currently if you open the `ctrl-0` log window then you'll see all the messages in there.  The remote dialogs have tended to try to be unobtrusive, which means that the messages go in the console so that the user doesn't get a dialog in their face for the normal use case (when push succeeds).

If we only need to see these when things fail, maybe we can integrate the messages into the existing dialog, or replace it with a custom one with a special "Remote messages" section?  Otherwise, if we want these when things succeed as well then we might need that to be opt-in (it's a change in behavior). 

Right now if things succeed then the Push/Fetch/Pull dialog is closed and all is good (unix rule: "silence is golden").  Let me know what you think.

bensmrs:
Hi, thanks for the feedback. Actually, I did not test the code when things fail… My test case was a Git(Hu|La)b repository cloned, then renamed. When pushing again, the server tells us that the repository has moved. I will add the relevant checks to deal with cases where it fails; I did not think about that.
On your side, could you confirm that the behaviour looks good to you with my test workflow?

davvid:
Ah, that makes sense.  I think if the remote sends messages on success then it probably does make sense to present them to the user, though there may be some chatty remotes so perhaps we can make that opt-out (enabled by default) so that folks working with chatty remotes have a way to silence the dialog.

For the fail case, it seems like the simplest thing to do is to only show the remote messages dialog when things succeed, since the fail case is already handled by the existing dialog.

PR #1023, opened 2019-12-11
stash: Allow stashes to be renamed

bensmrs:
These commits:
- Allow the user to rename a stash;
- Remove the superfluous constraints on stash names;
- Change the interface behaviour to something that feels more intuitive;
- Update the deprecated `git stash save` command to `git push -m`.

### DIFF
--- a/cola/models/stash.py
+++ b/cola/models/stash.py
@@ -1,4 +1,5 @@
 from __future__ import division, absolute_import, unicode_literals
+import re
 
 from .. import cmds
 from .. import core
@@ -32,7 +33,7 @@ class StashModel(observable.Observable):
 
     def stash_info(self, revids=False, names=False):
         """Parses "git stash list" and returns a list of stashes."""
-        stashes = self.stash_list(r'--format=%gd/%aD/%s')
+        stashes = self.stash_list(r'--format=%gd/%aD/%gs')
         split_stashes = [s.split('/', 3) for s in stashes if s]
         stashes = ['{0}: {1}'.format(s[0], s[2]) for s in split_stashes]
         revids = [s[0] for s in split_stashes]
@@ -89,10 +90,12 @@ class DropStash(cmds.ContextCommand):
         status, out, err = git.stash('drop', self.stash_ref)
         if status == 0:
             Interaction.log_status(status, out, err)
+            return re.search('\\((.*)\\)', out).group(1)
         else:
             title = N_('Error')
             Interaction.command_error(
                 title, 'git stash drop ' + self.stash_ref, status, out, err)
+            return ''
 
 
 class SaveStash(cmds.ContextCommand):
@@ -107,6 +110,33 @@ class SaveStash(cmds.ContextCommand):
             args = ['save', '--keep-index', self.stash_name]
         else:
             args = ['save', self.stash_name]
+        status, out, err = self.git.stash(*args)
+        if status == 0:
+            Interaction.log_status(status, out, err)
+        else:
+            title = N_('Error')
+            cmdargs = core.list2cmdline(args)
+            Interaction.command_error(
+                title, 'git stash ' + cmdargs, status, out, err)
+
+        self.model.update_status()
+
+
+class RenameStash(cmds.ContextCommand):
+    """Rename the stash"""
+
+    def __init__(self, context, stash_index, stash_name):
+        super(RenameStash, self).__init__(context)
+        self.context = context
+        self.stash_index = stash_index
+        self.stash_name = stash_name
+
+    def do(self):
+        # Drop the stash first and get the returned ref
+        ref = DropStash(self.context, self.stash_index).do()
+
+        # Store the stash with a new name
+        args = ['store', '-m', self.stash_name, ref]
         status, out, err = self.git.stash(*args)
         if status == 0:
             Interaction.log_status(status, out, err)
--- a/cola/models/stash.py
+++ b/cola/models/stash.py
@@ -107,9 +107,9 @@ class SaveStash(cmds.ContextCommand):
 
     def do(self):
         if self.keep_index:
-            args = ['save', '--keep-index', self.stash_name]
+            args = ['push', '--keep-index', '-m', self.stash_name]
         else:
-            args = ['save', self.stash_name]
+            args = ['push', '-m', self.stash_name]
         status, out, err = self.git.stash(*args)
         if status == 0:
             Interaction.log_status(status, out, err)
--- a/cola/utils.py
+++ b/cola/utils.py
@@ -197,13 +197,6 @@ def strip_prefix(prefix, string):
     return string[len(prefix):]
 
 
-def sanitize(s):
-    """Removes shell metacharacters from a string."""
-    for c in """ \t!@#$%^&*()\\;,<>"'[]{}~|""":
-        s = s.replace(c, '_')
-    return s
-
-
 def tablength(word, tabwidth):
     """Return length of a word taking tabs into account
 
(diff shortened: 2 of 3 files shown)


cola/models/stash.py
```diff
@@ -104,9 +107,36 @@ def __init__(self, context, stash_name, keep_index):
 
     def do(self):
         if self.keep_index:
-            args = ['save', '--keep-index', self.stash_name]
+            args = ['push', '--keep-index', '-m', self.stash_name]
         else:
-            args = ['save', self.stash_name]
+            args = ['push', '-m', self.stash_name]
```
davvid: `git stash save` wasn't deprecated until Git 2.16.0, and versions before 2.13 did not have that command at all.

Please add a version check (see `git grep version.check`) with a new `stash-push` entry for '2.16.0' in `cola/version.py` so that we can support old and new git versions.  That way we can use `git stash push` for 2.16+ only.

cola/models/stash.py
```diff
@@ -104,9 +107,36 @@ def __init__(self, context, stash_name, keep_index):
...
+        self.context = context
+        self.stash_index = stash_index
+        self.stash_name = stash_name
+
+    def do(self):
+        # Drop the stash first and get the returned ref
+        ref = DropStash(self.context, self.stash_index).do()
+
+        # Store the stash with a new name
+        args = ['store', '-m', self.stash_name, ref]
```
davvid: Ditto, if we wanted to be maximally compatible we can add a similar `stash-store` check for `2.2.0`, but only if there's a way to do this with the old `git stash save` here too.  We try to be maximally compatible.

cola/widgets/stash.py
```diff
@@ -192,7 +216,7 @@ def stash_apply(self, pop=False):
         context = self.context
         index = get(self.keep_index)
         cmds.do(stash.ApplyStash, context, selection, index, pop)
-        QtCore.QTimer.singleShot(1, self.accept)
+        self.update_from_model()
```
davvid: Just double-checking -- the deferred call to `self.accept` was making sure we close the dialog here.  For "Apply", I can see wanting to apply more than one stash (though it is atypical) so we'd want to keep the dialog open.

But for save below, we also remove that call -- but I think we would want to close the dialog after saving, or do you have a different workflow where you keep the dialog open for multiple actions?

cola/widgets/stash.py
```diff
@@ -206,21 +230,14 @@ def stash_save(self):
             parent=self)
         if not ok or not stash_name:
             return
-        # Sanitize the stash name
-        stash_name = utils.sanitize(stash_name)
-        if stash_name in self.names:
-            Interaction.critical(
-                N_('Error: Stash exists'),
-                N_('A stash named "%s" already exists') % stash_name)
-            return
```
davvid: Nicely done.  True, the stash "names" are really just the messages, and don't need to be unique.

cola/models/stash.py
```diff
@@ -89,10 +90,12 @@ def do(self):
         status, out, err = git.stash('drop', self.stash_ref)
```
davvid: For the block below, I've been trying to stick to single-exit.. so the logic flow would be..
```
if ...:
    result = ...
else:
    result = ...
return result
```
That way there's just a single return.  Not always possible, but seems sensible below.

cola/models/stash.py
```diff
@@ -89,10 +90,12 @@ def do(self):
...
         if status == 0:
             Interaction.log_status(status, out, err)
+            return re.search('\\((.*)\\)', out).group(1)
```
davvid: Please add a comment above the regex showing an example of the string it's trying to parse.  That way we can spot-check it in the future.

It's not reliable to rely on the output of `git stash drop` like this, so we shouldn't blindly assume that the regex matches.  It's better to assign to a variable and handle `None` since we can't predict what the output will look like in the future.  The output of `git stash` is "porcelain" (it's not "plumbing" in git jargon) so the output is not guaranteed to be stable between versions.  Thus, we have to be robust.

For the regex itself, please use `r'aw strings'` so that we don't have to escape the backslashes.  Also, named groups would make this a bit easier to grok.  Maybe something like this?

```
result = ''
match = re.search(r'\((?P<name>.*)\)', out)
if match:
    result = match.group('name')
```

cola/utils.py
```diff
@@ -197,13 +197,6 @@ def strip_prefix(prefix, string):
     return string[len(prefix):]
 
 
-def sanitize(s):
-    """Removes shell metacharacters from a string."""
-    for c in """ \t!@#$%^&*()\\;,<>"'[]{}~|""":
-        s = s.replace(c, '_')
-    return s
-
-
```
davvid: Brilliant.  Being able to delete this function is great.

cola/widgets/stash.py
bensmrs: You're completely right, such a workflow seems quite… dirty, to say the least. But I would expect the window to only close when I ask it to; the button reads "Save", not "Save and close"… What do you think regarding the UX?

cola/models/stash.py
bensmrs: I see… I'm going to look into the matter, the philosophy to keep supporting older versions is pretty relevant!

cola/models/stash.py
bensmrs: From what I see, `git stash store -m <message> <commit>` is available since git 1.8.4. Unfortunately, I don't see any command to achieve the goal before this version… Should we check the version and gray out/hide the rename button for versions prior to 1.8.4?

cola/models/stash.py
bensmrs: Right, I would also like to have your opinion on the return value if there is an error. I chose `''` to make the linter happy (same return type). Is it ok or would you see something else?

cola/models/stash.py
bensmrs: Looks good to me! I will try that.

cola/models/stash.py
davvid: Hiding the rename button for older versions makes sense.  Thanks for considering compatibility.

cola/widgets/stash.py
davvid: It's okay to keep it as "Save" since it matches the expected one-off nature of stashing something.  It's very convenient that we close the dialog, since otherwise the user would be burdened with having to do so for the common case.

Another concern is that we already have translations written for the `Save` string (see the `po/` directory), so sometimes it's limiting but I do try to avoid changing existing strings that already have established translations in place.  I should probably document that in the guidelines so that it's taken into consideration.

I drive it from the keyboard.  The "Save" button is the default action, so to save a stash it's `shift-alt-s enter stash-name enter` which is slick from a keyboard usability perspective.  I don't want to hit `esc` at the end to close the dialog because it encourages a sloppy workflow where the user has tons of stashes.

If we want to apply a different stash after saving, then it's not that inconvenient to hit the `shift-alt-s` shortcut and bring up the dialog again.  That said, we don't have an easy way to apply stash without using the mouse, so I'll try to add that shortly.

If I push that up, apologies if you have to rebase for it.

cola/models/stash.py
davvid: empty string is fine by me, especially if the linter was complaining about it.  It does make things simpler for callers since they wouldn't have to deal with any `None` vs `str` concerns.

cola/widgets/stash.py
davvid: Alrighty, I just pushed up some small changes to the widget that adds hotkeys for apply, pop, and drop.  Please rebase.  The changes were pretty minimal (mostly purely additive) so it should be easy to rebase your tree against it, but let me know if you run into any snags.

k,j navigates up,down the stash list.  Ctrl+Enter applies stashes.  Ctrl+Backspace pops the stash.  Ctrl+Shift+Backspace drops the stash.  Save is the default action so it gets the Enter hotkey by default from Qt.

We can later maybe add the Edit Ctrl+E hotkey for the new rename action here. We're "editing" the stash name, so it might make sense to take over that action/verb in this context.

Issue #1004, opened 2019-11-06
Change username field to a text field


mboekhold:
Currently when authenticating, the username prompt is a password field. It may be a good change to make it a text field as I misspell my username sometimes :stuck_out_tongue: . I'd gladly construct a PR if this change were to be approved

davvid:
Thanks for the heads-up.  We support the usual askpass stuff.  Were you able to get this resolved?  We do ship a fallback askpass handler (see share/git-cola/bin/) so if there's something that needs changing there let us know.

mboekhold:
Hi @davvid , where is the prompt for authentication found, the username and password prompt when initially trying to pull or push?

davvid:
Check your enviornment for SSH_ASKPASS and GIT_ASKPASS.  We have a fallback askpass handler, but it might only kick in if you're using ssh URLs.

I would strongly recommend using ssh public key authentication with ssh URLs instead of needing to type your password in all the type.  I'll leave this issue open so that we can document some of the details around ssh askpass handles and the git credentials stuff.

https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables  (see GIT_ASKPASS)

https://github.com/jeroen/askpass

https://git-scm.com/docs/gitcredentials

I already mentioned `share/git-cola/bin/ssh-askpass`.  That's the fallback handler.  Your distro might already provide one, otherwise it might be using that, or it's doing whatever Git is configured to do.

I'll lave this issue open so that we can add a FAQ page that includes this information.

mboekhold:
indeed, the situation I am describing above happens when there is not environment askpass set, specifically the file that you mentioned `share/git-cola/bin/ssh-askpass` the input box when filled in appears in asterisks. It's very minor but i'd like to change that to only appear when entering password. improves the useability a teeny tiny bit

davvid:
While this has already been updated in cola, we are going to upstream the fixes to git-gui by submitting the patch to the core git development list, address the review notes there, and then bring those changes back here.  We'll leave this issue open until we get a chance to do that.

Related: https://github.com/prati0100/git-gui/pull/38

davvid:
Upstreamed patch:

https://public-inbox.org/git/20201107222039.13900-1-davvid@gmail.com/T/#u

PR #1026, opened 2019-12-16
Added check for the input box

mboekhold:
Simple check to omit the asterisks on username prompt #1004


mboekhold:
I'll see what the guys from git-gui think of the change, I'm not certain if the Password field is localized, I'll do some investigation and if there's anything that needs a change I'll see if I can push it to the upstream

Issue #944, opened 2019-03-27
Make status panel more intuitive


piegamesde:
I'm currently trying out a lot of graphical git clients and I'd like to give you some feedback to make the status panel more intuitive:

- The icons are not consistent: Some icons refer to the file type of a file, but others symbolize the status of the file and yet others indicate the change of the type. Please settle on one information to display through the icon and use it everywhere in the panel. Personally, I'd go for added/removed/modified/renamed icons, but that is clearly biased from my GitKraken background
- The grouping of changes is not communicated clearly. The changes have a hierarchy but are presented as list. The gui currently does not make it clear that "Modified", "Untracked" etc. are not just files, but categories that can even be collapsed. To resolve this, I'd suggest indenting all entries to create a visual hierarchy. Alternatively, split the view into multiple list panels (one for staged and one for unstaged changes)
- This may be not true to how git works, but displaying untracked files as added file makes it more intuitive to users less familiar with git.

jm4R:
* 2nd is already fixed, but not released (indentation will be optional in the next `git-cola` release)
 * 3rd - one of the great thinks about `git-cola` is naming convention. Almost everything is named in the same way that `git` does. I wouldn't change that path.

piegamesde:
Oh yes, *way* better. Would you mind checking the option by default when updating (or at least for new users)?

And what about the icons? I'm willing to help out if needed (if so, please point me to some source files I should start looking into).

Issue #697, opened 2017-05-10
Several contextual menu entries in Status widget should be disabled when no items in list are selected


Vdragon:
As the following screenshot, several entries in contextual menu of the Status widget shouldn't be available when no files are selected:

![_025](https://cloud.githubusercontent.com/assets/1192163/25883793/ff93f056-357f-11e7-89b5-3e6b3db37f05.png)

We probably should only allow the "Launch Terminal" entry in this case

davvid:
Thanks again for all the great sugs!

PR #1030, opened 2019-12-28
StatusTree view as actual tree (Working Prototype)

dvdotsenko:
git-cola-with-status-tree3

**It's a functional prototype (used it to commit changes in this branch) of StatusTree redone as actual tree of folders and files.**

Not really asking you to merge this. Mostly a start of conversation about what it takes to move StatusTree view and other parts of the app into direction that is comparable to other modern Git clients.

## Why / What:

- Routinely work with in-repo python wheels or yarn offline tgzs. These are huge (thousands of files in case of yarn) folders. `git-cola`'s original StatusTree view was painful with these set ups. **Primary reason for this branch is to allow me to "stage a folder"**

- Have spent last 20+ years working with Git (on Windows with GitExtensions, on Mac with SourceTree and others) UIs and drove fairly large development teams. In process picked up "best practices" associated with Git workflows. Git-Cola is largely there in "DiffView" implementations (contextual switch from file, to hunk to selected lines is awesome!), but the "StatusTree" view sucked compared to (still crappy from stand point of UX) implementations in GitKraken, SmartGit, SourceTree etc). I need an open-source tool I can give to my dev team and know workflow does not suck. This branch is attempt to implement StatusTree tree with eye on latest best practices (as accompaniment to a solid IDE, like IntelliJ)

-  :( Thought I could make both versions of StatusTreeView work against present *Model but despite my initial attempts found it brittle. MVC here is implemented in React-like way - tree is expressed as abstract Nodes tree first, and then rendered to Qt-native tree. All actions on the tree are immediately converted into respective abstract Nodes and the Vews are re-rendered from there. Biggest change is in passing more of native Git status data from `gitcmds.diff_***` functions. See `diff_index_map` as example or sourcing and `StatusTreeWidget.refresh()` for use of this data.

- Pruned right-click menus drastically to focus on actions I know dev crews routinely rely on daily basis. This is subjective, strong opinion very loosely held. No religious assertions here... 

- Unwired some things (like filters, some global shortcuts) WITHOUT prejudice. Did not need that for proof of concept. Definitely would like to put MORE filters in-place, including per-type-of-status filters. Just focused on what I can do in one day (Christmas Day). Definitely aware I broke some things.

- Intentionally focused on Python 3. QtPy under 5 now gives warnings about Python 2.x official support ending in Jan 2020. Kinda see it pointless to continue to focus on 2.x line now, since even QtPy is moving on. 

- forcefully and rudely added typehinting where needed because developing on this codebase without hinting in IDE was originally a royal pain.

- broke formatting and did not care to run / adhere to formatting rules. Present formatter single-lines function signatures into unreadable mess. For this POC did not want to fight / waste time on reading code formatted for non-humans. Don't mind formatter long-term.

- Could not figure out why, but standard icons `QtGui.QIcon.fromTheme('folder')` did not work / show in the tree. Is the theme custom? Instead of fixing it, just used unicode symbols for folder, plus, minus, modified icons. 

## What's next:

Don't merge (obviously). It's just a sand-box to doodle with a different approach to StatusTree management. Let's chat about what parts / approaches in work here and are acceptable to you.

If there is enough common ground on next steps, willing to throw official $ and people support.
What I really care about as outcome is an open-source Git UI client that has a chance of competing with SourceTree / GitKraken for core workflows.


dvdotsenko:
Some of the "Issues" this may be tackling:

https://github.com/git-cola/git-cola/issues/944 - Status view UX
https://github.com/git-cola/git-cola/issues/697 - Context menu fit to selection

davvid:
This is really great.

python2.7 is still a reality for a bit for me, so I can help by keeping that part alive post-facto.  It shouldn't be too hard, I'm assuming.  Sorry about the typehints thing -- there's a comment-style typehint that doesn't break python2 -- please use those for now since we still support python2.  Ditto with the `f` strings -- we'll have to do a pass at doing that the old way.  I was able to hack it up enough to run in 2 but I'll wait until it's farther along before distracting with that.

The checkbox-to-stage should be pretty easy to switch to a `+` "add" button.  That's nicer because it's explicit.

A long time ago cola had the same feature as git-gui where if you clicked the file's icon it would stage that file.  Users would often click the file's icon hoping to continue staging parts of that file, but would accidentally click the icon not knowing that it would stage the whole file and lose the careful staged edits they made beforehand.  We removed that feature, and retained the double-click shortcut and keyboard hotkey.

In hindsight, the subtlety of the icon being clickable was the real problem.  Having an explicit button like you've proposed solves that issue in a really nice way since users won't be clicking it accidentally.

Navigation key affordances can probably make it still keep the easy-to-navigate-files-across-folders feel by having alt-j/k hotkeys that jump to the "next" file regardless of folder, basically skipping folders when navigating around.  That way if something is a few directories deep it can still be selected by a single "alt-j" and not run into any of the intermediate folders that might be there.

For example, moving from no selection, to robot-prod, splash, and then SettingsScreen is currently three keypresses: `j j j`

In the new widget the `j j j` interaction might become `alt j j j` instead, which would be fine.  In the browse widget, `j` pretty much behaves just like the down arrow.

The existing tree widgets (e.g. `git cola browse`) already has nice `h j k l` keyboard behavior for navigating nested trees, so it makes sense to be consistent with that in the status widget.

The auto-collapse of large subdirs is really slick.

I wonder if there's any benefit in keeping the old status mode around as a "flat" mode toggle?  I'll try to not think about that since it would be kinda bloaty to have two, and I think the new one will be enough of an improvement that the old one won't be missed.

Thanks for reaching out and taking this on, this is super cool.

davvid:
BTW feel free to ignore the python2 travis failures.  Those started failing on the travis side (it can't even create a virtualenv) so my guess is that maybe it's part of the whole python2-EOL thing.

I was waiting until after the holidays to see if there was an official announcement or notice about people having issues with python2 virtualenvs but nothing's materialized just yet.

I'll either switch it over to use `tox` or disable the travis/py2 tests next week.  They all run fine locally

davvid:
In case it helps, I pushed a couple of the py2-compat commits to my [fork](https://github.com/davvid/git-cola) in the [status-tree-as-tree](https://github.com/davvid/git-cola/tree/status-tree-as-tree) branch.

dvdotsenko:
> python2.7 is still a reality for a bit for me

:+1: 

> there's a comment-style typehint that doesn't break python2 -- please use those for now since we still support python2

:+1: 

> The checkbox-to-stage should be pretty easy to switch to a + "add" button. That's nicer because it's explicit.

Up to you.

Thought about this for a while and from multiple sides:
 
Native checkboxes felt like more of a foolproof way to go for me. They have native UX (space bar etc) handling on all OSes, have native (reactive) event handler (`on change`), scales and renders natively on all platforms (and people intuitively know what they are).

Faking what essentially is a checkbox was just a weird to me as faking HTML checkboxes using images.

Given I don't see a better way, so rather happy with native checkboxes for now.

I can totally see needing "custom" checkboxes (buttons) if there are more than 3 present (check, uncheck, partial) states needed.

> retained the double-click shortcut and keyboard hotkey.

Default double-click action kinda bugged me. "Switch focus to DiffView" made more sense to me, but i realize it's very very subjective. So, contemplating abstracting settings for these into configurable and ensure that Shortcuts Help window dynamically renders the right, custom, shortcuts. 

Not a front-burner for me, but it's definitely there... in my mental queue somewhere behind a "better multi-branch DAG view"...

Wanna chat through your UX philosophy. I see you are in US Pacific Timezone. Can probably do a quick Hangout at some point.

> `j j j`

Don't speak `vi` :)

> i'll either switch it over to use tox or disable the travis/py2 tests next week. They all run fine locally

@davvid How do you run tests? I have a docker-based set up for tests, but that locks it to one version of Python with one version of Qt. Do you have to install every possible version of Qt and every possible version of Python globally for run against `tox`?

dvdotsenko:
image

Added more commits:
- fixed (broken by me) Unmerged files display
- added back Status Tree Filter
- UX doodling (tab order, shortcuts to navigate between Tree, DiffView, CommitView)

dvdotsenko:
personal TODO left:

- single-branch-only DAG view is really holding me down. Need to see all branches there because of many workflows that require seeing them all.
- I think i broke revert / reset in some cases. 
- Need to figure out how to lock down / stabilize test env. Feel icky without tests. At this point run tests inside Docker only for Python 3, Qt 5.x.

davvid:
I just merged #1034 which fixed the tests on travis.  The `.travis.yml` should give you a good idea of how it's tested locally.  I use a similar setup -- I use debian/testing and I have `python-qt5` and `python3-qt5` installed so that pyqt is available in both versions (locally, python2.7 and 3.7).

The `make tox` target is the simplest way to run the tests and flake8 checks against all your available python versions.  On travis we use `make tox-check` as it also runs pylint, but that's slower which is why I typically save that for the end and use `make tox` while developing.  It's setup to run pylint against two different versions of python because it effectively uses two different versions of pylint because of how `requirements-dev.txt` is structured.

Not sure why, but `tox` on travis doesn't seem to honor the `--skip-missing-interpreters` flag which is why we have to override envs to use in the `.travis.yml`.

Regarding the checkboxes -- that's true.  Native controls have a lot of advantages.  I was thinking of something like a QPushButton but a checkbox is simple and does have a lot of nice affordances as you noted.

For a mult-branch dag view -- type `--all` into the input/branch field.  Like `gitk`, it's designed to accept the same arguments as `git log`.  Ditto `git dag --all` does the same thing.  Indeed it could be improved nonetheless.

Feel free to ping me on hangouts if you'd like to chat.  I'll have some high latency in responding sometimes but I'm generally around in the evenings.  Thanks again.

PR #1035, opened 2020-01-17
Add option to compare SVGs in plain text

### LINKED ISSUE #859: Allow plaintext diff for SVG files
Currently, Git Cola shows image diff for SVG files, however sometimes plaintext diff is much preferable in these files, the diff widget should allow switching between image diff and plaintext diff in these cases.

lah7:
This adds a simple checkbox to **Preferences** under **Settings** to toggle between the text and image diffs for files that end with `.svg`. The default is **unchecked**.

In some projects, it is preferable to compare SVGs for XML changes, as opposed to its image output differences.

Closes #859.

image

### DIFF
--- a/cola/models/prefs.py
+++ b/cola/models/prefs.py
@@ -37,6 +37,7 @@ SAVEWINDOWSETTINGS = 'cola.savewindowsettings'
 SHOW_PATH = 'cola.showpath'
 SORT_BOOKMARKS = 'cola.sortbookmarks'
 SPELL_CHECK = 'cola.spellcheck'
+SVG_PLAIN_TEXT = 'cola.svgplaintext'
 STATUS_INDENT = 'cola.statusindent'
 STATUS_SHOW_TOTALS = 'cola.statusshowtotals'
 THEME = 'cola.theme'
@@ -81,6 +82,7 @@ class Defaults(object):
     hidpi = hidpi.Option.AUTO
     status_indent = False
     status_show_totals = False
+    svg_plain_text = False
 
 
 def blame_viewer(context):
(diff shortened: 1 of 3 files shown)


xieofxie:
Just a quick thought.. Will it be better to be a button/switch in Diff view?

lah7:
I think so, it would more convenient for the user so they can quickly switch between the two (see mock-up at https://github.com/git-cola/git-cola/issues/859#issuecomment-575837648 :camera: :bulb:). This would be similar to GitHub's diff interface.

Though this PR gets the job done _for now_, I have missed this functionality ever since the image viewer was introduced. My use case would be "set and forget" as I'd most often be comparing SVGs for its XML metadata/attributes, rather then the rendered output.

I did look into the diff widget first, but I am puzzled on how to make it work with code. :confused:

davvid:
A toggle pushbutton that remembers its state per-file extension in the header area of the widget might be a nicer way, that way those of us that might want to see both of the diffs can easily toggle between them, while still satisfying the set-and-forget use case.

I don't think we should merge this as-is, but it is a topic we want to address.

lah7:
Agreed. :+1: I'll close this PR as it's noted in #859.

davvid:
I've started integrating this feature into the diff editor directly.   There's now a toggle button that enables/disables the image diff mode.  This lays the foundation for automatically remembering the state of this toggle, which will be added shortly.

davvid:
Please test master.  As suggested by @xieofxie  this is now implemented as a toggle in the diff viewer's header, and it automatically remembers its state.  There's no configuration involved now -- it "just works".

xieofxie:
@davvid it seems we use file link now? How to work around it in Windows? Thanks

xieofxie:
OK.. I got it. I should install instead run directly. This toggle is great~

lah7:
Thank you for integrating! Working great here on Manjaro Linux. :+1:

PR #1045, opened 2020-02-26
cmds: speed up showing binary untracked files

wm4:
Apparently, Qt's text renderer is pretty slow when text includes
unprintable characters (code points below 32), or many different code
points, or if code points are not part of any fonts. This is possibly
due to font selection fallback handling or so.

Selecting an untracked binary file  can make the git-cola GUI
unresponsive for up to 1 or 2 seconds or so, even if the file size is
well below the default read limit of 2 KB.

Fix this by accepting printable ASCII characters only. string.decode()
using the "ascii" encoding allows unprintable code points below 32, so
filter them manually.

Signed-off-by: wm4 <wm4@nowhere>

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -2044,11 +2044,31 @@ class SetUpstreamBranch(ContextCommand):
         cfg.set_repo('branch.%s.merge' % branch, 'refs/heads/' + remote_branch)
 
 
-def clean_nonprintable(s):
-    res = ""
-    for c in s:
-        res += c if ord(c) >= 32 or c == '\n' else '.'
-    return res
+def format_hex(data):
+    HEXDIGITS = '0123456789ABCDEF'
+
+    def hexbyte(v):
+        return HEXDIGITS[v >> 4] + HEXDIGITS[v & 0xF]
+
+    result = ''
+    offset = 0
+    while offset < len(data):
+        result += '%04u |' % offset
+        textpart = ''
+        for i in range(0, 16):
+            if i > 0 and i % 4 == 0:
+                result += ' '
+            if offset < len(data):
+                v = data[offset]
+                result += ' ' + hexbyte(v)
+                textpart += chr(v) if 32 <= v < 127 else '.'
+                offset += 1
+            else:
+                result += '   '
+                textpart += ' '
+        result += ' | ' + textpart + ' |\n'
+
+    return result
 
 
 class ShowUntracked(EditModel):
@@ -2076,8 +2096,7 @@ class ShowUntracked(EditModel):
         try:
             text_result = result.decode(encoding)
         except UnicodeError:
-            text_result = clean_nonprintable(result.decode('ascii',
-                                                           errors='replace'))
+            text_result = format_hex(result)
 
         if truncated:
             text_result += '...'
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -2044,6 +2044,13 @@ class SetUpstreamBranch(ContextCommand):
         cfg.set_repo('branch.%s.merge' % branch, 'refs/heads/' + remote_branch)
 
 
+def clean_nonprintable(s):
+    res = ""
+    for c in s:
+        res += c if ord(c) >= 32 or c == '\n' else '.'
+    return res
+
+
 class ShowUntracked(EditModel):
     """Show an untracked file."""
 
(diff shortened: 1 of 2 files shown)


davvid:
I realized that we already support `gui.encoding` and gitattributes if you use `cfg.file_encoding(filename)` to lookup the encoding, so we could get the best of both worlds by setting `git config --global cola.untrackedencoding ascii` and looking up the encoding like this:

        encoding = (
            cfg.file_encoding(filename)
            or cfg.get('cola.untrackedencoding', core.ENCODING))

I could have suggested that you can use `git config gui.encoding ascii`, but that recommendation would break partially staging utf-8 files, hence the new config variable that affects untracked files only.

wm4:
What exactly does `cfg.file_encoding(filename)` return by default? Also, I'd like this to work without touching the global git config.

Here's another idea: if the file contents do not cleanly convert to utf-8, use "cleaned" ascii as fallback, instead of using the character conversion `ignore` option. Instead of ascii text, we might also just display a hex dump, which would make more sense for binary files. (Is there a guarantee that the text window uses a monospace font?)

> Thanks for the speedups. Do you know of any related Qt bugs that we can link to?

No, that was a guess. When I looked this issue at first, I thought git-cola loaded the entire file into memory or so. But it quickly turned out that it must have been in the GUI somewhere. It makes sense to me that Qt would slow down on random character soup, and I've observed similar problems with other Qt editors as well as non-Qt text renderers (where it simply turned out that looking up and loading all these characters/fonts takes time), but I can't be fully sure.

davvid:
`file_encoding()` returns None by default.  The lookups done by that method all cached so the overhead of calling into it is only paid once.

That sounds like a good idea, and would keep utf-8 text working nicely while also doing something sensible for binary soup.  Nice!

wm4:
I changed it to the new idea. I didn't try to use `file_encoding`, do you want that? (Seems pointless... which developer would still use a non-utf-8 file encoding?)

Since the `core.read` wrapper will use legacy fall back encodings if UTF-8 conversion fails, I added something to make it return bytes. Is this OK this way?

I can also change it to display some sort of hex dump if that's welcome (I'd do this directly in `ShowUntracked.read` and assume a monospace font).

I didn't manage to run `make check`. I'm on Debian unstable. They are phasing out Python 2, and the PyQt wrappers are apparently available only to Python 3. But `make check` seems to explicitly invoke `python` (== python 2.x) in various places, and the shebangs of all the test files also use Python 2. How do I make `make check` use Python 3?

davvid:
`make check PYTHON=python3`

davvid:
That sounds like a good path.  The `bytes` thing is clever and clean.

It's still worth consulting `file_encoding` and using it only if it's non-None -- it's Git feature, and some users might have files in shift-jis or other encodings for historical, geographical, and other reasons.  It would be surprising if we didn't support it here because we do support it elsewhere.  Hopefully it doesn't make the code too complicated.

The hexdump thing sounds great when we detect that it's non-text binary data.  It's safe to assume a monospace font -- we tell Qt that we only want monospace fonts in the font selector, and our default font is the system monospace font when first setup.

The diff viewer also has an Image diff viewer that can display and diff images.  I've kinda wanted to hook that up for untracked files too for the display part, but that's something for the future for sure.  Just something to keep in mind when extending it.

wm4:
Updated. It uses `file_encoding` now. (I didn't test it, but judging by the other code which uses it, it should work.) Also there is a separate commit, which formats it as hex dump if it's not valid UTF-8. Might be a bit inefficient, but not a problem for the default 2KB read size.

One potential issue still exists: the read size could cut the file in the middle of a UTF-8 sequence. Then conversion would fail. Is there a good way to deal make Python ignore cut off trailing sequences?

davvid:
I just ran `make check` after merging locally and it all passed here, so all good.  In case it helps debugging it when you saw it error (maybe it was fixed recently?).  Thanks again.

```
flake8 --version
3.7.9 (mccabe: 0.6.1, pycodestyle: 2.5.0, pyflakes: 2.1.1) CPython 3.7.4 on Darwin
flake8  --max-line-length=80 --format=pylint --doctests  \
        bin/git-cola bin/git-dag share/git-cola/bin/git-xbase setup.py cola test extras contrib
pylint --version
pylint 2.3.1
astroid 2.2.5
Python 3.7.4 (default, Oct 28 2019, 19:58:47)
```

PR #1069, opened 2020-06-09
Hide passphrase

laerreal:
Adds switch to hide (default) passphrase to SSH authentication dialog.

**v2**:
- use `else` in `if` block

### DIFF
--- a/share/git-cola/bin/ssh-askpass
+++ b/share/git-cola/bin/ssh-askpass
@@ -27,7 +27,7 @@ entry .e -textvariable answer -width 50
 pack .e -side top -fill x -padx 10 -pady 10
 
 if {!$yesno} {
-	if {"Password" in $prompt} {
+	if {"Password" in $prompt || "passphrase" in $prompt} {
 		.e configure -show "*"
 	}
 }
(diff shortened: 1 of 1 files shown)


share/git-cola/bin/ssh-askpass
```diff
@@ -26,11 +26,26 @@ pack .m -side top -fill x -padx 20 -pady 20 -expand 1
 entry .e -textvariable answer -width 50
 pack .e -side top -fill x -padx 10 -pady 10
 
-if {!$yesno} {
-	if {"Password" in $prompt} {
+proc on_hide_input_changed {args} {
+	global hide_input
+	if {$hide_input} {
 		.e configure -show "*"
+	} {
```
davvid, line 33: Looks nice!  Hmm.. on this part, how about please include the `else` keyword here so that it's clear to the reader what's going on.  I don't really know tcl that well, but I suspect that maybe always using `else` is preferred -- at least `gitk` does it that way.

The tcl/tk style guide does say, "However, the else word should always be used as it does impart some semantic information and it is more like the C language.".

Looks good otherwise.  Is there a way to do this without a global?  I'm guessing not, so this looks good enough.  Please amend the commit with the `else` fixup and we'll merge this in.

davvid:
Just a note for myself so that I remember to address the notes in https://github.com/prati0100/git-gui/pull/38#pullrequestreview-446793344 and submit the final patch to the git list for review.

davvid:
Upstreamed patch:

https://public-inbox.org/git/20201107222039.13900-1-davvid@gmail.com/T/#u

PR #1073, opened 2020-07-03
Remove workaround no longer needed in pynsist preamble

takluyver:
Since a fix (https://github.com/takluyver/pynsist/pull/149 ) you made in Pynsist a couple of years ago, this workaround is no longer needed.

I suspect you also don't need to add `pythondir` to `PATH` any more - the command launchers do this already, and when launching from a shortcut that's the exe directory, which I think should achieve the same thing. But I haven't checked this carefully, so I'm not touching it at the moment.


davvid:
Oh sweet, thanks @takluyver I'll test out the pythondir removal.  Thanks for the heads up.

PR #1079, opened 2020-08-06
Change /usr/share/appdata to /usr/share/metainfo

guoyunhe:
The latest AppStream standard changed its directory name to support not only apps but also libs, etc.


davvid:
Thanks for the heads-up.  Are there any compat concerns we need to consider -- eg. are there older users that would be impacted?  If there's little concern then I'll merge this shortly.

guoyunhe:
I think no users will be impacted. The change has been there [since 2016](https://github.com/ximion/appstream/blob/master/NEWS#L1363). Users of older distros probably won't get updates of git-cola.

davvid:
It looks like you forgot to rename the directory ~ I fixed that up locally.  Thanks for the patch

PR #1081, opened 2020-08-06
Use remote.pushDefault if default branch is set

### LINKED ISSUE #1078: The remote selection doesn't use the setting from remote.pushDefault
in `remote.pushDefault` I have set a remote called `fork` to be the default one but when I open the git-cola GUI it still has the remote `origin` selected when I try to push.

I would expect to read the `remote.pushDefault` settings and select whatever is in the git config.

thk123:
If a remote.pushDefault is set then use that, rather than origin.

Fixes #1078

I would appreciate some advice on how/where to write a unit test for this new function.

### DIFF
--- a/cola/widgets/remote.py
+++ b/cola/widgets/remote.py
@@ -90,6 +90,15 @@ def strip_remotes(remote_branches):
     return [branch for branch in branches if branch != 'HEAD']
 
 
+def get_default_remote(context):
+    """Get the name of the default remote to use for pushing.
+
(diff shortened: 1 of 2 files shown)


thk123:
@davvid Thanks - I'v added a test as described. When running `make check` locally I run into a bunch of errors:

```
test/git_test.py:8:0: Cannot import 'unittest.mock' due to syntax error 'invalid syntax (<unknown>, line 119)' (syntax-error)
```

But this appears to be a problem with using python 3.8 - hopefully CI will pass.

PR #1086, opened 2020-09-11
Replace startup dialog's bookmark list with icons instead of text list

malpas:
+ Use an icon layout instead of the original text list. This reduces visual clutter from repeated path suffixes. The (minor) trade-off is that two repos with the same name will look the same.
+ Remove the sorting by path that StartupDialog performed. This lets the settings manage the sorting, and its behaviour seems more intuitive.
  + Bookmarks go first.
  + Recent repos go in reverse chronological order. This will visually separate same-name repos over time.
+ Add path tooltips to bookmarks

## Motivation

When I first started using Git Cola, the startup page felt a bit cluttered to me, so I finally decided to see if I could polish it.

## Pictures

Before:
old-startup-dialog

After:
new-startup-dialog

### DIFF
--- a/cola/widgets/startup.py
+++ b/cola/widgets/startup.py
@@ -69,7 +69,7 @@ class StartupDialog(standard.Dialog):
         all_repos = bookmarks + recent
 
         added = set()
-        for repo in sorted(all_repos, key=lambda repo: repo['path'].lower()):
+        for repo in all_repos:
             if repo['path'] in added:
                 continue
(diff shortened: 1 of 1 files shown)


davvid:
This patch is beautiful in every way, thanks for putting this together.  I'll merge this shortly, this is really great.

I'm going to tag `v3.8` right now so that this can be the first feature for the subsequent release.

PR #1088, opened 2020-09-24
Fix Startup bookmark view not showing folders' short paths on launch

malpas:
This does not affect the bookmark view if list mode is the user's startup preference.


davvid:
Thanks for catching this.

PR #1115, opened 2021-02-25
Fix test for signed commits

harmathy:
If git is configured to sign commits, the signature will be shown by git log,
thus breaking this test. The --no-show-signature parameter avoids printing
the signature.

### DIFF
--- a/test/gitops_test.py
+++ b/test/gitops_test.py
@@ -14,7 +14,7 @@ class ColaBasicGitTestCase(helper.GitRepositoryTestCase):
         self.run_git('add', 'A', 'B')
 
         self.git.commit(m='initial commit')
-        log = self.run_git('log', '--pretty=oneline')
+        log = self.run_git('log', '--pretty=oneline', '--no-show-signature')
 
         self.assertEqual(len(log.splitlines()), 1)
 


davvid (changes requested):
Thanks for the PR, and sorry for taking so long to get to it.

`--no-show-signature` was introduced in git v2.10.0, which not everyone will have yet.

Please add a new entry in `cola/version.py` for `no-show-signature` and then add a `if version.check_git(self.context, 'no-show-signature'): ...` here to decide whether or not to use the flag.

cola is maximally compatible with git versions, so the `check_git(...)` facility makes it easy to handle old versions.  Thanks for catching this.

harmathy:
I would like to see the CI go red, before making the suggested change. So it would be nice, if there was a CI job for checking git compatibility for different versions. I'll have a look into this and possibly open another PR.

harmathy:
I would like to see the CI go red, before making the suggested change. So it would be nice, if there was a CI job for checking git compatibility for different versions. I'll have a look into this and possibly open another PR.

davvid:
I merged this in after adding a small tweak to the test to use `git -c log.showsignature=false log ...` instead of `git log ...` and that satisfies the requirement for supporting older versions of Git.

Please don't let that discourage you in case you're able to come up with a Github Actions matrix for testing multiple versions of Git. These fixes are helpful improvements so I figured it'd be worth doing, but it does mean that you may not see it go red in CI unless another version-dependent test were to be added.

Extensive multi-version Git tests being run via CI would be pretty awesome. Git v1.6.3 is old enough that most users don't have it, but we still technically support it so having that enforced by CI is indeed the way to go.

Here are some Git versions that seem like they would be interesting to test:

- `v1.6.3` is Cola's oldest supported Git version
- `v1.8.6` for RHEL 7
- `v2.0.0` was the first tagged GIt v2 release
- `v2.5.0` is 5 minor releases later.
- `v2.10.0` is 5 minor releases later.
- `v2.15.0` is 5 minor releases later.
- `v2.20.1` for Debian/buster
- `v2.24.1` for Ubuntu focal/20.04
- `v2.27.0` for RHEL8
- `v2.30.1` for Debian/bullseye (testing)
- `v2.32.0` for Debian/sed (unstable)
- latest stable/tagged git release

PR #1119, opened 2021-03-12
Allow cola to be started as a module

florisla:
I took the easiest route (for me, on Windows) to get cola to run from a git checkout.

    python3.9 -m venv .venv
    .\.venv\scripts\activate
    pip install .
    pip install pyqt5

Then, I lacked an easy to find way to launch cola.

This offers such a way.  It allows to do:

    python -m cola

### DIFF
--- a/cola/__main__.py
+++ b/cola/__main__.py
@@ -1,0 +1,19 @@
+"""
+Run cola as a Python module.
+
+Usage:
+
+    python -m cola
+
+"""
+
+from cola import main
+
+
+def run():
+    """Start the command-line interface."""
+    main.main()
+
+
+if __name__ == "__main__":
+    run()
--- a/cola/__main__.py
+++ b/cola/__main__.py
@@ -7,6 +7,9 @@ Usage:
 
 """
 
+from __future__ import absolute_import, division, unicode_literals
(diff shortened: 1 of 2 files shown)


davvid:
This looks really cool.  Usually I use the `bin/` launcher scripts and have a symlink at `~/bin/git-cola` pointing to the script inside a clone at `~/src/git-cola/bin/git-cola`. `$HOME/bin` is in my `$PATH` so it all works out. I like the idea of having this way as well.

cola/__main__.py
```diff
@@ -0,0 +1,19 @@
+"""
+Run cola as a Python module.
+
+Usage:
+
+    python -m cola
+
+"""
```
davvid, line 8: All of our source file start with the same import, and our pylint checks enforce it.  Please add this line immediately after the docstring:

    from __future__ import absolute_import, division, unicode_literals

Sorry about that.

davvid (changes requested):
Nicely done.  I wonder if we should include a blurb in the readme about it. That can always come in a follow-up after this initial change if you prefer.

ramonbrugman (approved):

PR #1122, opened 2021-03-15
Start cola through console_script entry points

florisla:
This is related to #1119, it offers another additional way to start cola.

I'm not totally sure it won't cause any side effects, but to me this is a nice and modern way to start a Python program.

This makes use of [entry points](https://packaging.python.org/specifications/entry-points/).  On Linux these are Python shebanged files in Python's `bin` folder, on Windows these are small `.exe` files linked against `python.dll` in the `Scripts` folder.

Advantages:
* No need for Bash scripts on Windows.  Some users don't have Git Bash available.
* No need to search for a Python executable
* No need to track Python versions in the shebang
* Plays well with virtual environments and [pipx](https://github.com/pipxproject/pipx)
* Higher chance of the folder being already present in `$PATH`

In short, if the package is `pip`-installed then you use the entry point instead of a shell script or shebanged Python file.

Stuff to investigate:
* Do old releases of `pip` actually support these entry points?  This could be tested through Travis.
* What's the risk of users having `cola` twice in their `$PATH`?
* ... ?

### DIFF
--- a/cola/__main__.py
+++ b/cola/__main__.py
@@ -8,12 +8,19 @@ Usage:
 """
 from __future__ import absolute_import, division, unicode_literals
 
+import sys
+
 from cola import main
 
 
 def run():
     """Start the command-line interface."""
     main.main()
+
+
+def run_dag():
+    args = ["dag"] + sys.argv[1:]
+    main.main(args)
 
 
 if __name__ == '__main__':
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,3 +39,10 @@ testing =
 	# local
 	PyQt5; python_version>='3.0'
 	PySide2; python_version=='2.7'
+
+[options.entry_points]
+console_scripts = 
+	cola = cola.__main__:run
+	dag = cola.__main__:run_dag
+	git-cola = cola.__main__:run
+	git-dag = cola.__main__:run_dag


florisla:
The goal would be to make cola installable through pipx.  Either from a git repository / branch

    pipx install git+https://github.com/florisla/git-cola.git@entry-points

or from PyPI

    pipx install git-cola

Currently the git install succeeds but the virtual environment is missing PyQt5 and Qtpy.  Also, the application icons are not installed.

What do you think about this approach?

setup.cfg
```diff
@@ -39,3 +39,10 @@ testing =
 	# local
 	PyQt5; python_version>='3.0'
 	PySide2; python_version=='2.7'
+
+[options.entry_points]
+console_scripts = 
+	cola = cola.__main__:run
+	dag = cola.__main__:run_dag
```
davvid, line 46: I think we've been asked to not use the generic `dag` name -- so we should drop that one.

cola/__main__.py
```diff
@@ -8,6 +8,8 @@
 """
 from __future__ import absolute_import, division, unicode_literals
```
davvid, line 10: Drop this blank line `__future__` is part of the regular stdlib imports so should be grouped alongside `sys`

cola/__main__.py
```diff
@@ -16,5 +18,10 @@ def run():
     main.main()
 
 
+def run_dag():
+    args = ["dag"] + sys.argv[1:]
+    main.main(args)
```
davvid, line 23: I think it'd be better to `from cola import dag` and then `dag.main` here, which matches what `bin/git-dag` does.

setup.cfg
```diff
@@ -39,3 +39,10 @@ testing =
...
+	git-cola = cola.__main__:run
+	git-dag = cola.__main__:run_dag
```
davvid, line 48: What about the makefile and the scripts in `bin/`?  Won't these clash? Do we have to inhibit the installation of the `bin/` scripts?

davvid: One thing that might not be readily apparent is that we currently can run directly from the source tree without any python or build system getting in our way. In other words, I can clone the repo and directly run `./git-cola/bin/git-cola` and it works.

The bin/ scripts currently also handle the Windows installer `$prefix/pkgs/` path where modules get installed. I'm curious if that works -- have you tested generating a win32 installer and setting if the installation works when using the entry points?

florisla: Yes I do think this kind of competes with the scripts in `bin/`.  But at this point this is is not yet polished enough to do that.

Running straight from the repo is only possible if you already have PyQt5 (or an alternative) available in the *right* Python installation.  Installing through pip could potentially solve that.

I did not consider the Windows installer yet.

florisla:
Thanks for the review, I'll look into it.

davvid:
Hello. We're now dropping python2 from the build system and I'm going to be looking into likey replacing distutils usage with setuputils (or some other newer approach).

I think we'll probably switch to entry points now, so I may use your branch here as inspiration. I think we'd be okay with a setup that resulted in `python setup.py develop` inside of a virtualenv being an acceptable workflow, and that should enable pipx as well.

It seems like we'll probably have to drop our `./bin/` wrapper scripts and just let setuptools/etc. handle it.

Sorry I haven't been able to look at this one since the last update. Unless you beat me to it I expect that we should be able to close this MR alongside #1201. 

We also have a `cola/__main__.py` these days and the README already mentions `python -m cola` as a supported way of launching it from the source tree. That's probably better than the `./bin/` scripts.

Switching over to entry points and being pip-installable is basically the goal now.

davvid:
Thanks again! `pip install git-cola` is now a thing! 😄 

https://pypi.org/project/git-cola/

PR #1123, opened 2021-03-17
Added word wrapping

abid1998:
Changes in this Merge Request:-

Added code to enable word wrapping and made changes in UI (diff options).

### DIFF
--- a/cola/widgets/diff.py
+++ b/cola/widgets/diff.py
@@ -589,6 +589,7 @@ class Options(QtWidgets.QWidget):
         )
 
         self.show_line_numbers = self.add_option(N_('Show line numbers'))
+        self.enable_word_wrap = self.add_option(N_('Enable Word Wrapping'))
 
         self.options = qtutils.create_action_button(
             tooltip=N_('Diff Options'), icon=icons.configure()
@@ -621,6 +622,7 @@ class Options(QtWidgets.QWidget):
         menu.addAction(self.ignore_space_change)
         menu.addAction(self.ignore_all_space)
         menu.addAction(self.show_line_numbers)
+        menu.addAction(self.enable_word_wrap)
         menu.addAction(self.function_context)
 
         # Layouts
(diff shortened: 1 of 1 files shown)


abid1998:
Can someone help me with check ?

abid1998:
WordWrap
Demo of feature.

cola/widgets/diff.py
```diff
@@ -745,8 +753,19 @@ def show_line_numbers(self):
...
+
     def update_options(self):
         self.numbers.setVisible(self.show_line_numbers())
+        if self.is_enable_word_wrapping_checked():
+            self.setWordWrapMode(QtGui.QTextOption.WordWrap)
+            self.setLineWrapMode(QtWidgets.QPlainTextEdit.WidgetWidth)
+            self.options.enable_word_wrap.setChecked(True)
+        else:
+            self.setWordWrapMode(QtGui.QTextOption.NoWrap)
+            self.setLineWrapMode(QtWidgets.QPlainTextEdit.NoWrap)
```
davvid, line 767: Please move the Qt `self.set{Word,Line}WrapMode` calls into a standalone `def set_word_wrap_enabled(self, enabled):` method and call it from here.. maybe as `self.set_word_wrap_enabled(self.is_enable_word_wrapping_check())`

It's kinda funky that we check the checkbox and then set the checkbox to the same value.. will have to see why that's setup that way.

This should probably handle the `apply_state()` / `export_state()` callbacks so that this preference persists across application restarts. eg. if you disable it, it should stay disabled when restarting cola.

cola/widgets/diff.py
```diff
@@ -689,6 +691,12 @@ def __init__(self, context, options, parent):
 
         # "Diff Options" tool menu
         self.options = options
+
+        # Enabling Word Wrapping by default
+        self.setWordWrapMode(QtGui.QTextOption.WordWrap)
+        self.setLineWrapMode(QtWidgets.QPlainTextEdit.WidgetWidth)
+        self.options.enable_word_wrap.setChecked(True)
+
```
davvid, line 699: If we add the standalone function then this section should also condense down into a single call.

davvid:
I went ahead and merged this in after applying a few local tweaks on top.

I ended up using dedicated callbacks (that don't call update_options() aka set_options()) for these checkboxes and integrated the new one into the export_state/apply_state() handlers.

BTW, please let me know if you don't mind adding your full name (rather than abid1998) to doc/thanks.rst.  Thanks again!

abid1998:
Sure i won't mind :D. Full Name:- Abid Ahmad

abid1998:
Just went through your commit will have to learn some more Qt before i could've thought of that haha.

PR #1131, opened 2021-04-06
Status widget: automatically select topmost header

### LINKED ISSUE #1130: Feature request: Select staged changes when all files are staged
I would like the diff view to display staged changes automatically when staging all.

My workflow is usually like this:

1. Make some changes in my repo
2. Hit Ctrl+K to focus the Status widget
3. Hit Ctrl+A then Ctrl+S to select and stage all files
4. (All files get staged, selection in Status widget is cleared, Diff widget goes blank)
5. Hit ArrowDown to select the "Staged" line in Status widget
6. Hit Ctrl+L to focus Commit widget
7. Create commit etc.

My issue is with step 4 and 5 here. When files are staged, it's unhelpful that the Status widget selection (and, consequently, the Diff view) is cleared.

In my opinion, it would be better if staged changes are selected automatically in the Status widget.

image

birgersp:
Staging/unstaging files will in some cases clear the selection
of the Status widget. Consequently, the diff view is also blank.

With this patch the StatusTreeWidget._restore_selection will
automatically select the (topmost) header if no other item is
selected.

This fixes #1130

### DIFF
--- a/cola/widgets/status.py
+++ b/cola/widgets/status.py
@@ -412,6 +412,9 @@ class StatusTreeWidget(QtWidgets.QTreeWidget):
                         reselect(j, current=True)
                         return
 
+        # No item was selected, select (topmost) header
+        self.select_header()
+
     def _restore_scrollbars(self):
         vscroll = self.verticalScrollBar()
         if vscroll and self.old_vscroll is not None:


davvid:
Hmm, selecting the top-most header can be a slow operation; a simpler alternative might be to select the first item instead of the header.

If we wanted to be really fancy we would account for the specific cases where the selection can be cleared and make it do something that is a bit more nuanced.

When staging/unstaging files we currently have a convenient behavior where the next file in the same category (Staged/Modified/Untracked) will be selected when a file is staged w/ ctrl-s / s. This doesn't account for what should happen when the last file in that section is acted upon. Currently this is where we lose the selection.

I'm only considering the behavior when operating on one file at a time. When multiple files are selected then it might make sense to try and transfer the selection over to the new category for each of the acted-upon files.

For single files, when we have both Modifed and Untracked files, and we stage the last Untracked file, then we should select the first Modified file. The idea is that the user is still performing staging operations, so if there are no more Untracked files then it makes sense to go to the top of the Modified section.

If there are no Modified files, then the same file that was previously Untracked and is now Staged should be selected in the staged section.

When staging the last Untracked file, causing it to lose the selection today, we should instead select either the first Modified file or the same file in the Staged section if there are no modified files.

When unstaging the last Staged file, causing it to lose the selection, select either the same file in the Modified section, or the same file in the Untracked section if it doesn't exist in the modified section.

Existing files are special as well. When staging the last Modified file, we should select that same file in the Staged section.  We do not go to the untracked section. The rationale is that it's common to have unstaged cruft (binaries, build artifacts) in a worktree, so we optimize for tracked files that are known to git.

--- The above is a stream of thought on how I think it should work. I'll leave this PR open since I think there's a way to make it work in a more intuitive way, but I don't think we can have it just select the top-most header item.

What do you think? Are there workflows where someone would want to the topmost header selected, where we might want to have a preference to control this behavior? In general we try to avoid preferences in favor of intuitive / specially-tuned behavior in support of the keyboard-driven cola workflow, so hopefully not.

birgersp:
I agree, it's probably best to select the previously selected file. And I don't really think it needs a preference, since it is quite intuitive for the user that the selection remained on the current file (even though it moved from unstaged to staged).

And if mulitple files is selected, I guess it should be possible to re-select those multiple files when staging them.

Also I should add; I've been using this patch for a couple of weeks now and it is isn't really working as intended. So this PR would need some more work either way.

So here is a description of the new, proposed functionality:

In selection update; If no files are currently selected, obtain the file or files that were previously selected and re-select them.

davvid:
Please test out the latest development version from the main branch. This idea has been implemented and now we're much better about retaining the selection.

Thanks again for the heads-up and suggestions. There are likely still some edge cases that could be improved, for example the current implementation is a bit single-file centric, but this should be an improvement all around and should catch the 80%+ majority use cases.

These kinds of changes take some time to make perfect so I tagged the latest stable release before these updates were merged. This enhancement will be released in the next `v3.11` minor release. If we have other edge case enhancements then we'll probably want to save them for `v3.12` to give them adequate time to bake.

birgersp:
Thanks @davvid, I really appreciate you implementating this and I'm confident others will find it useful too! Im currently AFK these days because of holidays and travel but I'll be sure to test once I get the chance.

PR #1137, opened 2021-05-03
add `--includes` when reading user git config

### LINKED ISSUE #1136: `.gitconfig.user` file ignored when reading git config
Hi, my `~/.gitconfig` file has a include directive in the end:

    [include]
        path = .gitconfig.user

When git-cola reads configuration from `~/.gitconfig` file, it ignores this directive, because it's the default behavior of 
`git config --file` according to [docs](https://git-scm.com/docs/git-config#Documentation/git-config.txt---no-includes).

Adding `--includes` to `git config --file` fix the issue.

I'll be sending a PR soon with a possible fix.

Thank's to the awesome tool!

nandalopes:
Add `--includes` option only when reading user config.

From [`git config` doc](https://git-scm.com/docs/git-config#Documentation/git-config.txt---no-includes):

    --[no-]includes
    Respect include.* directives in config files when looking up values.
    Defaults to off when a specific file is given (e.g., using --file,
    --global, etc) and on when searching all config files.

Fix #1136

### DIFF
--- a/cola/gitcfg.py
+++ b/cola/gitcfg.py
@@ -124,7 +124,6 @@ class GitConfig(observable.Observable):
         self._cache_key = None
         self._configs = []
         self._config_files = {}
-        self._context = context
         self._attr_cache = {}
         self._find_config_files()
 
@@ -213,7 +212,7 @@ class GitConfig(observable.Observable):
             return self._read_config_file(path)
 
         dest = {}
-        if version.check_git(self._context, 'config-includes'):
+        if version.check_git(self, 'config-includes'):
             args = ('--null', '--file', path, '--list', '--includes')
         else:
             args = ('--null', '--file', path, '--list')
--- a/cola/gitcfg.py
+++ b/cola/gitcfg.py
@@ -10,6 +10,7 @@ import struct
 from . import core
 from . import observable
 from . import utils
+from . import version
 from .compat import int_types
 from .git import STDOUT
 from .compat import ustr
@@ -123,6 +124,7 @@ class GitConfig(observable.Observable):
         self._cache_key = None
         self._configs = []
         self._config_files = {}
+        self._context = context
         self._attr_cache = {}
         self._find_config_files()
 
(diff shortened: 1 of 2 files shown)


nandalopes:
This PR fix #1136 issue

cola/gitcfg.py
```diff
@@ -204,14 +206,14 @@ def update(self):
...
-    def read_config(self, path):
+    def read_config(self, path, *opt_args):
         """Return git config data from a path as a dictionary."""
 
         if BUILTIN_READER:
             return self._read_config_file(path)
 
         dest = {}
-        args = ('--null', '--file', path, '--list')
+        args = ('--null', '--file', path, '--list') + opt_args
```
davvid, line 216: Thanks for adding this. I think we would always want to use `--includes`, even for system and repo configs, so it doesn't seem worth making this optional at the `read_config(...)` level. In other words, can be just add the logic here rather than having the callers worry about it?

Also, this feature was added in git 1.7.10. Can you please add a `config-includes` entry to `cola/version.py` and then have a runtime check here where we check `if version.check_git(context, 'config-includes'): ...` in order to decide whether or not to use the flag? That'll make the code compatible with older versions.

BTW, the context is passed to the gitcfg constructor but we don't hold onto it. You'll need to store it in a member variable in the constructor (eg. `self._context = context`) so that we can use it to query git here.

davvid (changes requested):
Thanks again. When you update, please also sign-off your commits (`git commit --amend --signoff` or the `ctrl-i` hotkey in the cola commit message editor).

cola/gitcfg.py
nandalopes: I've made the change, grateful for your feedback.

nandalopes:
Seems like holding context create another issue:

```

  File "/home/nanda/.local/share/git-cola/lib/cola/cmds.py", line 2934, in do
    return cmd.do()

  File "/home/nanda/.local/share/git-cola/lib/cola/cmds.py", line 2015, in do
    self.cfg.update()

  File "/home/nanda/.local/share/git-cola/lib/cola/gitcfg.py", line 206, in update

  File "/home/nanda/.local/share/git-cola/lib/cola/observable.py", line 31, in notify_observers
    method(*args, **opts)

  File "/home/nanda/.local/share/git-cola/lib/cola/widgets/commitmsg.py", line 590, in emit_config_updated
    self.config_updated.emit()

RuntimeError: wrapped C/C++ object of type MessageValidator has been deleted
```

What if we push git version itself to 1.7.10?

davvid:
Hmm that's weird about holding onto the context. I didn't realize that's why you had made that change. I cherry-picked that commit back -- sorry for not noticing earlier. I had originally reverted it in 440d622ddc1388ab793fdbf97cb01c187e3db4dd but I just noticed your comment and re-applied it.

PR #1138, opened 2021-05-04
read files included with include.path config

### LINKED ISSUE #1136: `.gitconfig.user` file ignored when reading git config

nandalopes:
Solve #1136

- set git minimal version to 1.7.10
- Fix #1136 - read files included in ~/.gitconfig

add `--includes` arg when reading git config

From [git config doc](https://git-scm.com/docs/git-config#Documentation/git-config.txt---no-includes):

    --[no-]includes:
    Respect include.* directives in config files when looking up values.
    Defaults to off when a specific file is given (e.g., using --file, --global,
    etc) and on when searching all config files.

Git version is `> 2.17` on Ubuntu Bionic, and `> 2.1.4` on Debian Jessie (oldoldstable), just for comparison.

### DIFF
--- a/cola/gitcfg.py
+++ b/cola/gitcfg.py
@@ -211,7 +211,7 @@ class GitConfig(observable.Observable):
             return self._read_config_file(path)
 
         dest = {}
-        args = ('--null', '--file', path, '--list')
+        args = ('--null', '--file', path, '--list', '--includes')
         config_lines = self.git.config(*args)[STDOUT].split('\0')
         for line in config_lines:
             if not line:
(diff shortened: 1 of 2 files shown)


davvid:
We resolved this in #1137 .  Thanks again!

PR #1143, opened 2021-06-08
call git-cola-sequence-editor without python

bobvanderlinden:
Currently the executable `git-cola-sequence-editor` is called always using python. However, some distros need to have this executable replaced with a wrapping script in order to inject the right environment variables. The wrapping script is a bash script and thus failed to be executed using Python. See https://github.com/NixOS/nixpkgs/issues/126131

For such cases it helps to run the executable as-is without explicitly calling Python.

I'm not sure whether this is wanted behavior in general, but this PR attempts to do this for `git-cola-sequence-editor`.

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -1830,7 +1830,7 @@ def unix_path(path, is_win32=utils.is_win32):
 def sequence_editor():
     """Set GIT_SEQUENCE_EDITOR for running git-cola-sequence-editor"""
     xbase = unix_path(resources.command('git-cola-sequence-editor'))
-    editor = core.list2cmdline([unix_path(sys.executable), xbase])
+    editor = core.list2cmdline([xbase])
     return editor
 
 


davvid (changes requested):
Hmm I think the only platform that might be hurt by this would be windows. I'm not a Windows user so I can't test it to be sure, but maybe we can continue to use the interp on Windows?

Checking for `if utils.is_win32():` here and then making the new behavior conditional seems like it would make sense so that we don't have to worry about that.

For poor python souls who sometimes use a mix of interpreters ~ python2.7, 3.6, 3.7, 3.8, ... this could be a downside when hacking on the code because we may have done `python3 ./bin/git-cola` and since we use neutral `#!/usr/bin/env python` shebang lines, it'll now use the old py2 interpreter rather than the one we originally used. That's fine IMO, just noting that this is reason we were doing this originally.

This is a dev concern only; users never see this because the install scripts are setup to replace the `#!` lines unless a flag is passed to allow the original env lines to remain.

I wonder if our in-source `#!` lines are going to make hacking on the code inconvenient on eg. rhel8 or others where `python` resolves to nothing and it's up to the user to configure a symlink or use a virtualenv to make `python` resolve to their interpreter of choice. In another 2-4 years we'll consider changing our in-source `#!` lines to `#!/usr/bin/env python3`, though, which'll eliminate most instances of where this would be an issue in practice.

Hacking on the project is easiest in a virtualenv and it won't harm that scenario at all so overall it's fine, especially since it makes things better for nix.

bobvanderlinden:
Thanks for picking up the Windows change :partying_face:

Issue #1084, opened 2020-09-04
"Stage selected lines" for untracked files


mighty-d:
Would it be possible to enable staging selected lines from an untracked file?
Currently an untracked file needs to be staged in its entirety and then unwanted lines must be unstaged.

living180:
We could probably add a "Begin tracking" context menu item for untracked files, which would invoke `git add --intent-to-track` on the selected file to add it to the index but without staging any of its contents.  Then selected lines could be staged as usual.

davvid:
Yup, exactly that -- after the initial `git add --intent-to-add` staging will "just work".  A "Begin tracking" context menu makes a lot of sense.

This is basically the same request as #493 and would resolve that issue as well.

mighty-d:
Not sure what the general direction of the project is here, but is it necessary to have one action per git command?
Begin tracking implies a need for "end tracking".

This is how it works in gitextensions, the software I used before:
changed and untracked are all together, they just have with a different icon (green + for untracked and pencil for edited, if I recall)

user: select a file
user: select some wanted lines
user: stage lines (S)
system: track if untracked
system: stage lines


From the staging area (which is in a different gui widget, so ctrl+a it's actually useful):
user: select a file
user: select some unwanted lines
user: unstage lines (U)
system: unstage lines
system: untrack if untracked and file is empty

Also revert (R) works the same and if the file is untracked lines get deleted instead of going to the previous version.
I find that workflow quite handy for precision commit crafting.

davvid:
👍 thank you, that's a great idea.  Indeed, it would be nice if we could make this seamless for users.

On Unix we can "fake" a full diff by leveraging `/dev/null` when diffing:
```
git diff --no-index /dev/null <untracked-file>
```
We could build upon that to create a diff-centric untracked file interface that takes care of doing `git add --intent-to-add <untracked-file>` under the hood automatically when doing partial staging for an untracked file.

Is there's an equivalent to `/dev/null` on Windows?  I guess we can always disable this feature on Windows, but it would be nice if there were a simple approach that worked cross-platform.

I'll have to dust off my old Windows VM and see if Git for Windows supports using `/dev/null` like that -- it might, which would be really cool.

PR #1146, opened 2021-07-07
Diff: Allow partial stage/revert for untracked file

mmargoliono:
Currently an untracked file needs to be staged first and then unwanted lines must be unstaged. This change will allow "Stage selected line" and "Revert selected lines..." to be used on untracked file.

The logic will be as followed:
When the user select "Stage selected line" on an untracked file, these selected line will be added to staging, and the rest of file will then be listed under modified. When the user select "Revert selected lines.." on an untracked file, those line will be deleted from the file

To generate the diff of the new file, dev null is used. os.devnull is used instead of hard-coding '/dev/null/'. This will ensure it will works in windows environment, where 'nul' is needed instead of '/dev/null'

The downside of using this approach is the return code. git diff will return `1` when used against new file or file outside of index. Alternate check condition is used by checking "new file mode" string from the output. This should be enough to differentiate it from genuine error such as filepath not found

Issue: #1084
Reported-By: mighty-d
Signed-off-by: mmargoliono

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -2255,7 +2255,9 @@ class ShowUntracked(EditModel):
         super(ShowUntracked, self).__init__(context)
         self.new_filename = filename
         self.new_mode = self.model.mode_untracked
-        self.new_diff_text = self.read(filename)
+        self.new_diff_text = gitcmds.diff_helper(
+            self.context, filename=filename, cached=False, untracked=True
+        )
         self.new_diff_type = main.Types.TEXT
         self.new_file_type = main.Types.TEXT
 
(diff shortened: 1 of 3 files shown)


mmargoliono:
There is an existing issue where if you stage untracked file, and then select "Unstage selected lines" for all the lines, it will not unstage the file. In that scenario, the new file will still be listed under modified and staged. Partially staged lines using these change will have the same issue.

davvid:
Very nicely done! Thanks again for your help.

davvid:
I made some edits on top to better detect binary files. #1045 added some really nice handling for binary files (a hexdump visualization) so I made it so that the untracked diffing is only enabled when the file isn't detected as a binary file.

We use the same heuristic as Git for detecting binary files ("does NULL exist in the first 8000 bytes?") and we allow overriding our default detection via `.gitattributes` by setting, for example:

```
*.png binary
*.other -binary
```

This marks `*.png` file as binary, and `*.other` as text by using `-binary` to unset the "binary" flag.

Issue #1110, opened 2021-02-03
Staged diffs aren't displayed in a repository without commits


kyanres:
Hello!

**To reproduce :**
```
mkdir test && cd test
git init
echo -e "a\nb" > foo
git cola
```
Then use git cola to stage the file.

**Observed behavior :**
When foo is Untracked, the Diff view shows its contents.
When foo is Staged, the Diff view is empty.

**Expected behavior :**
I need to see what's staged before committing  :-) + to be able to unstage some lines.

Hope that helps.
I got around that problem by doing a first commit for .gitignore and squashing it later with rebase -i --root

git version 2.25.1
cola version 3.6

nyanpasu64:
Still an issue on commit f4cdf65dcb950a761d0ff79149ad89caee32ab2a.

This occurs in all repositories with no HEAD (both newly `git init` repos, and `git checkout --orphan $branch_name`).

nyanpasu64:
Using the git-cola-git AUR package, pointing to main 74e94ae1fc4299d36adbb1cc456df9a674fa72b3. Now showing a file diff works, but staged images don't show anything. And if I stage an image, quit and reopen Git Cola, and scroll on the image, the app crashes:

```
Traceback (most recent call last):
  File "/usr/share/git-cola/lib/cola/widgets/imageview.py", line 330, in wheelEvent
    self.zoomROITo(scene_pos, sign)
  File "/usr/share/git-cola/lib/cola/widgets/imageview.py", line 214, in zoomROITo
    roi_dims = QtCore.QPointF(roi.width(), roi.height())
AttributeError: 'NoneType' object has no attribute 'width'
fish: Job 1, 'cola' terminated by signal SIGABRT (Abort)
```

davvid:
Thanks for the heads-up @nyanpasu64  -- I just pushed b63bc427413b51622fe3088b2423119d4f65db44 which should guard against the scenario you ran into. Interesting I couldn't reproduce it, but I added some extra guard checks.

BTW, I noticed the traceback above says `/usr/share/...` but if you're running from the source tree it should have a path that points into your git clone. You can verify this in the `Help -> About` dialog by checking the listed version.

Nonetheless, this should be better behaved going forward. Cheers

nyanpasu64:
I'm using the `git-cola-git` AUR package, pointing to the newly pushed b63bc427413b51622fe3088b2423119d4f65db44:

[...]

PR #1149, opened 2021-07-12
Diff: Fixed staged file diff in a commit-less repo

mmargoliono:
When there is no commit in the repo, there is no HEAD revision. Therefore the diff command returns an error (fatal: bad revision 'head'), and git-cola then use empty text as the diff fallback.

To fix the issue, the diff_helper is updated to check if `ref` exists, and if `ref` doesn't exist, we drop it and not pass it to the git diff. The check is done through the `rev_parse` wrapper in `gitcmds`, which will simply return back `ref`, when it is not in git.

When viewing the diff of a staged file in a commit-less repo, the ref would be 'HEAD'. Dropping 'HEAD' from parameter will make git compare the staged file against file in the working directory.

Issue: #1110
Reported-By: kyanres
Signed-off-by: mmargoliono

### DIFF
--- a/cola/gitcmds.py
+++ b/cola/gitcmds.py
@@ -417,8 +417,14 @@ def diff_helper(
     if ref and endref:
         argv.append('%s..%s' % (ref, endref))
     elif ref:
-        for r in utils.shell_split(ref.strip()):
-            argv.append(r)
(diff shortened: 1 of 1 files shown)


davvid:
Thanks again.

PR #1163, opened 2021-09-16
doc: note editor-awareness in diff screen, too

scop:
(Removing backticks around the "grep" because the other instance in this file doesn't have that either, and for the "diff" view they'd be out of place, and having them different here would look odd.)


davvid:
Thanks!

Issue #1167, opened 2021-09-20
Replace "Select manually..." with "Browse..." in the open respository dialog.


safield:
The standard language used for this operation is "Browse". The term is quite ubiquitous.

jackrsteiner:
@safield I'm pretty sure I see where this is coming from. Name attribute in calls to QFileDialog are empty strings (caption='') in [qtpy/compat](https://github.com/git-cola/git-cola/blob/main/qtpy/compat.py#L78).

So button labels must be showing defaults based on your system. On my machine, these buttons have reasonable labels, generated from my platform native file browser. I'm using Ubuntu 20.04. Can you clarify, what platform you're on and where exactly you're seeing this? I suspect your system might be different? Maybe the Qt libraries are using a custom file browser.

I think the solution is to set default caption to desired text in https://github.com/git-cola/git-cola/blob/main/cola/qtutils.py#L443

`def opendir_dialog(path, caption='Browse...'):`

But this only impacts systems that are using platform specific Qt libraries that do not use the platform native file browser dialog. So, I think this is why I can't test it.

atharvapatil123:
Hi @safield, Is the issue still open?

meaviral17:
Hi @safield just sent my PR. Please check if it functions properly.

davvid:
Sorry, I should have closed this issue! My mistake. This was resolved a little bit ago.

PR #1171, opened 2021-10-02
Set caption default for opendir_dialog

jackrsteiner:
Changes the caption default value to 'Browse...' for opendir_dialog as per issue #1167. I think this will only impact systems that use Qt implemented file browsers, not system using the system default file browser.

### DIFF
--- a/cola/qtutils.py
+++ b/cola/qtutils.py
@@ -440,7 +440,7 @@ def open_files(title, directory=None, filters=''):
     return result[0]
 
 
-def opendir_dialog(caption, path):
+def opendir_dialog(path, caption='Browse...'):
     """Prompts for a directory path"""
 
     options = (


jackrsteiner:
As this change I think has no impact on my system, whoever is seeking this change probably needs to test it first.

cola/qtutils.py
```diff
@@ -440,7 +440,7 @@ def open_files(title, directory=None, filters=''):
     return result[0]
 
 
-def opendir_dialog(caption, path):
+def opendir_dialog(path, caption='Browse...'):
```
davvid, line 443: I think this would have to updated at the call site since different call sites have different captions. I guess `Browse...` would be okay. I figured the more specific caption was helpful.

Would an alternative for that issue be to use a more specific caption such as, "Open Git Repository..."?

Alternatively, if we wanted to make most of the call sites not specify a value, then it'd be best to use the value of `None` as the default value here, and then when it's `None` we can call `N_('Browse...')` to get the caption within the function body.

We have to defer the call to `N_(...)` until after the translations have been activated so we cannot `N_(...)` when providing default values for keyword arguments. Keyword arg defaults are global variables evaluated at module scope (not deferred) which is why we'd have to use the `None` trick to defer it.

davvid:
Sorry for not noticing this open PR. This was fixed in bcdbf6501b12632461dce9973bd9700ca3b17257 and further refined when @nakanoi updated the startup dialog in https://github.com/git-cola/git-cola/pull/1280

PR #1183, opened 2021-11-12
about: use setWidth(int)

mp-chet:
Passing floats to setWidth() is not supported. It worked by accident
in older versions of PyQt.

This is a follow-up commit to 1b10542ec32c1ce798257a758975c564ff9ba893.

### DIFF
--- a/cola/widgets/about.py
+++ b/cola/widgets/about.py
@@ -36,7 +36,7 @@ class ExpandingTabBar(QtWidgets.QTabBar):
     """
 
     def tabSizeHint(self, tab_index):
-        width = self.parent().width() / max(1, self.count()) - 1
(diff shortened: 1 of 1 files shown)


cola/widgets/about.py
```diff
@@ -36,7 +36,7 @@ class ExpandingTabBar(QtWidgets.QTabBar):
     """
 
     def tabSizeHint(self, tab_index):
-        width = self.parent().width() / max(1, self.count()) - 1
+        width = self.parent().width() // max(1, self.count()) - 1
```
davvid, line 39: Thanks! I'll squash in a fixup to make this use `max(2, ...)` to remove any possibility of a divide-by-zero. That's pre-existing, though.

Issue #1176, opened 2021-10-17
double-click doesn't honor lock layout


sthalik:
When locked layout is enabled, it's still possible to ruin the layout through double-clicking. To reproduce it, double-click onto the bolded text on top of each widget (i.e. `Commit`, `Status`, `Diff`, `Console`, etc.). The expectation is for the click to be a no-op in case of a locked layout.

While here, I'd like to thank you again, @davvid for your immensely useful tool. My VCS skill would remain at `git commit -ambuffer\ flush` level if not for you! 🍺

PR #1198, opened 2022-01-30
Double click no longer creates sub-windows in locked layout

WNguyen14:
References issue #1176 , where the issue creator would like to see double clicking the widget title bars not pop open into sub-windows when they are in locked layout mode. This makes widgets no longer able to be popped out in locked layout mode. I'm not sure if this is intended behavior, but since the issue hadn't been closed yet, here's an attempt at fixing it.


davvid:
Thanks for this update, this looks worth trying! Please amend the commit to remove the commented-out code and I'll merge this in.

PR #1202, opened 2022-02-06
No longer pass a parent to Channel in Task.__init__()

living180:
Currently, when a `Channel` for a `Task` is created, the `Channel` is given a parent.  If for some reason that parent gets deleted from the main thread while the task is running, the `Channel` will also be deleted at the C++ level, though it will still exist at the Python level.  Then once the task has completed and `Task.run()` tries to emit the .result signal on the `Channel,` a `RuntimeError` with the message `"wrapped C/C++ object of type Channel has been deleted"` will be raised.

Although this is unlikely to happen in practice it still represents a potential race condition that is better avoided.  This commit addresses the issue by no longer passing a parent when creating the `Channel` in `Task.__init__()`, instead relying on Python's garbage collection to clean up the `Channel` when it is no longer needed.

### DIFF
--- a/cola/app.py
+++ b/cola/app.py
@@ -514,7 +514,7 @@ def async_update(context):
 
     """
     update_status = partial(context.model.update_status, update_index=True)
-    task = qtutils.SimpleTask(context.view, update_status)
+    task = qtutils.SimpleTask(update_status)
     context.runtask.start(task)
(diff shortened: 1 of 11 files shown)


davvid:
Nicely done, thanks again!

PR #1206, opened 2022-02-08
Replace Observable notifications with Qt Signals

### LINKED ISSUE #1191: highly reproducible crash on Windows
To make git-cola crash, press `C-q` immediately upon the program's window appearing. The current working directory must be a repo. pynotify is installed.

living180:
Notifications from `Observable` subclasses do not play well with lifetimes of Qt objects.  There are cases where underlying C++ object can be deleted while the Python object stays alive.  If a Qt object in that state is an observer for an `Observable` notification, a `RuntimeError` will generally result when the notification is issued (see #1191 for more details of a specific situation where this occurs).

Qt signals however do not have this problem as Qt is able to ensure that signals do not get delivered to Qt objects that have been deleted.  Thus this PR converts all usages of `Observable` notifications into Qt signals and then removes the `Observable` class.  In many cases, the observers were simply converting the notification into a Qt signal.  Thus this PR removes most of those intermediate Qt signals in favor of the new signals which replace the notifications.

Fixes #1191.


davvid:
Top notch, thanks for making this happen. I'll start reviewing now.

living180:
Tweaked the commit "dag: convert COMMITS_SELECTED notification to signal" to use `select_commits()` as the new name for the `commits_selected()` methods instead of `set_commits()`, as the latter name is misleading.

davvid:
Very nicely done. So far 0c0a2552840fe73e4df934509ae2f7ed7c81ce61 is the only fallout, and that was a pretty easy thing to miss (and fix).

living180:
Thanks for catching that.  I tried to to test things after my changes, but obviously some of the less core functionality didn't get exercised.  Hopefully that was the only issue!

PR #1207, opened 2022-02-12
windows: do not exclude vcruntime140.dll

### LINKED ISSUE #1094: Windows installer contains unused DLLs taking up space
I just installed Git Cola 3.7 64-bit on Windows 10. The folder in Program Files is 176.2 MB, which is very large, and many of the files are unnecessary.

- `C:\Program Files\git-cola\pkgs\PyQt5\Qt\bin\opengl32sw.dll` is 20 megabytes, and the app runs fine with OpenGL software rasterization removed. Same with `d3dcompiler_47.dll` and `libGLESv2.dll`.
- `Qt5Quick.dll` and `Qt5Qml.dll` are fairly small, but probably unnecessary considering you don't use QML. I ended up removing a pile of dlls, both Qt5[Foo].dll and a few others.
- `pkgs\PyQt5\Qt\plugins` can be cleaned up. `sceneparsers` is bloated and unused, all DLLs in `platforms` can be removed except `qwindows.dll` (IDK what qminimal.dll is used for), and else except `imageformats`, `styles`, and `iconengines`.
- `pkgs\PyQt5\Qt\qml` is small in size, but has 1033 files, and extracting them one by one slows down installation on NTFS.
- Don't know if `pkgs\PyQt5\Qt\translations` is in use.
- Why are there both `pkgs\PyQt5\Qt\bin\libcrypto-1_1-x64.dll` and `Python\libcrypto-1_1.dll`?

I managed to cut my folder size to just under 100 MB without removing any files in use.

I'm not sure how you're packaging Qt. You use a makefile, qtpy (which I haven't touched), and don't use windeployqt (which I've seen used for C++ Qt programs).

### LINKED ISSUE #1196: git-cola does not launch on Windows without Microsoft Visual C++ 2015 Redistributable
I installed git-cola on a Windows 10 64-bit virtual machine with very few other programs installed.  When I tried to launch git-cola, I saw essentially the same stack trace as #787.  Digging into a bit, I found that PyQt5.QtCore could not be imported, and that the root cause was that vcruntime140.dll could not be found.  Once I downloaded the Microsoft Visual C++ 2015 Redistributable from https://www.microsoft.com/en-us/download/details.aspx?id=52685 and installed it, git-cola launched successfully.

I don't know if it is possible or desirable to include the necessary Microsoft Visual C++ 2015 Redistributable DLLs in the git-cola Windows installer, but if not, this should probably be mentioned in the README so that other users don't run into this same issue.

davvid:
Remove the Visual Studio C++ 2015 DLLs from the pynsist
exclusion list so that we can import Qt modules on Windows
machines that do not have these DLLs installed.

Closes #1196
Closes #1094

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,18 @@ requirement, but not needed at runtime.
 
 # Windows (Continued)
 
+## Microsoft Visual C++ 2015 Redistributable
+
+Earlier versions of Git Cola may have shipped without `vcruntime140.dll`  and may
+not run on machines that are missing this DLL.
+
+To fix this, download the
+[Microsoft Visual C++ 2015 Redistributable](https://www.microsoft.com/en-us/download/details.aspx?id=52685)
+and install it
+
+Git Cola v4.0.0 and newer include this DLL and do not require this to be installed
+separately.
+
 ## Development
 
 In order to develop Git Cola on Windows you will need to install
--- a/pynsist.cfg
+++ b/pynsist.cfg
@@ -86,9 +86,7 @@ exclude = pkgs/PyQt5/Qt/bin/Qt5Bluetooth.dll
     pkgs/PyQt5/Qt/bin/concrt140.dll
     pkgs/PyQt5/Qt/bin/d3dcompiler_47.dll
     pkgs/PyQt5/Qt/bin/libeay32.dll
-    pkgs/PyQt5/Qt/bin/msvcp140.dll
     pkgs/PyQt5/Qt/bin/opengl32sw.dll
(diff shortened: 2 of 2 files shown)


davvid:
I also restored msvcp140 since that's some service pack DLL and seems somewhat related, so I'd rather be safe than sorry. If this looks good to you @living180  I'll merge this in.

living180:
Yes, the Qt DLLs depend on both `msvcp140.dll` and `vcruntime140.dll`, so this change looks good.

PR #1208, opened 2022-02-16
Diff line numbers work

living180:
Some cleanups, optimizations, and one bugfix to the diff line numbers code.


davvid:
Thanks again!

PR #1212, opened 2022-03-07
github/workflows: use a virtualenv to get a newer setuptools

davvid:
The setuptools in Ubuntu 20.04 is too old. Activate a newer
one using a Python virtualenv.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
+          set -x
           sudo apt-get update
           sudo apt-get install \
               black \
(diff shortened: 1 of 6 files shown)


davvid:
We have custom distutils.commands entry points for stuff like `python setup.py build_mo` and other stuff. We also do stuff like hook into `python setup.py build` in an underhanded manner. This is relying on deprecated features (Cf. `SETUPTOOLS_USE_DISTUTILS=stdlib`) so work probably remains down the line to come up with a better solution.

The core problem I'm trying to solve is that the `share/locale` files are generated files that need to be included in the distribution. Yup.. we're also relying on `data_files` because we want to install our translations into the standard `$prefix/share/local/*/LC_MESSAGES` area.

I do like that we can `pip install git-cola` now, but this seems like a pretty shaky foundation.

PR #1213, opened 2022-03-09
Update ja.po

kenyuy:
Fix a typo.


fu7mu4:
Yes, It is typo in Japanese. Thanks a lot.

davvid:
Thanks both! much appreciated.

PR #1222, opened 2022-04-15
Custom Themes

brccabral:
User can create a `.qss` file in `%USERPROFILE%\.config\git-cola\themes\` or `$HOME/.config/git-cola/themes/`. `.qss` would contain Qt Style Sheet format to set colors for Qt widgets. The file name becomes the style name.

### DIFF
--- a/cola/themes.py
+++ b/cola/themes.py
@@ -31,7 +31,6 @@ class Theme(object):
         if self.style_sheet == EStylesheet.CUSTOM:
             return self.style_sheet_custom(app_palette)
         elif self.style_sheet == EStylesheet.FLAT:
-            print(self.style_sheet_flat())
             return self.style_sheet_flat()
         else:
             return self.style_sheet_default(app_palette)
--- a/cola/themes.py
+++ b/cola/themes.py
@@ -16,7 +16,6 @@ class EStylesheet(object):
     DEFAULT = 1
     FLAT = 2
     CUSTOM = 3  # Files located in ~/.config/git-cola/themes/*.qss
-    SYSTEM = 4  # Application qt5ct (Qt5 Configuration Tool)
 
 
 class Theme(object):
@@ -577,7 +576,6 @@ def get_all_themes():
         Theme(
             'flat-dark-green', N_('Flat dark green'), True, EStylesheet.FLAT, '#42a65c'
         ),
-        Theme('system', N_('System'), False, EStylesheet.SYSTEM, None),
     ]
 
     # check if themes path exists in user folder
--- a/cola/themes.py
+++ b/cola/themes.py
@@ -583,9 +583,7 @@ def get_all_themes():
         for file in files:
             name, ext = os.path.splitext(file)
             if ext == '.qss':
-                themes.append(
-                    Theme(name, N_(name.capitalize()), False, EStylesheet.CUSTOM, None)
-                )
+                themes.append(Theme(name, N_(name), False, EStylesheet.CUSTOM, None))
     return themes
 
 
--- a/cola/themes.py
+++ b/cola/themes.py
@@ -570,9 +570,7 @@ def get_all_themes():
         Theme(
             'flat-dark-green', N_('Flat dark green'), True, EStylesheet.FLAT, '#42a65c'
         ),
-        Theme(
-            'system', N_('System'), None, EStylesheet.SYSTEM, None
-        ),
+        Theme('system', N_('System'), None, EStylesheet.SYSTEM, None),
     ]
 
     # check if themes path exists in user folder
@@ -586,7 +584,7 @@ def get_all_themes():
             name, ext = os.path.splitext(file)
             if ext == '.qss':
                 themes.append(
-                    Theme(name, N_(name.capitalize()), None, EStylesheet.CUSTOM, None)
+                    Theme(name, N_(name.capitalize()), False, EStylesheet.CUSTOM, None)
                 )
     return themes
 
--- a/cola/themes.py
+++ b/cola/themes.py
@@ -1,5 +1,5 @@
-"""Provides themes generator"""
-from __future__ import absolute_import, division, unicode_literals
+"""Themes generators"""
+from __future__ import absolute_import, division, print_function, unicode_literals
 import os
 
 from qtpy import QtGui
@@ -17,9 +17,9 @@ class EStylesheet(object):
 
 
 class Theme(object):
-
-    def __init__(self, name, hr_name, is_dark,
-                 style_sheet=EStylesheet.DEFAULT, main_color=None):
+    def __init__(
+        self, name, hr_name, is_dark, style_sheet=EStylesheet.DEFAULT, main_color=None
+    ):
         self.name = name
         self.hr_name = hr_name
         self.is_dark = is_dark
@@ -43,11 +43,11 @@ class Theme(object):
         if not palette_dark and not self.is_dark:
             return app_palette
         if self.is_dark:
-            bg_color = QtGui.QColor('#202025')
+            bg_color = QtGui.QColor("#202025")
         else:
-            bg_color = QtGui.QColor('#edeef3')
+            bg_color = QtGui.QColor("#edeef3")
 
-        txt_color = QtGui.QColor('#777')
+        txt_color = QtGui.QColor("#777")
         palette = QPalette(bg_color)
         palette.setColor(QPalette.Base, bg_color)
         palette.setColor(QPalette.Disabled, QPalette.Text, txt_color)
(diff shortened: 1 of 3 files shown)


cola/themes.py
```diff
@@ -12,6 +13,7 @@
 class EStylesheet(object):
     DEFAULT = 1
     FLAT = 2
+    CUSTOM = 3
```
davvid, line 16: Nicely done. Yup, this is pretty much exactly what I had in mind on how to implement this feature.

cola/themes.py
```diff
@@ -521,9 +525,29 @@ def style_sheet_flat(self):
...
+
+        # get user path Windows
+        userpath = os.getenv('USERPROFILE')
+        # if not found Windows, try Unix
+        if not userpath:
+            userpath = os.getenv('HOME')
+        # check if path exists
+        filepath = os.path.join(
+            userpath, '.config', 'git-cola', 'themes', self.name + '.qss'
+        )
```
davvid, line 541: There's some duplicate code here. I think `cola/resources.py` is where a helper function `def custom_theme_path(name)` could go.

I *think* this section can be replaced with:

```
filepath = resources.config_home('themes', self.name + '.qss')
```

That way it respects the XDG .config variables, etc. We don't really handle USERPROFILE in there, though, but I want to believe that the `expanduser('~')` in Python is doing the right thing and handling that for us.

If it's not, then that would be a good place to centralize that behavior and have an `if utils.is_win32(): ...` special-case where we check for USERPROFILE.

davvid (changes requested):
Thanks, this looks great! Please amend the commit with these fixups and this should be good to merge.

cola/themes.py
```diff
@@ -43,11 +43,11 @@ def build_palette(self, app_palette):
         if not palette_dark and not self.is_dark:
             return app_palette
         if self.is_dark:
-            bg_color = QtGui.QColor("#202025")
+            bg_color = QtGui.QColor('#202025')
```
davvid, line 46: Thank you for using single quotes! Squash this into the same commit too.

cola/themes.py
brccabral: it is resources.config_home, I didn't know about this, learned something today!

cola/themes.py
```diff
@@ -30,11 +30,12 @@ def build_style_sheet(self, app_palette):
         if self.style_sheet == EStylesheet.CUSTOM:
             return self.style_sheet_custom(app_palette)
         elif self.style_sheet == EStylesheet.FLAT:
+            print(self.style_sheet_flat())
             return self.style_sheet_flat()
         else:
             return self.style_sheet_default(app_palette)
 
-    def build_palette(self, app_palette):
+    def build_palette(self, app_palette) -> QtGui.QPalette:
```
davvid, line 38: Heads-up that I'm going to remove the type annotation here. It's ok in docstrings, but sometimes I do still run this under python2 (I know).

cola/themes.py
```diff
@@ -553,6 +571,21 @@ def get_all_themes():
...
 
+    # check if themes path exists in user folder
+    path = resources.config_home('themes')
+    if not os.path.exists(path):
+        return themes
+
+    # get only files with extension .qss
+    for root, dirs, files in os.walk(path):
+        for file in files:
+            name, ext = file.split('.')
```
davvid, line 582:
```
name, ext = os.path.splitext(file)
```
is more appropriate because it handles filenames like `foo.bar.qss`. This code will blow up if it sees that filename. Another alternative would be `name, ext = file.rsplit('.', 1)` but splitext is a little clearer

splitext is more robust because it also won't trip up over files that don't contain `.` and will always return 2 elements. That edge case is demonstrated by `name, ext = 'file-with-no-extension'.split('.')` because the resulting tuple has only 1 element but the left-hand `name, ext =` wants 2.

Also (this one is annoying) I'd suggest avoiding the name `file` for variables. That one is so annoying because I totally want to use the name `file` all the time. The suggested alternative is `filename`.

That's actually a python2-ism.. `file` is a builtin symbol in python2. In python3 it's not, so I guess it's okay to start using that name now. Well.. okay that's a good python3 change I guess.

Anyways.. since we need to tweak this for the `splitext` thing then we might as well rename the variable to `filename` while we're here so that it doesn't get caught by any of the linters.

Do note that `splitext` will return `.qss` so we'd have to adjust the conditional below in tandem.

davvid (changes requested):
Last sugs, I promise! Thanks again for making this happen and for verifying the config_home behavior on Windows.

cola/themes.py
davvid: I later noticed the other review comments below so I'll let you remove the type annotations when updating this file for those notes.

brccabral: done

cola/themes.py
brccabral: done, it is better indeed

brccabral:
If you don't mind, I added a global system configuration. When creating my theme, managing all .qss options was getting complicated, adding a global system was much simpler.

cola/themes.py
```diff
@@ -551,8 +570,26 @@ def get_all_themes():
...
+    if not os.path.exists(path):
+        return themes
+
+    # get only files with extension .qss
+    for root, dirs, files in os.walk(path):
+        for file in files:
+            name, ext = os.path.splitext(file)
+            if ext == '.qss':
+                themes.append(
+                    Theme(name, N_(name.capitalize()), None, EStylesheet.CUSTOM, None)
```
davvid, line 589: Is it okay if we didn't do the `name.capitalize()` thing?

cola/themes.py
```diff
@@ -1,17 +1,21 @@
...
 from . import icons
 from . import qtutils
+from . import resources
 
 
 class EStylesheet(object):
     DEFAULT = 1
     FLAT = 2
+    CUSTOM = 3  # Files located in ~/.config/git-cola/themes/*.qss
+    SYSTEM = 4  # Application qt5ct (Qt5 Configuration Tool)
```
davvid, line 18: I always considered `DEFAULT` the same as what `SYSTEM` does...

cola/themes.py
```diff
@@ -521,9 +527,22 @@ def style_sheet_flat(self):
...
+        File name is saved in variable self.name.
+        If user has deleted file, use default style"""
+
+        # check if path exists
+        filepath = resources.config_home('themes', self.name + '.qss')
+        if not os.path.exists(filepath):
+            return self.style_sheet_default(app_palette)
+
+        with open(filepath) as file:
+            return file.read()
```
davvid, line 541: Maybe we should catch the exception here?

```
try:
    with open(filepath) as fp:
        return fp.read()
except (IOError, OSError) as err:
    sys.stderr.write('warning: unable to read custom theme %s: %s\n' % (filepath, err))
    return self.style_sheet_default(app_palette)
```

That way we'll get an error message if we're unable to read the custom config for some reason.

cola/themes.py
```diff
@@ -551,8 +570,26 @@ def get_all_themes():
         Theme(
             'flat-dark-green', N_('Flat dark green'), True, EStylesheet.FLAT, '#42a65c'
         ),
+        Theme(
+            'system', N_('System'), None, EStylesheet.SYSTEM, None
+        ),
```
davvid, line 575: Note: `'system'` looks like it's identical to `'default'` -- what's with the `False` vs. `None` difference in the 3rd parameter?

Do we need `'system'` in light of `'default'`?

cola/themes.py
davvid: The 3rd parameter should be `False` not `None`. `is_dark` is a bool.

davvid (changes requested):
Let me know what you think. Thanks for the updates so far.

cola/themes.py
brccabral: removed capitalize, changed to False

cola/themes.py
brccabral: Strange, I just updated Ubuntu from 20.04 to 22.04. In 20.04 I couldn't get Default to use system config. But now that I have updated to 22.04 and I don't see any differences. I removed from this patch

cola/themes.py
brccabral: added exception

cola/themes.py
brccabral: removed system. As explained above, it was a problem in Ubuntu 20.04 but not 22.04

davvid:
In `share/doc/git-cola/git-cola.rst` there's a section about "dark mode and window manager themes" where it would make sense to document this new feature.

That probably can be its own commit. Before the doc updates commit, can you please rebase and squash all of these 13 commits into a single commit?

Thanks for making all these updates. This is a cool feature but most users won't know about it unless we advertise it in the docs.

brccabral:
I was able to add a custom theme section to the doc .rst and rebase on my local pc, but I couldn't push the same branch to GitHub. I created another branch just to push to GitHub, I am going to create a new pull request.

PR #1225, opened 2022-04-30
Add additional AppStream metadata

Mailaender:
This is required for https://github.com/flathub/flathub/pull/3072

### DIFF
--- a/share/metainfo/git-cola.appdata.xml
+++ b/share/metainfo/git-cola.appdata.xml
@@ -16,4 +16,7 @@
     </screenshot>
   </screenshots>
   <url type="homepage">https://git-cola.github.io/</url>
+  <releases>
+    <release version="3.12.0" date="2022-01-20" />
+  </releases>
 </component>
--- a/share/metainfo/git-cola.appdata.xml
+++ b/share/metainfo/git-cola.appdata.xml
@@ -19,4 +19,5 @@
   <releases>
     <release version="3.12.0" date="2022-01-20" />
   </releases>
+  <content_rating type="oars-1.1" />
 </component>


share/metainfo/git-cola.appdata.xml
```diff
@@ -16,4 +16,8 @@
     </screenshot>
   </screenshots>
   <url type="homepage">https://git-cola.github.io/</url>
+  <releases>
+    <release version="3.12.0" date="2022-01-20" />
+  </releases>
+  <content_rating type="oars-1.1" />
```
davvid, line 22: Thanks! This `<releases>` bit here is an extra step of maintenance that I'm going to be very prone to forget.

note to self: add a note about this to the `checklist` file in the `todo` branch.

Even better ~ update the `set-version` script in the `todo` branch to automatically append the new version here.

That script is run whenever we update the version ~ it goes and modifies the hard-coded version number that ends up needing to be written down in various files. I'll extended the script to handle this file as well before merging this PR and tagging the next release.

Thanks!

davvid:
Quick question if you don't mind ~ we have both a git-cola.appdata.xml file and a git-dag.appdata.xml file, one for each standalone tool. Do we need to include this metadata in both files?

davvid:
I updated the script I run when tagging new versions in 05ceb36c9efa4c4c02313cf4546310496048b511 so we're ready to merge this update. Let me know what you think about the other xml file, though.

If anything, I imagine that we'd just copy/paste the same `<releases>` tag in there. I already have the script setup to handle that tag in both files in case the answer is "yes, we should maintain the list of releases in both xml files".

Mailaender:
I am quite new to @flathub, so I don't know if it supports two applications in one package.

davvid:
Thanks for the response. I don't know much about flathub either so I'm just going to assume that adding the metadata won't be a concern. I've added this to git-dag.appdata.xml as well. Thanks again!

PR #1226, opened 2022-05-03
Support for Custom Themes

brccabral:
User can create a `.qss` file in `~/.config/git-cola/themes/`. `.qss`
would contain Qt Style Sheet format to set colors for Qt widgets. The
file name becomes the style name.


davvid:
Nicely done! Thanks again for adding this feature!

PR #1235, opened 2022-05-28
Error thrown when changing boolean preference

TypeDefinition:
This error is thrown when changing a boolean preference, such as "Insert spaces instead of tabs".

image


davvid:
Thanks for the heads-up. It looks like the root cause is the `Signal(str, str, str)` signature that was introduced recently.

We need to preserve bools, so calling `str()` on them will break things. We can change the signal signature to `Signal(str, str, object)` so that we can pass bool, str and None through the interface.

PR #1240, opened 2022-06-27
enable window icon on wayland (on supported desktop environments)

bearsh:
kde for example still shows a window icon but on wayland, this only
works if the icon can be retrieved from the .desktop file. the name of
the .dektop file needs to be set from the application.

Signed-off-by: Martin Gysel <me@bearsh.org>


davvid:
Thanks, this looks great! I'm going to merge this in shortly.

Please note that we prefer `'single quotes'` over `"double"` quotes since they're easier to type. This file uses single quotes exclusively so I'll swap this over post-merge.

PR #1244, opened 2022-07-01
updated spanish translation

victorhck:


cola/i18n/es.po
```diff
@@ -2769,7 +2854,7 @@ msgid "git-cola diff"
 msgstr "git-cola diff"
 
 msgid "git://git.example.com/repo.git"
-msgstr ""
+msgstr "git://git.ejemplo.com/repo.git"
```
davvid, line 2857: Gracias! I'll probably switch this one back to `example.com` to avoid linking to a random domain that might one day contain who-knows-what on it.

The reason we use example.com here is because it's one of the special reserved domain names for exactly these purposes.

https://www.rfc-editor.org/rfc/rfc6761.html

https://www.example.com/

PR #1245, opened 2022-07-04
Add Open Worktree command to context menu of Status Widget

laerreal:
Note, I don't really realize why analogue commands (`default_app_action`, `parent_dir_action`) are forbidden for `win32`.
I forbid the new command too because of that.``

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -1683,24 +1683,37 @@ class OpenDefaultApp(ContextCommand):
         core.fork([self.launcher] + self.filenames)
 
 
-class OpenParentDir(OpenDefaultApp):
+class OpenDir(OpenDefaultApp):
+    """Open directories using the OS default."""
+
+    @staticmethod
+    def name():
+        return N_('Open Directory')
+
+    @property
+    def _dirnames(self):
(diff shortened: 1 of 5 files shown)


davvid:
Thanks! The only reason we don't allow it on Windows is because I don't know the equivalent to `open` (macOS) or `xdg-open` (Linux) that we can use to open the default app.

TODO: I should probably switch `xdg-open` over to the more modern `gio open`.

Thanks for these updates.

davvid:
Nevermind, `xdg-open` is plenty fine.

It looks like on Windows there's an `os.startfile(...)` that can be used to open the default application.

I'm not able to test it, but the code could possibly be changed to detect win32 and use `os.startfile(...)` to provide a Windows solution.

The cmds interface allows passing multiple paths to it, though, and the Python API only takes a single file. It might require calling that function in a loop. Just noting in case there's a Windows user that might want to implement this.

davvid:
I implemented Windows support after doing some testing in a VM. I enabled all of these actions on Windows now too (including the new "Open Worktree" action).

Thanks again for these updates, good stuff!

PR #1249, opened 2022-07-30
i18n: Fix a small part of the Japanese translation

kisaragi-hiu:
- Fix the display for "&Edit" and "&File", which are supposed to look like "編集(&E)" (displayed as "編集(E)", with the underline appearing   under E with Alt held).
- Fix a bizarre mistranslation of "<do stuff>..." to "<do stuff>中"
- Fix the Japanese names of Simplified Chinese and Traditional Chinese
- Use "未追跡" for "Untracked"


davvid:
Thank you, much appreciated!

PR #1256, opened 2022-08-30
Fix documentation rendering errors

kurtmckee:
While reading the docs I noticed a few issues with the rendering. In particular:

* The `cola.terminalshellquote` defaults table renders not as a table, but as a long line of text.
* The CLI options are not quoted with double backticks and were subject to conversion to em-dashes.

While fixing these I also noticed a few typos and fixed those as well.

Please let me know if these rendering issues need to be fixed in another way.

### DIFF
--- a/docs/git-cola.rst
+++ b/docs/git-cola.rst
@@ -13,32 +13,32 @@ Git Cola is a sleek and powerful Git GUI.
 OPTIONS
 =======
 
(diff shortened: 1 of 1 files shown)


davvid:
thanks!

kurtmckee:
You're welcome!

PR #1257, opened 2022-09-18
sequenceeditor: select, change, and shift multiple commits

milestone12:
# Issue
In the commit list view for the interactive rebase you can only select and modify a single entry. It becomes annoying really quickly needing to go one commit at a time when you, say, prepare a series of commits for a pull request and you want to squash multiple sets of commits or need to do final edits on multiple commit messages.

# Solution
* use extended selection mode for RebaseTreeWidget, so that multiple commits can be selected holding Ctrl or Shift key
* add context action "Enable Selected" to enable selected commits
* add context action "Disable Selected" to disable selected commits

### DIFF
--- a/cola/sequenceeditor.py
+++ b/cola/sequenceeditor.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import sys
 import re
 from argparse import ArgumentParser
-from functools import partial
+from functools import partial, reduce
 
 from cola import app  # prints a message if Qt cannot be found
 from qtpy import QtCore
@@ -320,6 +320,7 @@ class RebaseTreeWidget(standard.DraggableTreeWidget):
         )
         self.header().setStretchLastSection(True)
         self.setColumnCount(5)
+        self.setSelectionMode(QtWidgets.QAbstractItemView.ExtendedSelection)
 
         # actions
         self.copy_oid_action = qtutils.add_action(
@@ -446,10 +447,10 @@ class RebaseTreeWidget(standard.DraggableTreeWidget):
         self.commits_selected.emit(commits)
 
     def toggle_enabled(self):
-        item = self.selected_item()
-        if item is None:
-            return
-        item.toggle_enabled()
+        items = self.selected_items()
+        logic_or = reduce(lambda res, item: res or item.is_enabled(), items, False)
+        for item in items:
+            item.set_enabled(not logic_or)
 
     def select_first(self):
         items = self.items()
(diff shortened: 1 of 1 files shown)


davvid:
This is cool. We have hotkeys for changing the modes (press `?` to see them) so we should also make these hotkeys respect the multi-selection.

Right now we can hit `j` and `{1,2,3,4}` quickly in succession to change the modes, which helps avoid the need to use the mouse, but being able to hold shift while selecting and then just one of `{1,2,3,4,..}` to change the mode would make it pretty slick.

Right now we have a `j` and `J` hotkeys, and `J` shifts the row down. The downside is that means that we can't use `shift+j` to select the next row with multi-select.

Maybe we can let `alt-j` and `alt-k` shift rows up and down instead of `shift-j/k` so that the UI is more amenable to multi-select.

This seems okay to merge. Let me know what you think about the hotkeys and I can look into extending them for multi-select.

davvid:
It looks like `spacebar` is already the hotkey for toggling the enabled/disabled state, and we already have a context menu entry `Toggle Enabled` that is responsible for this action.

Sorry about that -- I take that back -- we shouldn't merge this as-is since we shouldn't need new translations and new context menu entries for it. It seems like we should be able to extend the existing action instead.

milestone12:
Thank you for your feedback.

I absolutely agree with you that the hotkeys should apply in the same way to multi-selection as they do for single-selection. I will look into this. The mode changes in particular should be straight forward as they have set/override semantics.

The open topic of combining multi-selection with the existing _Toggle Enabled_ context action and `space` hotkey begs the question how the toggle action should behave in the case that the initial _Enabled_ state of the selected commits is mixed, i.e. 2 enabled and 2 disabled. I'd propose to disable any enabled commits initially.

davvid:
That's a good point about what to do about a mixed selection. An alternative would be to toggle each one, but I think your proposal sounds simpler from a usage perspective, so 👍 to disabling them all. If they're all disabled then we can enable them all.

milestone12:
> Maybe we can let `alt-j` and `alt-k` shift rows up and down instead of `shift-j/k` so that the UI is more amenable to multi-select.

If not absolutely necessary, I would not change existing behavior of the UI, i.e. remapping `Shift-j/k` to a different action.
With the latest commit it's technically not required to take your hands off the keyboard to do any of the multi-selection work, but one of the keyboard-based commit selection actions might be a little bit funky:
* `space` key works to toggle enabled state of multiple commits
* keys `1, 2, 3, 4, 5` work to change the mode of multiple commits
* `Shift-j/k` move multiple commits up and down, even if the selected commits are not arranged in a contiguous block
* `Shift-Up/Down` allows to select a contiguous block of commits
* Now for the funky one: `Ctrl-Up/Down` moves a kind of pre-selection cursor up/down the list. By hitting `Ctrl-space` you can (de)select non-contiguous commits. I guess this last one is Qt's out-of-the-box behavior.

davvid:
Very nIcely done! This is great, thanks again for adding these really useful features.

PR #1261, opened 2022-10-04
Add support for git rebase with `--update-refs`

virtualstaticvoid:
Adds support to the sequence editor for updating dependent branches when rebasing with `–update-refs`. 

screenshot

This feature was introduced in Git v2.38, see [Rebase dependent branches with –update-refs](https://github.blog/2022-10-03-highlights-from-git-2-38/#rebase-dependent-branches-with-update-refs) for more information on the feature.


davvid:
Nicely done, thanks for adding this!

Another recent rebase features is `git rebase --rebase-merges` which uses `merge` and `label` tokens. The sequence editor doesn't currently support those.

That'll be good to look into adding after this gets merged.

I'll be merging this PR soon. `v4.0.3` is only going to contain fixes, and there's a pending fix in #1267 that i'm going to merge first. After that's merged I'll tag `v4.0.3` and then I'll merge this change in immediately after.

virtualstaticvoid:
Hi @davvid 

I've managed to implement support for the `--rebase-merges` option, which introduces the `label`, `reset` and `merge` commands as can be seen the screenshot:

Screenshot

However, I don't think it makes sense to manage the interactive rebase in this way, as for example, the `label`, `reset` and `merge` commands need to work together, so reordering or removing/disabling them will cause the rebase to fail in peculiar ways.

I think more design work is required to get this right, so you can go ahead and merge this PR.

I'll create another to deal with the `--rebase-merges` functionalitty, and perhaps we can discuss the design.

davvid:
Thanks again! v4.0.3 was just tagged. This topic will be in the next feature release, v4.1.0.

davvid:
And, indeed, a bigger redesign will be needed in order to handle the preserve-merges feature.

Issue #289, opened 2014-06-08
"Bookmarks" feature needs to be re-designed


Vdragon:
#### Original issue: recently used repository are not listed in "Bookmarks" dialog

![20140609_004](https://cloud.githubusercontent.com/assets/1192163/3212097/5f088b1c-ef44-11e3-9f63-4be8aa561033.jpg)
Currently only user added bookmark entry listed in the "Bookmarks" dialog, recently used repository are not listed
### Expected behavior

I personally believe the Bookmarks feature needs a re-design
#### Don't include recently-used repositories
- We already have a sub-menu dedicated for this purpose.
  ![20140609_005](https://cloud.githubusercontent.com/assets/1192163/3212125/829c5bec-ef46-11e3-8aa1-ad54b90e01fc.jpg)
#### Get rid of the "Bookmarks" dialog

![20140609_006](https://cloud.githubusercontent.com/assets/1192163/3212133/dc14289e-ef46-11e3-9c4c-9156fbf8b343.jpg)
- Not much exclusive function, duplicate "Open" button, meaningless "Save" button, etc.
- Everything can be integrated to the "Bookmarks" panel(Add/Remove/Edit path...) as qtutils.create_action_button as far as I can see
#### Rename "Bookmarks" itself
- "Bookmarks" is not guess-friendly(except to Mozilla Firefox users?)
- "Favorite repo" or other descriptive names would be better
### Proposed qtutils.create_action_button of "Bookmarks" panel
- <del>Open selected bookmarked repository</del>(already implemented)
- Add current opened repository to Bookmarks
- Remove selected repository from Bookmarks
- Edit repository path(in case of moved to another path, etc.)
- Move selected repository up in the list
- Move selected repository down in the list
- (optional)Add repository from user specified directory recursively
#### Operating system

Ubuntu 14.04LTS
#### git-cola

commit 852ba089

PR #1280, opened 2023-01-01
Add context menu in prompt dialog (git cola --prompt)

### LINKED ISSUE #1199: Cannot remove old folders from git-cola main window
When opening git-cola, you are presented with a selection of folders you have previously opened in git-cola.  There does not appear to be a way to remove these from within the GUI.

nakanoi:
According to this issue, "[Cannot remove old folders from git-cola main window #1199](https://github.com/git-cola/git-cola/issues/1199)", I implemented context menu in prompt dialog (`git cola --prompt`).

The content of the context menu is same as favorite(and recent) widget (in `cola/widgets/bookmarks.py`) in main UI.
A lot of context menu functions come from the code.

- Open
- Open in New WIndow
- Open Using Default Application
- Copy
- Launch Editor
- Launch Terminal
- Clear Deafult Repository
- Rename Repository
  - **only activated in "Folder" tab** [1]
- Delete from Bookmakrs List
  - **only in `Favorites` view**
- Delete from Recent List
  - **only in `Recent` view**
- Prune Missing Entries

prompt-capture

Fixes #1199 
**Thank you!**

### DIFF
--- a/cola/widgets/startup.py
+++ b/cola/widgets/startup.py
@@ -72,26 +72,21 @@ class StartupDialog(standard.Dialog):
         recent = settings.recent
         all_repos = bookmarks + recent
 
-        directory_icon = icons.directory()
-        user_role = Qt.UserRole
         normalize = display.normalize_path
         paths = {normalize(repo['path']) for repo in all_repos}
         short_paths = display.shorten_paths(paths)
         self.short_paths = short_paths
 
         added = set()
+        builder = BuildItem(self.context)
         for repo in all_repos:
             path = normalize(repo['path'])
+            name = normalize(repo['name'])
             if path in added:
                 continue
             added.add(path)
 
-            item = QtGui.QStandardItem(path)
-            item.setEditable(False)
-            item.setData(path, user_role)
-            item.setIcon(directory_icon)
-            item.setToolTip(path)
-            item.setText(self.short_paths.get(path, path))
+            item = builder.get(path, name)
             bookmarks_model.appendRow(item)
             items.append(item)
 
@@ -274,6 +269,38 @@ class StartupDialog(standard.Dialog):
         if selected and selected[0].row() != 0:
             return self.bookmarks_model.data(selected[0], Qt.UserRole)
         return None
+
+
+class BuildItem(object):
+    def __init__(self, context):
+        self.star_icon = icons.star()
+        self.folder_icon = icons.folder()
+        cfg = context.cfg
+        self.default_repo = cfg.get('cola.defaultrepo')
+
+    def get(self, path, name):
+        is_default = self.default_repo == path
+        if is_default:
+            icon = self.star_icon
+        else:
+            icon = self.folder_icon
+        return PromptWidgetItem(path, name, icon, is_default)
+
+
+class PromptWidgetItem(QtGui.QStandardItem):
+    def __init__(self, path, name, icon, is_default):
+        QtGui.QStandardItem.__init__(self, icon, name)
+        self.path = path
+        self.is_default = is_default
+        self.name = name
+
+        user_role = Qt.UserRole
+        self.setEditable(False)
+        self.setData(path, user_role)
+        self.setIcon(icon)
+        self.setText(name)
+        self.setToolTip(path)
+        self.setFlags(self.flags() | Qt.ItemIsEditable)
 
 
 def make_size(size):
--- a/cola/widgets/startup.py
+++ b/cola/widgets/startup.py
@@ -7,16 +7,22 @@ from qtpy import QtGui
 from qtpy import QtWidgets
 
 from ..i18n import N_
+from .. import cmds
 from .. import core
 from .. import display
 from .. import guicmds
 from .. import hotkeys
 from .. import icons
 from .. import qtutils
+from .. import utils
 from .. import version
 from . import clone
 from . import defs
 from . import standard
+
+
+ICON_MODE = 0
+LIST_MODE = 1
 
 
 class StartupDialog(standard.Dialog):
@@ -73,33 +79,28 @@ class StartupDialog(standard.Dialog):
         all_repos = bookmarks + recent
 
         normalize = display.normalize_path
-        paths = {normalize(repo['path']) for repo in all_repos}
-        short_paths = display.shorten_paths(paths)
-        self.short_paths = short_paths
 
         added = set()
         builder = BuildItem(self.context)
-        for repo in all_repos:
+        default_view_mode = ICON_MODE
+        for idx, repo in enumerate(all_repos):
             path = normalize(repo['path'])
             name = normalize(repo['name'])
+            is_bookmark = idx < len(bookmarks)
             if path in added:
                 continue
             added.add(path)
 
-            item = builder.get(path, name)
+            item = builder.get(path, name, default_view_mode, is_bookmark)
             bookmarks_model.appendRow(item)
             items.append(item)
 
-        selection_mode = QtWidgets.QAbstractItemView.SingleSelection
-        self.bookmarks = bookmarks = QtWidgets.QListView()
-        bookmarks.setSelectionMode(selection_mode)
-        bookmarks.setModel(bookmarks_model)
-        bookmarks.setViewMode(QtWidgets.QListView.IconMode)
-        bookmarks.setResizeMode(QtWidgets.QListView.Adjust)
-        bookmarks.setGridSize(make_size(defs.large_icon))
-        bookmarks.setIconSize(make_size(defs.medium_icon))
-        bookmarks.setDragEnabled(False)
-        bookmarks.setWordWrap(True)
+        self.bookmarks = bookmarks = BookmarksListView(
+            context,
+            bookmarks_model,
+            self.open_selected_bookmark,
+            self.set_model,
+        )
 
         self.tab_layout = qtutils.vbox(
             defs.no_margin, defs.no_spacing, self.tab_bar, self.bookmarks
@@ -134,14 +135,9 @@ class StartupDialog(standard.Dialog):
         qtutils.connect_button(self.clone_button, self.clone_repo)
         qtutils.connect_button(self.new_button, self.new_repo)
         qtutils.connect_button(self.close_button, self.reject)
-        # Open the selected repository when "enter" is pressed.
-        self.action_open_repo = qtutils.add_action(
-            self, N_('Open'), self.open_selected_bookmark, *hotkeys.ACCEPT
-        )
 
         # pylint: disable=no-member
         self.tab_bar.currentChanged.connect(self.tab_changed)
-        self.bookmarks.activated.connect(self.open_bookmark)
 
         self.init_state(settings, self.resize_widget)
         self.setFocusProxy(self.bookmarks)
@@ -155,22 +151,36 @@ class StartupDialog(standard.Dialog):
 
     def tab_changed(self, idx):
         bookmarks = self.bookmarks
-        if idx == 0:
-            bookmarks.setViewMode(QtWidgets.QListView.IconMode)
-            bookmarks.setIconSize(make_size(defs.medium_icon))
-            bookmarks.setGridSize(make_size(defs.large_icon))
+        if idx == ICON_MODE:
+            mode = QtWidgets.QListView.IconMode
+            icon_size = make_size(defs.medium_icon)
+            grid_size = make_size(defs.large_icon)
             list_mode = 'folder'
-            for item in self.items:
-                path = item.data(Qt.UserRole)
-                item.setText(self.short_paths.get(path, path))
+            view_mode = ICON_MODE
+            rename_enabled = True
         else:
-            bookmarks.setViewMode(QtWidgets.QListView.ListMode)
-            bookmarks.setIconSize(make_size(defs.default_icon))
-            bookmarks.setGridSize(QtCore.QSize())
+            mode = QtWidgets.QListView.ListMode
+            icon_size = make_size(defs.default_icon)
+            grid_size = QtCore.QSize()
             list_mode = 'list'
-            for item in self.items:
-                path = item.data(Qt.UserRole)
-                item.setText(path)
+            view_mode = LIST_MODE
+            rename_enabled = False
+
+        self.bookmarks.set_rename_enabled(rename_enabled)
+        self.bookmarks.set_view_mode(view_mode)
+
+        bookmarks.setViewMode(mode)
+        bookmarks.setIconSize(icon_size)
+        bookmarks.setGridSize(grid_size)
+
+        new_items = []
+        builder = BuildItem(self.context)
+        for item in self.items:
+            if isinstance(item, PromptWidgetItem):
+                item = builder.get(item.path, item.name, view_mode, item.is_bookmark)
+                new_items.append(item)
+
+        self.set_model(new_items)
 
         if list_mode != self.list_mode:
             self.list_mode = list_mode
@@ -246,11 +256,10 @@ class StartupDialog(standard.Dialog):
                 return
             self.accept()
 
-    def handle_broken_repo(self, index):
+    def handle_broken_repo(self, index, mode):
         settings = self.context.settings
         all_repos = settings.bookmarks + settings.recent
         repodir = self.bookmarks_model.data(index, Qt.UserRole)
-
         repo = next(repo for repo in all_repos if repo['path'] == repodir)
         title = N_('Repository Not Found')
         text = N_('%s could not be opened. Remove from bookmarks?') % repo['path']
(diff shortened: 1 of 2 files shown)


cola/widgets/startup.py
```diff
@@ -265,16 +336,397 @@ def handle_broken_repo(self, index):
...
+                other_index = other_model.index(i, 0)
+                other_item = other_model.itemFromIndex(other_index)
+
+                if (
+                    isinstance(other_item, PromptWidgetItem)
+                    and other_item.path == repo['path']
+                ):
+                    other_items.remove(other_item)
+                    other_model.removeRow(other_index.row())
+                    return
```
davvid, line 354: This is pretty complex because it's trying to mutate the view to quickly remove this entry.

The original code was a little fast and loose in that it modified the models directly but it wasn't as complex. Would it be possible to eliminate this coupling?

This code also originally only dealt with the bookmarks list. It'd almost be better to not change that behavior for now so that it doesn't need to have this complicated "other model" coupling going on.

Update: after testing, I see a little bit of why this is being done -- the view is now split and it uses a different widget for the Recents and the Favorites.

I'll have to think about that.. in the past it was all presented as a single widget that contained both bookmarks and recent items. The rationale is that a single widget is easier to use and understand, even if it is a little overloaded under the hood.

With two widgets it means there are two things to scroll and there's some redundancy across the lists because repos can be in both the Recents and Favorites, which is not quite as nice.

What do you think? Is it better to present them in a single view or is the screen real estate and complexity of two views worth it?

cola/widgets/startup.py
```diff
@@ -265,16 +336,397 @@ def handle_broken_repo(self, index):
...
 
+    def set_model(self, items, mode):
+        if mode == LIST_VIEW_BOOKMARK:
+            model = self.bookmarks_model
+            self.bookmark_items = new_items = []
+        elif mode == LIST_VIEW_RECENT:
+            model = self.recent_model
+            self.recent_items = new_items = []
+        else:
+            return
```
davvid, line 373: Is the `else:` ever reached, or is it unreachable?

This is another example where the internal complexity might be able to be reduced. The `mode` is basically deciding which of the internal two models to use. That seems like it can be improved by factoring that stuff out.

cola/widgets/startup.py
```diff
@@ -265,16 +336,397 @@ def handle_broken_repo(self, index):
...
+        self.is_default = is_default
+        self.name = name
+
+        user_role = Qt.UserRole
+        self.setEditable(False)
+        self.setData(path, user_role)
+        self.setIcon(icon)
+        self.setText(name)
+        self.setToolTip(path)
+        self.setFlags(self.flags() | Qt.ItemIsEditable)
```
davvid, line 718: Is it better to use `setEditable(bool)` rather than using `setFlags()`?

davvid: Also.. it looks like we call `setEditable(False)` right above here...

cola/widgets/startup.py
```diff
@@ -68,46 +76,52 @@ def __init__(self, context, parent=None):
...
 
-        directory_icon = icons.directory()
-        user_role = Qt.UserRole
-        normalize = display.normalize_path
-        paths = {normalize(repo['path']) for repo in all_repos}
-        short_paths = display.shorten_paths(paths)
-        self.short_paths = short_paths
+        self.recent_title_label = qtutils.label(
+            text="Recent", align=Qt.AlignHCenter | Qt.AlignVCenter, selectable=False
+        )
```
davvid, line 99: TBD whether splitting Recents and Favorites is worth it, but just a note that all translatable UI strings should use the `N_(...)` function for i18n.

This should be `N_('Recents')` here, and `N_('Favorites')` down below.

cola/widgets/startup.py
```diff
@@ -7,18 +7,29 @@
...
 from . import standard
 
 
+ICON_MODE = 0
+LIST_MODE = 1
+
+LIST_VIEW_BOOKMARK = 0
+LIST_VIEW_RECENT = 1
+
+LABEL_TITLE_FONT_SIZE = 16
```
davvid, line 30: We probably don't want to hard-code a font size... I'll see where it's used to suggest an alternative.

cola/widgets/startup.py
```diff
@@ -68,46 +76,52 @@ def __init__(self, context, parent=None):
...
-        normalize = display.normalize_path
-        paths = {normalize(repo['path']) for repo in all_repos}
-        short_paths = display.shorten_paths(paths)
-        self.short_paths = short_paths
+        self.recent_title_label = qtutils.label(
+            text="Recent", align=Qt.AlignHCenter | Qt.AlignVCenter, selectable=False
+        )
+        title_font = self.recent_title_label.font()
+        title_font.setBold(True)
+        title_font.setPixelSize(LABEL_TITLE_FONT_SIZE)
```
davvid, line 102: Instead of using a hard-coded font size, lookup the current font size and then scale by some factor (eg. `0.8`). Is the default size too small or too large?

cola/widgets/startup.py
```diff
@@ -158,24 +167,58 @@ def __init__(self, context, parent=None):
...
+            self.bookmarks.change_rename_enabled(rename_enabled)
+            self.recent.change_rename_enabled(rename_enabled)
+
+            model.setViewMode(mode)
+            model.setIconSize(icon_size)
+            model.setGridSize(grid_size)
+            for item in items:
+                if isinstance(item, PromptWidgetItem):
+                    item_name = item.change_path_text(name_mode)
+                    model.rename_entry(item, item_name)
```
davvid, line 221: I'm confused about this part -- is this going and calling `rename_entry()` on each item?

In theory that's renaming all of the repos. Am I reading that correctly?

I recall that switching between modes was pretty slow.. if so, this might explain why. It's actually renaming things to their same values and saving the settings file on each line in this loop.

That seems.. not good. That makes me think we shouldn't even allow renaming in this view. What do you think?

davvid (changes requested):
Thanks for picking this up. I've got a bunch of question and notes, and there's a key behavior in `tab_changed()` that looks like a bug.

I think it'd be good to add more context menu actions, but I would suggest that we not add renaming, and that we try to avoid splitting the view if possible.

I'm open to hearing other opinions as to whether or not we should split the views, but that's my current take on it at this time.

Let me know what you think.

davvid:
Meta: we have this weird special `Open...` folder entry that behaves unlike the rest of them in the UI.

I almost think we should remove that special entry and rename the `Open...` button on the bottom left to `Browse...` for clarity. I suspect that might make the code a little simpler as well.

cola/widgets/startup.py
```diff
@@ -265,16 +336,397 @@ def handle_broken_repo(self, index):
...
+        settings = context.settings
+        normalize = display.normalize_path
+
+        paths = set([])
+        if self._is_bookmark_mode():
+            for bookmark in settings.bookmarks:
+                paths.add(normalize(bookmark['path']))
+        elif self._is_recent_mode():
+            for recent in settings.recent:
+                paths.add(normalize(recent['path']))
```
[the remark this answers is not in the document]
nakanoi: I see. Yes, `_is_bookmarks_mode()` and `_is_recents_mode()` become unnecessary.
I'll try to change it!

cola/widgets/startup.py
```diff
@@ -265,16 +336,397 @@ def handle_broken_repo(self, index):
...
+
+        if isinstance(item, PromptWidgetItem):
```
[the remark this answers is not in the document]
nakanoi: This is because each view("bookmarks" and "recent") has `Open...` item, and it is `QStandardItem`, not `PromptWidgetItem`. Not to open context menu with `Open...` item click, I added `isinstance` filter.
 
> Meta: we have this weird special Open... folder entry that behaves unlike the rest of them in the UI.
> 
> I almost think we should remove that special entry and rename the Open... button on the bottom left to Browse... for clarity. I suspect that might make the code a little simpler as well.

As following this comment, removing `Open...` item and all `isinstance(item, PromptWidgetItem)` filters from code seems good. How about this?

cola/widgets/startup.py
```diff
@@ -265,16 +336,397 @@ def handle_broken_repo(self, index):
...
+        if not item:
+            return
+        ok, _, _, _ = cmds.do(cmd, context, item.path, item.name, icon=icons.discard())
+        if ok:
+            self.refresh()
+
+    def remove_missing(self):
+        """Remove missing entries from the favorites/recent file list"""
+        settings = self.context.settings
+        settings.remove_missing_recent()
```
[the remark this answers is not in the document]
nakanoi: OK! I'll do it.

cola/widgets/startup.py
```diff
@@ -265,16 +336,397 @@ def handle_broken_repo(self, index):
...
+        self.refresh()
+
+
+def _disable_rename(_x, _y, _z):
+    return False
```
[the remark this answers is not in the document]
nakanoi: this method is never reached and for completeness.
OK, I'll remove it.

cola/widgets/startup.py
```diff
@@ -158,24 +167,58 @@ def __init__(self, context, parent=None):
...
-            list_mode = 'list'
-            for item in self.items:
-                path = item.data(Qt.UserRole)
-                item.setText(path)
+        models_items = [
+            (self.bookmarks, self.bookmark_items),
+            (self.recent, self.recent_items),
+        ]
+        for model, items in models_items:
+            if idx == 0:
```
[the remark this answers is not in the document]
nakanoi: Yes! I'll improve it.

cola/widgets/startup.py
nakanoi: I see. I had no such idea.
I splited into two views because of the reasons below. But doing that made many things complecated and might feels redundant.

> - View spliting
> I thought too.
> But a case like below could be possible. `bookmarks` and `recent` widgets in main UI display repository's name, and I thought splitiing the view is one of the solutions for clearfying which name should be displayed.
> And I also thought that displaying two views seems to be more usable.
> But after I read all comments, I though *view shouldn't be splited*.

So, changing the code to display only bookmarks list view might be good. Is it OK?

cola/widgets/startup.py
nakanoi: This `else:` never reached and for completeness too, and it might be redundant.
OK, I'll eliminate it.

cola/widgets/startup.py
nakanoi: Oh, using `setEditable(bool)` is enough and `setFlags` is overload.
I'm really sorry.

cola/widgets/startup.py
nakanoi: OK. I'll change it.

cola/widgets/startup.py
nakanoi: Yes, as this comment, I'll improve it.

>Instead of using a hard-coded font size, lookup the current font size and then scale by some factor (eg. 0.8). Is the default size too small or too large?

cola/widgets/startup.py
nakanoi: That is the coolest way and I've never come up with it.
Thank you! I'll improve it.

cola/widgets/startup.py
nakanoi: Yes, it is corrent. And I think so too.

When change Tabs between "Folder" and "List", item text should be changed(`name` <=> `path`).
I added `rename_entry()` here because somehow `item.setText()` in `change_path_text()` emits `itemChnaged` signal. So, `context.settings.bookmarks` and `context.settings.recent` get changed.

To keep consistency, call `rename_entry()` after `change_path_text()` call.
But, by disabling renaming, `itemChange` signal becomes unnecessarry and this part gets to be able to eliminated.

nakanoi:
Thank you for detailed comments!

- Disable rename entry action
I agree to it. I added this action because `bookmarks` and `recent` widgets have it. So I'm going to remove it from context menu.

- View spliting
I thought too.
But a case like below could be possible. `bookmarks` and `recent` widgets in main UI display repository's name, and I thought splitiing the view is one of the solutions for clearfying which name should be displayed.
And I also thought that displaying two views seems to be more usable.
But after I read all comments, I though *view shouldn't be splited*.

```python
# Double name entry for the same repository("path": "/foo")
context.settings.bookmarks = [
  {"name": "foo", "path": "/foo"},
]
context.settings.recent = [
  {"name": "buzz", "path": "/foo"},
]
```

> This code also originally only dealt with the bookmarks list. It'd almost be better to not change that behavior for now so that it doesn't need to have this complicated "other model" coupling going on.

So, changing the code to display only bookmarks list view might be good. How about it?

---
I'm very motivated to improve all of them. Thank you for your cooperation!

davvid:
> Meta: we have this weird special `Open...` folder entry that behaves unlike the rest of them in the UI.
> 
> I almost think we should remove that special entry and rename the `Open...` button on the bottom left to `Browse...` for clarity. I suspect that might make the code a little simpler as well.

We can ignore this comment now. I familiarized myself with the code and made some minor aesthetic changes to the UI to make that item a little more intuitive. Please rebase.

I've renamed the special `Open...` item to `Browse...`. We'll keep that entry. It's convenient because it lets us hit enter when the dialog pops up to go straight into a file browser.

The separate button, which also had an `Open...` label, was renamed to `Open Git Repository` so that it makes it a little clearer that pressing it will open the currently selected item.

I want to call out this section in the original code where it builds the items:

```python
        bookmarks = settings.bookmarks
        recent = settings.recent
        all_repos = bookmarks + recent
        ...
        for repo in all_repos:
            ...
            item = QtGui.QStandardItem(path)
```

If we stick to a single view then we can push the "is this path bookmarked?" knowledge down into the item at this point by tweaking this loop. eg. by using `enumerate` we can detect whether we're looping over a bookmark or a recent repo.

```python
for idx, repo in enumerate(all_repos):
    is_bookmark = idx < len(bookmarks)
    ...
```

and then it can use `is_bookmark` when building the item so that the context menu can be customized?

I take it back about not having the "Rename" action -- that's pretty useful so no reason not to have if possible.

Sorry for not being in favor of having two list views instead of one. If anything, I hope it might simplify the implementation a bit.

Also, it might be the case that we want to remove a bookmark but still keep the item in the "Recents" area, so having the handling be that it detects bookmarks and only removes Bookmarks/Favorites when detected, and otherwise attempts to remove from the Recents is probably the right behavior. That's better than removing from both lists using a single action.

I hope that makes sense.

Also, not sure if you noticed, but the code uses the term "bookmark" but the UI presents it as "Favorites". That's a historical thing -- it was originally called "Bookmarks" to match firefox jargon. #289 is when that was renamed. We didn't change the class names when we did that -- we only changed the labels use the term "Favorites" but kept everything else using the Bookmarks nomenclature.

Let me know what you think so far. Here's a quick summary of my general sugs...

* Use a custom item so that it can remember whether or not it's from the bookmarks
* If it's from the Bookmarks then the bookmark-related remove/rename should be used
* If it's from the Recents then the recents-related remove/rename should be used
* Keep all of the context menu actions you've added. Those are great.

Thanks again!

davvid:
>  But after I read all comments, I though view shouldn't be splited.

Thank you, I kinda reached the same conclusion. I was a little on the fence too..

it's kinda nice with them split, but sometimes less is more, so I tend to go with a "simpler" choice when possible. It also helps out new users that don't yet have any Recents of Favorites ~ seeing the `Browse...` entry in two lists without anything else in them might be kinda confusing. Thanks again!

cola/widgets/startup.py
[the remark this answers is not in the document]
davvid: About `remove_missing` -- I think this action doesn't even need to be item-specific. I think we would want to remove missing bookmarks and missing recent entries all in a single call here.

For what it's worth, the remove-missing context menu entry seems like it should be present even when right-clicking in the widget in the blank area without any icons. Not important for now, but just noting in case it helps. This should be the only UI action that will modify both the bookmarks and recents from a single action. The rest can limit their scope to bookmarks when `is_bookmark` is True.

davvid:
(I think this is implied in what I wrote earlier, but I didn't respond to this question so I'll do so now)

> So, changing the code to display only bookmarks list view might be good. Is it OK?

We'll want to continue displaying both the Bookmarks and the Recent items, in the same order it currently displays them (bookmarks first, and then recent items) in a single View. Using the `is_bookmark` flag to special-case the Items seems like it should give us enough context to specialize their context menu entries and editable states.

Having some `if self.is_bookmark` logic to decide on the functions/cmds to run down inside the item hopefully won't be too bad. Better might be to push that into BuildItem so that the custom Item class can be simple, but it's fine if it's just a small handful of spot checks.

nakanoi:
Thank you for many great tips and idea!

It's OK. Implementing SIngle View might make things simpler.
I got understand  that "bookmarks" and "favorite". I'll care of them.
I'll add `is_bookmark` flag attribute to custom Item Class(`PromptWidgetItem`) for identifying between bookmarks and recent.

I think your suggestions are best and make sense. I'll try to implement improvemnt all of them!

I might ask some questions while implemnting them, please forgive me.
Thank you!

nakanoi:
Hi @davvid!
I implemented all given suggestions!

- Changed to single view
- Removed followings
  - `_disable_rename()` and unreachable `else` statements
  - unnecessary `setFlags()` and remain `setEditable()`
  - translatable `N_('Favorites')` and `N_('Recent')`
    - There is no need to use them because labels become unnecessary.
  - Font size hard coding
    - Same as `N_('Favorites')` one.
- `remove_missing()`
  - Remove missing repository from both bookmarks and recent list.
- Use custom item to keep whether the item comes from bookmarks list or not
- Delete/Rename item....
  - item in bookmarks list if it comes from bookmarks list
  - item in recent list if it comes from recent list
  - item in bookmarks list if it comes from both bookmarks and recent list

---

On the other hand,

- Still use `isinstance(item, PromptWidgetItem)`
  - To filter items because `self.items` contains a `QStandardItem` for `Browse...` item and repository items for `PromptWidgetItem`.
- Remove `model.rename_entry()` call.
If `item.setText()` called, then `itemChanged` signal will be emitted and that slot changes item name unproperly.
So, instead of calling `item.setText()`, I implemented `tab_change` to re-instancinate all items like `StartupDialog.set_model()`.

---

If I misunderstood something, please tell me.
Thank you!

davvid (changes requested):
This looks really great, it's almost ready to merge. Mostly small cosmetic things at this point.

Once it's ready, please rebase this branch and squash the commits together a bit. We won't really need the intermediate commits -- might as well squash things together a bit more. It seems like 2 or 3 commits at most should be sufficient:

* `starup.py: Change item to QStandardItem implement`
* [new squashed commit that adds the new widgets and tests]

If you're comfortable with git you can even tease apart a 3rd commit so that the tests can be committed separately.

Thanks for taking it this far! It's pretty much ready to go.. this last part is mostly just patch polish. This looks great!

nakanoi:
Thank you for re-reviewing!
After change, I'll clearfy commits and push it!

nakanoi:
Hi @davvid!
I fixed code based on the gaven comment and reabase commits!
Thank you.

davvid:
👏 👏 👏  very nicely done! Thanks again for improving cola.

nakanoi:
@davvid
Thank you for merging this PR!
I could'nt be done without your greateful help. 
Thank you so much again!

Issue #1090, opened 2020-09-28
UI for lots of repos


ewjoachim:
Hey there,
I'm a long time fan of the project.

In the last months, I've became involved in many different repositories (30+). I used to keep all the repos I interacted with in "Favorites" but it's becoming less and less usable. I can always open a repo with ctrl+o but that's not the most practical thing to do. I'm wondering if there's something I might contribute (or some existing feature I missed) that could help for my case. Ideally I'd use a shortcut, type a few letters of the repo I'm interested in, and then either click or use the arrow keys to select it. A "repo quick-switcher", kind of like what there is in recent editors (usually mapped to `cmd/ctrl+p`).

So is there something of the likes already existing ? If not, are there some places in the codebase you recommand as inspiration if I were to build it ?

davvid:
This seems like it'd be pretty useful. The "Recent" and "Favorites" widgets do let you type in stuff and it'll select one of the items in the list. Then, hitting "enter" will switch to that repo. It'd be cool if this were available globally through a shortcut like you suggested.

ewjoachim:
Hey :) Thanks for you answer. I'm not sure I will have time to tackle this now. If anyone is interested, please feel free!

nakanoi:
Hi @davvid @ewjoachim !
I tried to implement these features in this PR #1282.
- filtering repositories on bookmarks(recent) widget
- global shortcut for filtering repositroies

I would be very happy if you could take a look at it.
Thank you so much!

ewjoachim:
[Screencast from 13-01-2023 13:28:39.webm](https://user-images.githubusercontent.com/1457576/212320577-162245e1-6daf-4e85-a39e-bcd3d95745ca.webm)

Nice !
2 very optional comments:
- The way "ctrl+p" menus usually work is that when you press enter, the first element is activated. In this case, you need to press down. Ideally, make the first element already selected
- It's filtering by trying to search the string you typed inside each repo. It would be marginally better if it was looking for any repo that have the same letters as you typed, but with potentially other letters in between. For example, if we type `pycocoac`, `python-coverage-comment-action` should be available (if we want it to work the same way as the ctrl+p or ctrl+t menus from Sublime text, VS Code, Slack, Discord etc)

ewjoachim:
For the 2nd item, I think it would be simple as
```python
self.proxy_model.setFilterRegExp(".*".join(input_text))
```
[...]

PR #1282, opened 2023-01-12
Add shortcut for filtering repositories in bookmarks widget and main widget

nakanoi:
Add repository name filtering component by [QSortFilterProxyModel](https://doc.qt.io/qtforpython/PySide6/QtCore/QSortFilterProxyModel.html). `Switcher` class in `cola/widgets/switcher.py` inherits it.

> NOTE
> - In this sentence, I use a word `moving key`. It means "selection move key like `UP`/`DOWN`".
> All moving keys is in `selection_move_keys` variable.
>

## Filtering Class

For versatility and make it reusable, impmlemented `Switcher` class to be able to be with it's list view or without it.
By passing candidate items which is or inherits `QStandardItemModel`, this can filter them by item name in it.

In this PR, I implemented to filter repositories. But, for example, if `QStandardItemModel` like "file" or "commit" passed, this can filter them.

I'll explain with-view/without-view class details below.

### with-view class(`SwitcherInnerView`)

This contains "input field"(`self.filter_input`) and "candidate items list view"(`self.switcher_list`) in one widget.

### without-view class(`SwitcherOuterView`)

This will show only "input field". So, this class only do filtering candidate items by `self.proxy_model`.

When some characters inputted to the field, this just filters model(`self.proxy_model`).
By sharing model between view class in other widget class and `SwitcherOuterView`, the candidate items will be filtered even though the view class is in other class.

- To notify selection change event happended(moving key has pressed) while focusing on *input field*, this will emit `switcher_selection_move` signal.
- To notify that some character has benn inputted while focusing on *bookmamarks tree view*, this will emit `switcher_text` signal.

These are to be able to do input text and select item without changing focus.

### Filtering repositories in bookmarks(recent) widget on main widget

- used `SwitcherOuterView`. That means this only shows input field and uses the original `BookmarksTreeView` class(but changed `TeeWidget` to `TreeView` for sharing model).
- by pressing some key except for move key while bookmarks(recent) widget is focused, input field will be showd and filter items by inputted text.

### Global shortcut for filtering repositories on main widget

- used `SwitcherInnerView`. That means one dialog(`SwitcherInnerView`) contains input field and candidate items view.
- Enter/Double click/Activated will open selected repository.
- `ctrl/command+p` has already mapped to Pull action, mapped this widget opening to `Alt+p`.

checked `make check`
improves #1090

Thank you!

### DIFF
--- a/cola/widgets/switcher.py
+++ b/cola/widgets/switcher.py
@@ -122,6 +122,8 @@ class SwitcherInnerView(Switcher):
         self.filter_input.switcher_selection_move.connect(
             self.switcher_list.keyPressEvent
         )
+        # escape key pressed while focusing on input field
+        self.filter_input.switcher_escape.connect(self.close_widget)
         # some key except moving key has pressed while focusing on list view
         self.switcher_list.switcher_inner_text.connect(self.filter_input.keyPressEvent)
 
@@ -145,6 +147,9 @@ class SwitcherInnerView(Switcher):
         item = self.switcher_list.model().itemFromIndex(index)
         if item:
             self.enter_action(item)
+        self.close()
+
+    def close_widget(self):
         self.close()
 
 
@@ -179,6 +184,7 @@ class SwitcherLineEdit(text.LineEdit):
     # while focusing on this line edit widget
     switcher_selection_move = Signal(QtGui.QKeyEvent)
     switcher_visible = Signal(bool)
+    switcher_escape = Signal()
 
     def __init__(self, place_holder=None, parent=None):
         text.LineEdit.__init__(self, parent=parent)
@@ -197,7 +203,9 @@ class SwitcherLineEdit(text.LineEdit):
         selection_moving_keys = moving_keys()
         pressed_key = event.key()
 
-        if pressed_key in selection_moving_keys:
+        if pressed_key == Qt.Key_Escape:
+            self.switcher_escape.emit()
+        elif pressed_key in selection_moving_keys:
             self.switcher_selection_move.emit(event)
         else:
             super().keyPressEvent(event)
--- a/cola/widgets/switcher.py
+++ b/cola/widgets/switcher.py
@@ -1,6 +1,8 @@
 """Provides quick switcher"""
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import re
+
 from qtpy import QtCore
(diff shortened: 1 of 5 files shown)


cola/widgets/switcher.py
```diff
@@ -0,0 +1,254 @@
...
+            parent=parent,
+        )
+        self.filter_input.setVisible(False)
+
+        self.main_layout = qtutils.vbox(defs.no_margin, defs.spacing, self.filter_input)
+        self.setLayout(self.main_layout)
+
+    def filter_input_changed(self):
+        input_text = self.filter_input.text()
+        self.proxy_model.setFilterRegExp(input_text)
```
ewjoachim, line 160:
```suggestion
        super().filter_input_changed()
```

cola/widgets/switcher.py
```diff
@@ -0,0 +1,254 @@
...
+        self.filter_input = SwitcherLineEdit(place_holder=place_holder, parent=self)
+
+        self.proxy_model = SwitcherSortFilterProxyModel(entries_model, parent=self)
+        self.switcher_list = None
+
+        self.filter_input.textChanged.connect(self.filter_input_changed)
+
+    def filter_input_changed(self):
+        input_text = self.filter_input.text()
+        self.proxy_model.setFilterRegExp(input_text)
```
ewjoachim, line 85:
```suggestion
        pattern = ".*".join(re.escape(c) for c in input_text)
        self.proxy_model.setFilterRegExp(pattern)
```

davvid:
This is a really slick feature, thanks @nakanoi and thanks @ewjoachim for reviewing.

Heads-up that I've pushed a few additional commits on top of these changes to the repo-name-filtering-issue-1090 branch in my [davvid/git-cola](https://github.com/davvid/git-cola) fork.

I addressed the review notes here and adjusted the widget to not close itself by accident. I also changed it into a window-modal dialog.

I plan to merge this soon in the next day or so. Please give that branch and spin and let me know what you think of the tweaks I made.

It's in the main File menu now too. Good stuff!

nakanoi:
@davvid Thank you for reviewing and improving my change!

I saw your change on my computer. That is so good than my change!
I assumed "ctrl+p" that appears small at the top of the window like VSCode, but I think your implementation is easier to use.

I like it!

PR #1290, opened 2023-02-12
Allow editing of patches before staging / unstaging / reverting

### LINKED ISSUE #794: [Feature Request] Allow editing hunks in staged view
During coding I often come across wanting to stage some changes I made some hours before so I can cherry pick them into my upstream/integration branch. This already works very well and convenient with git-cola due to its feature of selecting single lines for staging.

But I also come often find that I already added some feature or refactoring into the hunks which is far from being ready but I still want to just fix that one bug. What I currently do is using copying the relative path from git-cola, then `git add --patch <filename>` and manually edit the hunk to be compatible with upstream. Afterwards, it automatically shows up correctly in git-cola which is good. But I'd like to do this right inside git-cola.

I could imagine doing this directly from the stage view: It should be possible to right click on a hunk, select "Edit hunk" and an internal editor (or external editor if configured) fires off to let me do the final tweaks before committing. I'd prefer an internal editor to have it well integrated and this is really only for very easy tweaks (otherwise you're doing something wrong probably). The editor should be able to check that the hunk stays valid and support the usual workflows of `git add --patch` by allowing to convert between `+`, `-` and ` ` patch lines.

Such a feature would also need another context menu item in the stage view: You should be able to split hunks (if possible) so that you can break it down to just the hunk you want to edit.

I hope I could properly line out what I'm trying to achieve...

Here's some more read-up:
  1. https://kennyballou.com/blog/2015/10/art-manually-edit-hunks/
  2. https://johnkary.net/blog/git-add-p-the-most-powerful-git-feature-youre-not-using-yet/

living180:
This PR contains some significant refactoring of the diff parsing code in `cola/diffparse.py` in order to facilitate adding the ability for the user to manually edit patches in an external editor before staging / unstaging / reverting.

I did make one significant departure from git's behavior in `git restore --patch`:  when the user runs this command and elects to edit a hunk, git launches the editor with the normal forward patch, then reverses it and applies it after the user edits it.  This has always been extremely difficult for me to get right if I want partially unstage / revert a line.  Thus the behavior I have implemented for git-cola is launching the editor with the already-reversed patch, which will be applied directly after being edited by the user.  I am open to input if this difference from git's behavior will be problematic for users, but at least for me it is much more intuitive.  (The behavior of editing a patch to stage mirrors the behavior of `git add --patch` however.)

I am also open to suggestions to improvements on the actual text of the menu item names.  I also added hotkeys of _Ctrl-Shift-S_ for _Edit to Stage_ / _Unstage_ and _Ctrl-Shift-U_ for _Edit to Revert_ but I don't know if those are the best choices.

There are additional improvements that could be made to this feature, such as trimming the context lines sent to the editor when selecting just a few lines of a large hunk for editing, but I decided to wait on those until getting feedback on the core implementation.

I am submitting this as a draft for now and can convert it to a normal PR once everything looks good.

Closes #794


davvid:
> I  did make one significant departure from git's behavior in git restore --patch: when the user runs this command and elects to edit a hunk, git launches the editor with the normal forward patch, then reverses it and applies it after the user edits it. This has always been extremely difficult for me to get right if I want partially unstage / revert a line. Thus the behavior I have implemented for git-cola is launching the editor with the already-reversed patch, which will be applied directly after being edited by the user. I am open to input if this difference from git's behavior will be problematic for users, but at least for me it is much more intuitive. 

:+1:  on me on this. I also feel like what's implemented here is more intuitive. It's certainly simpler since it ends up being applied directly, which at least for me makes it easier to understand.

I haven't started using the `def x(..., *, ...)` trick to force use of named keyword parameters. That's pretty neat.

Overall this looks good to me. I'll probably merge this as-is unless you have any changes you think are worth making beforehand. Please feel free to squash in a tweak to change it to use `utils.tmp_filename('edit', suffix='.patch')` in the keyword arguments spirit of things.

For a moment here I thought that maybe there was a bug where we weren't checking an exit code. It turns out that I found a bug in nvim-qt while testing this. In gvim if we exit with `:cq` it'll use exit code 1 when exiting. It looks like nvim-qt exits with exit code 0, and thus didn't abort from applying the patch.

I'll file a bug for that one. Anyways, that's unrelated to these changes.

Thanks for adding this feature. This is really supreme and it landed just in time to be a flagship feature for v4.2.0.

davvid:
I pushed the merge of the pull request and the trivial keyword argument tweak in  c7ef611fa993 to the `next/edit_patches` branch of my fork https://github.com/davvid/git-cola/tree/next/edit-patches

I'll push that branch to main as soon as you mark this PR as ready. (unrelated: nvim-qt had already fixed the bug I ran into ~ I was running an old version)

living180:
Sounds good - if you are happy with the user-facing parts such as the menu item names and the keyboard shortcuts, I'm happy with the internals.  Good for merging on my end.

PR #1292, opened 2023-02-14
Clean up dev and tox requirements

living180:
Remove obsolete and unneeded packages from `tox.ini` and `requirements-dev.txt`.  Also drop `pylint-flake8` since it is essentially abandoned upstream, doesn't work with pytest>=7.0 (released February 2022) or flake8>=5.0 (released in July 2022) and instead simply invoke `flake8` directly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,8 +49,7 @@ jobs:
               pylint \
               python-is-python3 \
               python3-setuptools-scm \
-              python3-pytest \
-              python3-pytest-flake8
+              python3-pytest
           # Runtime dependencies (required)
           sudo apt-get install \
               python3-qtpy \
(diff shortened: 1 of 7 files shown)


davvid:
Thanks for updating these.

At some point I probably should start drinking my own koolaid and [replace the Makefile with a garden.yaml file](https://github.com/davvid/garden/blob/main/garden.yaml).

Well.. maybe not "replace" it since everyone already has `make` installed. It's tempting because `garden` has simpler syntax and IMO easier to understand compared to a Makefile for what we do with it.

`garden` isn't intended to be a `make` replacement, but our `make` usage is pretty much just, "run these commands" and would be simplified a bit if we did swap over.

It would be a hassle for packagers, though, and we already had one breaking build system change when I got rid of our distutils usage as part of the PEP-517/518 build system updates. With that in mind, it's probably best to leave the Makefile in-place for the next few years at least.

I could have them side-by-side and mirror the functionality between them as a first step. It'd be a little bit of duplication for a while but would also start setting the road for an eventual transition if we ever decided to do so.

Anyways, I digress. I should probably see if there are other places where I'm using these deprecated modules and clean stuff up there too. Thanks again!

PR #1294, opened 2023-02-17
Updated qtutils.py (Replace "Select manually..." with "Browse..." in the open repository dialog.)

meaviral17:


davvid (changes requested):
This doesn't work. You should specify this parameter at the call site.

Also, this is breaking all of the current callers of this function because they all use the current signature.

Please don't change the signature of this function. Update the call site so that the title can be changed there.

davvid:
For reference, the call sites should be in `cola/guicmds.py`, specifically `def open_repo()`.

That said, I'm not sure if this is the "Select Manually..." text you're trying to change. I don't even think we use that label anymore?

davvid:
Also, I believe this might be referring to the dialog you see when running `git cola --prompt`

Also, that dialog already uses the text "Browse..." in the current version in git. Have you tested that version? Feel free to reach out if you have any questions of if you think this PR can be reopened for a different purpose.

PR #1299, opened 2023-02-25
Correct PyQt version that pip installs

wojnilowicz:
On a system where the Qt version is 5.13.2 "./env3/bin/pip install git-cola" installs PyQt in the version 5.15.9 (the latest). Subsequent "./env3/bin/git-cola" doesn't open the app window and causes following error to appear "No Qt bindings could be found". This patch fixes the issue by installing the PyQt version matching the Qt version on the system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Git Cola is a powerful Git GUI with a slick and intuitive user interface.
 
 Git Cola uses QtPy, so you can choose between PyQt6, PyQt5 and PySide2 by setting
 the `QT_API` environment variable to `pyqt6`, `pyqt5` or `pyside2` as desired.
-`qtpy` defaults to `pyqt6` and falls back to `pyqt6` and `pyside2` if `pyqt5`
+`qtpy` defaults to `pyqt5` and falls back to `pyqt6` and `pyside2` if `pyqt5`
 is not installed.
 
 Any of the following Python Qt libraries must be installed:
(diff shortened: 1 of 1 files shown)


davvid:
Just curious -- we ~somewhat~ recently patched those instructions to use `--system-site-packages`. Have you tested creating a new virtualenv from scratch to see whether or not you need this step? You'll have to `rm -r env3` to test it out from a clean slate.

My (naive) assumption was that using `--system-site-packages` when building the virtualenv would ensure that the system-provided PyQt packages would be used when they've already been installed.

I just tested this out by creating the virtualenv from scratch and it seemed to do the right thing. That said, `--system-site-packages` only helps if the user already have PyQt5 installed. It does not help if they *only* have Qt5 installed.

Not all users will have `qmake` installed, though, so I think we can augment the instructions slightly to mention these caveats ~ namely that it is only for the situation when:
* You don't already have PyQt5 installed
* You already have Qt5 and its qmake devtools installed
* You want the virtualenv to install a PyQt5 version that is compatible with your existing Qt5 installation

davvid:
NOTE: The CI / Test error has already been fixed so you'll pickup that fix when you rebase against the latest.

wojnilowicz:
> Just curious -- we somewhat recently patched those instructions to use --system-site-packages. Have you tested creating a new virtualenv from scratch to see whether or not you need this step? You'll have to rm -r env3 to test it out from a clean slate.
Yes.

> My (naive) assumption was that using --system-site-packages when building the virtualenv would ensure that the system-provided PyQt packages would be used when they've already been installed.

No PyQt packages on my system, so they need to be downloaded by pip.

> Not all users will have qmake installed, though, so I think we can augment the instructions slightly to mention these caveats ~ namely that it is only for the situation when:

You're right. It's part of a devel package on my system. I'm not sure what to do. Your instruction doesn't work for me but on the other hand the updated instruction might not work for others.

davvid:
Thanks for the sug. I tweaked the details a little bit to clarify and hopefully help more users that read this section.

wojnilowicz:
You're welcome, but I'm not sure if 

> QT_VERSION_MAJOR=$(qmake -query QT_VERSION)

is what you wished to write. This will return the full version signature instead of it's major part.

davvid:
Yup, I had a typo in the initial write-up. The commit that followed should have cleared that up. Thanks for the heads-up.

Issue #1312, opened 2023-05-05
Custom date/time format in git-dag list


jm4R:
Can I somehow customize date/time format in git-dag list? It seems hardcoded and I would prefer `YY/MM/DD hh::mm::ss` and day of the week is not important to me.

davvid:
This would be a good thing to add as a configuration variable.
Right now `git log` defaults to `--date=short` but we can make it accept any of the `--date` parameters.

See `git help log` and look for the `--date=` documentation for what's supported by git.

Let me know if any of these options would result in a better display. If so we'll make it configurable.

https://www.git-scm.com/docs/git-log#Documentation/git-log.txt---dateltformatgt

kshitiz305:
I would like to contribute please let me know which one would you like to choose?

davvid:
Thanks @kshitiz305 that'd be awesome. You can find most of the handling for preferences in the following files:
* `cola/models/prefs.py`
* `cola/widgets/prefs.py`

As far as what to call the preference there may be two options here. I'm open to hearing opinions on which we should follow.

Option (1) is to follow the existing `log.date` configuration value that `git` already supports[1]. This is kinda cool because the command-line `git log` and the GUI `git dag` interfaces will display the same date values.

[1] https://www.git-scm.com/docs/git-log#Documentation/git-log.txt-logdate

The downside is that they are forced to display the same value and there's no way to differentiate them from a configuration perspective.


Option (2) is to use a new preference `cola.logdate` that accepts the same values as core git's `log.date` configuration value. I currently believe that this might be the best option because the upside is that we can configure these two features separately.


In order to integrate / implement it, you'll want to adjust the `RepoReader` constructor in `cola/models/dag.py`. There's a `git log` invocation there and it can read the config from `context.cfg.get('cola.logdate')`.

If `cola.logdate` is unset or an empty value then it should use the existing `self._cmd` argument list as-is. But, if `cola.logdate` is configured then it should append `--date=...` to the `self._cmd` list.

Lastly, please make sure to update the configuration section in `doc/git-cola.rst` to document this new configuration value.

The description of the config value should probably be something like this (this is basically the documentation for `log.date` but tweaked for cola's configuration):

> Set the default date-time mode for the DAG display. Setting a value for `cola.logdate` is similar to using `git log`'s `--date` option. See `git-log(1)` for details.

DasaniT:
Since it's been a month and no activity, can I work on this? And if so, should I implement option 2?

[...]

PR #1319, opened 2023-06-20
prefs: add log date format to preferences for git-dag

DasaniT:
I've added the log date format pref discussed in #1312.

### DIFF
--- a/cola/models/dag.py
+++ b/cola/models/dag.py
@@ -3,6 +3,7 @@ import json
 
 from .. import core
 from .. import utils
+from ..models import prefs
 
 # put summary at the end b/c it can contain
 # any number of funky characters, including the separator
@@ -273,7 +274,12 @@ class RepoReader(object):
 
         self.reset()
         ref_args = utils.shell_split(self.params.ref)
-        cmd = self._cmd + ['-%d' % self.params.count] + ref_args
+        cmd = (
+            self._cmd
+            + ['-%d' % self.params.count]
+            + ref_args
+            + ['--date=%s' % prefs.logdate(self.context)]
+        )
         self._proc = core.start_command(cmd)
 
         while True:
--- a/cola/widgets/prefs.py
+++ b/cola/widgets/prefs.py
@@ -141,6 +141,10 @@ class RepoFormWidget(FormWidget):
         self.tabwidth = standard.SpinBox(value=8, maxi=42)
         self.textwidth = standard.SpinBox(value=72, maxi=150)
 
+        self.logdate = qtutils.combo(prefs.date_formats())
+        tooltip = N_('This value is passed to git log --date')
+        self.logdate.setToolTip(tooltip)
+
         tooltip = N_('Detect conflict markers in unmerged files')
         self.check_conflicts = qtutils.checkbox(checked=True, tooltip=tooltip)
 
@@ -160,6 +164,7 @@ class RepoFormWidget(FormWidget):
         self.add_row(N_('Email Address'), self.email)
         self.add_row(N_('Tab Width'), self.tabwidth)
         self.add_row(N_('Text Width'), self.textwidth)
+        self.add_row(N_('Log Date Format'), self.logdate)
         self.add_row(N_('Merge Verbosity'), self.merge_verbosity)
         self.add_row(N_('Number of Diff Context Lines'), self.diff_context)
         self.add_row(N_('Patches Directory'), self.patches_directory)
@@ -196,6 +201,7 @@ class RepoFormWidget(FormWidget):
                     Defaults.display_untracked,
                 ),
                 prefs.ENABLE_GRAVATAR: (self.enable_gravatar, Defaults.enable_gravatar),
+                prefs.LOGDATE: (self.logdate, Defaults.logdate),
                 prefs.MERGE_DIFFSTAT: (self.merge_diffstat, Defaults.merge_diffstat),
                 prefs.MERGE_SUMMARY: (self.merge_summary, Defaults.merge_summary),
                 prefs.MERGE_VERBOSITY: (self.merge_verbosity, Defaults.merge_verbosity),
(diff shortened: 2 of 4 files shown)


cola/widgets/prefs.py
```diff
@@ -140,6 +142,12 @@ def __init__(self, context, model, parent, source):
         self.show_path = qtutils.checkbox(checked=True)
         self.tabwidth = standard.SpinBox(value=8, maxi=42)
         self.textwidth = standard.SpinBox(value=72, maxi=150)
+        self.logdate = QtWidgets.QLineEdit()
+        tooltip = N_('Similar to git log --date option')
+        self.logdate.setToolTip(tooltip)
+        regex = QtCore.QRegExp(logdate.create_regex())
+        validator = QtGui.QRegExpValidator(regex, self.logdate)
+        self.logdate.setValidator(validator)
```
davvid, line 150: If the setup for the combobox gets too long (it's already longer than the rest) then that should probably be put into a separate `def build_date_formats_combo(parent)` helper function.

docs/git-cola.rst
```diff
@@ -762,6 +762,12 @@ Defaults to `true`.  This setting is configured using the `Preferences`
 dialog, but it can be toggled for one-off usage using the commit message
 editor's options sub-menu.
 
+cola.logdate
+--------------
```
davvid, line 766: Make sure that the `---------` line is the exact same length as `cola.logdate` otherwise sphinx will complain about it.

docs/git-cola.rst
```diff
@@ -762,6 +762,12 @@ Defaults to `true`.  This setting is configured using the `Preferences`
...
+Set the default date-time mode for the DAG display. Setting a value for
+cola.logdate is similar to using git log's --date option.
+See `git log(1) <https://www.git-scm.com/docs/git-log>`_ for details.
+
```
davvid, line 770: We should use this URL (with the anchor value) so that we link directly to the `--date=<format>`  docs.

`https://git-scm.com/docs/git-log#Documentation/git-log.txt---dateltformatgt`

https://git-scm.com/docs/git-log#Documentation/git-log.txt---dateltformatgt


Also, instead of "Similar", maybe it should say, "This value is passed to `git log --date=<format>`"?

davvid (changes requested):
Nicely done. Please squash these sugs into your original commit and force-push over your review branch.

davvid:
Kindly adjust the commit to use your full name. There's a short note in `CONTRIBUTING.md` about this:

> Also notice that a real name is used in the Signed-off-by: line. Please don't hide your real name.

DasaniT:
Thanks for your feedback! I've made the requested changes, and I made some other small improvements as well.

About the combo, actually at first I did use `qtutils.combo()` and then, as you mentioned, I changed it to text box just because of the `format` option. Right now, the combo box is not editable, because I think that's not necessary. As a result, I removed the validator as well.

cola/widgets/prefs.py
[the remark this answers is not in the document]
DasaniT: About the combo, actually at first I did use `qtutils.combo()` and then, as you mentioned, I changed it to text box just because of the `format:...` option. After the changes, the combo box is not editable, because I think that's not necessary. As a result, I removed the validator as well.

cola/logdate.py
```diff
@@ -0,0 +1,35 @@
...
+        Option.LOCAL,
+        Option.ISO,
+        Option.ISO_STRICT,
+        Option.RFC,
+        Option.SHORT,
+        Option.RAW,
+        Option.HUMAN,
+        Option.UNIX,
+        Option.FORMAT,
+    ]
```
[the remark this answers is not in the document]
DasaniT: I changed "iso" and "iso-strict" to "iso8601" and "iso8601-strict" too.

cola/models/dag.py
```diff
@@ -273,7 +273,12 @@ def get(self):
 
         self.reset()
         ref_args = utils.shell_split(self.params.ref)
-        cmd = self._cmd + ['-%d' % self.params.count] + ref_args
+        cmd = (
+            self._cmd
+            + ['-%d' % self.params.count]
+            + ref_args
+            + ['--date=%s' % self.context.cfg.get('cola.logdate')]
```
davvid, line 280: sorry for not spotting this earlier (I saw it but forgot to note it).

When we call `get()` please also specify `default=prefs.DateFormat.DEFAULT` and then move the entirety of the `logdate.py` module into `cola/models/prefs.py`. That way we don't have to add a new module just for this small bit of config.

davvid: If we don't specify `default=...` then it'll error out when the user doesn't have `cola.logdate` configured.

docs/git-cola.rst
```diff
@@ -762,6 +762,13 @@ Defaults to `true`.  This setting is configured using the `Preferences`
 dialog, but it can be toggled for one-off usage using the commit message
 editor's options sub-menu.
 
+cola.logdate
+------------
+Set the default date-time mode for the DAG display. This value is
+passed to `git log --date=<format>`. See `git log(1) <https://git-scm
+.com/docs/git-log#Documentation/git-log.txt---dateltformatgt>`_ for details.
```
davvid, line 769: I'd suggest moving the "See ..." line to another line so that we don't have to split the URL across two lines. I'm not sure it'll do the right thing with the newline before the `.com`.

cola/models/dag.py
DasaniT: I've added a function `logdate(context)' in `cola/models/prefs.py` and I'm calling the function instead of `self.context.cfg.get('cola.logdate'). Is that ok?

And one more thing, I made the change but I would prefer having the configs for logdate in separate file. Because  having them in `cola/models/prefs.py` would make this file messy. There are similar config files like `cola/hidpi.py`.

davvid:
Very nicely done, thanks!

PR #1325, opened 2023-08-08
Tried to add a part of notification

Ra00f1:
So I tried to add the notification a couple of days ago and this is what I could do.

I did add the notification like was requested and only works if the library used is installed and if not won't create any errors and tried to add it to interactions as well however to be honest I don't completely understand how that file ties into the other parts of the code so I just left it empty for now.

I'll still look into it and try to understand more of the code but if you have any tips that could help I'd gladly take it :)


davvid:
Closing since no response. It looks like #1404 might be on a closer path in case you want to snipe.

PR #1328, opened 2023-08-16
Prevent double-hyphen in git-dag doc

BaumiCoder:
like it was done in commit 85d7fdb by @kurtmckee for git-cola.rst

---

The double-hyphen at the [docs](https://git-cola.readthedocs.io/en/latest/git-dag.html#log-options)  or in the manpage `man git dag`  can easily mixed up with a single hyphen. At first I used a single hyphen and was confused about the error message I got.

### DIFF
--- a/docs/git-dag.rst
+++ b/docs/git-dag.rst
@@ -15,21 +15,21 @@ graphical interface.
 OPTIONS
 =======
 
---prompt
---------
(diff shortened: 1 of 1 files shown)


kurtmckee (approved):
Looks good to me!

PR #1332, opened 2023-08-20
Fix tests on Windows

### LINKED ISSUE #1331: Test suite fails on Windows
The test suite currently has many failures and errors when running on Windows.

For example, `test\git_test.py:test_find_git_dir_never_found` demonstrates hard-coded assumptions about paths:

```
>               raise AssertionError(
                    f'{problem}\n'
                    f'Expected: {_CallList(calls)}'
                    f'{self._calls_repr(prefix="Actual").rstrip(".")}'
                ) from cause
E               AssertionError: Calls not found.
E               Expected: [(('/does/not/exist',), {}),
E                (('/does/not/exist/.git',), {}),
E                (('/does/not',), {}),
E                (('/does/not/.git',), {}),
E                (('/does',), {}),
E                (('/does/.git',), {}),
E                (('/',), {}),
E                (('/.git',), {})]
E               Actual: [call('C:\\does\\not\\exist'),
E                call('C:\\does\\not\\exist\\.git'),
E                call('C:\\does\\not'),
E                call('C:\\does\\not\\.git'),
E                call('C:\\does'),
E                call('C:\\does\\.git'),
E                call('C:\\'),
E                call('C:\\.git')]
```

There are other types of failures on Windows, including hard-coded assumptions about newlines (`\n` vs `\r\n`), and many, many `os.unlink()` calls that raise `PermissionError`.

I'll take a crack at resolving these and submit a PR (or PRs, if possible, to keep reviewing easy).

kurtmckee:
This PR introduces the following:

* Fix a `GIT_CEILING_DIRECTORIES` parsing bug (Windows uses semicolons as a delimiter, not colons)
* Fix a bug in the `shutil.rmtree()` recovery function caused by an incorrect function comparison
* Resolve all path-related failures
* Resolves all newline-related failures

The design of the test suite currently uses `make` as the authoritative test runner; `tox` simply invokes `make`. This is incompatible with Windows platforms; `tox` needs to be the authoritative test runner which `make` invokes in order for tests to run cross-platform reliably. This redesign is not addressed in this PR.

Fixes #1331

### DIFF
--- a/cola/git.py
+++ b/cola/git.py
@@ -15,6 +15,7 @@ from .compat import ustr
 from .compat import WIN32
 from .decorators import memoize
 from .interaction import Interaction
+from .utils import is_win32
 
 
 GIT_COLA_TRACE = core.getenv('GIT_COLA_TRACE', '')
@@ -111,9 +112,7 @@ class Paths(object):
         self.common_dir = common_dir
 
     def get(self, path):
-        separator = ":"
-        if sys.platform == "win32":
-            separator = ";"
+        separator = ';' if is_win32() else ':'
         ceiling_dirs = set()
         ceiling = core.getenv('GIT_CEILING_DIRECTORIES')
         if ceiling:
--- a/cola/git.py
+++ b/cola/git.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import sys
 from functools import partial
 import errno
 import os
@@ -15,7 +14,6 @@ from .compat import ustr
 from .compat import WIN32
 from .decorators import memoize
 from .interaction import Interaction
-from .utils import is_win32
 
 
 GIT_COLA_TRACE = core.getenv('GIT_COLA_TRACE', '')
@@ -112,11 +110,10 @@ class Paths(object):
         self.common_dir = common_dir
 
     def get(self, path):
-        separator = ';' if is_win32() else ':'
         ceiling_dirs = set()
         ceiling = core.getenv('GIT_CEILING_DIRECTORIES')
         if ceiling:
-            ceiling_dirs.update([x for x in ceiling.split(separator) if x])
+            ceiling_dirs.update([x for x in ceiling.split(os.pathsep) if x])
 
         if path:
             path = core.abspath(path)
--- a/cola/git.py
+++ b/cola/git.py
@@ -1,4 +1,6 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
+
+import sys
 from functools import partial
 import errno
 import os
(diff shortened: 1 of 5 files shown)


cola/git.py
```diff
@@ -109,10 +111,13 @@ def __init__(self, git_dir=None, git_file=None, worktree=None, common_dir=None):
         self.common_dir = common_dir
 
     def get(self, path):
+        separator = ":"
+        if sys.platform == "win32":
+            separator = ";"
```
davvid, line 116: Thanks for getting these fixed up. We have a `utils.is_win32()` helper function that checks a few other predicates, and is better for grepping. Please use that function whenever special-casing windows.

TIny sug, but please prefer `'` single-quotes for strings.

todo: I should probably update our `garden fmt` code formatting setup to use `blonde` instead of  `black` so that we can enforce the preference for single-quotes.

davvid (changes requested):

cola/git.py
kurtmckee: If code linting (and style enforcement) could come in the form of a pre-commit hook then it will happen automatically.

If you'd like, I'll create a separate PR to introduce a preliminary pre-commit config for trailing whitespace and end-of-file newlines. You can then extend that to use black or isort or blonde or whatever other tools are desirable. 😁

kurtmckee: I think I've incorporated your feedback correctly; if you spot anything else please let me know!

living180: Even better than using `utils.is_win32()` to switch between separators would be to use Python's built-in `os.pathsep` variable which is defined to be the correct separator depending on the platform.

test/git_test.py
```diff
@@ -2,8 +2,12 @@
 # pylint: disable=redefined-outer-name
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import pathlib
+import sys
```
davvid, line 6: Almost there! It looks like the CI checks caught this unused `import sys` and the one in `cola/git.py` as well.

Thanks for taking these on.

kurtmckee: > unused import

Ah, yep, I missed that. A pre-commit hook for flake8 can catch that, too. I'll introduce brunette, flake8, and text standardization (end-of-file newlines and trailing whitespace) in a separate PR.

cola/git.py
kurtmckee: @living180 Great callout! I'll switch to that.

kurtmckee: I've add a commit that switches to `os.pathsep`. Thanks for pointing this out, @living180!

test/git_test.py
kurtmckee: Jiminy Christmas, I need pre-commit hooks to develop effectively!

I've fixed another issue caught by pylint in CI. Probably the pre-commit hook needs to be pylint, not flake8.

kurtmckee: Update: #1333 introduces a [pre-commit](https://pre-commit.com/) config so style can be checked and enforced prior to allowing developers to even commit their changes. Brunette and pylint are part of that config. :grin:

PR #1333, opened 2023-08-21
Use pre-commit

kurtmckee:
Add a pre-commit configuration with the following style enforcements:

* Ends-of-file newlines are enforced
* Trailing whitespace is stripped
* Python 3.6 idioms are enforced. This includes (among other things)

    * Removal of useless `__future__` imports
    * Removal of Python 2-compatible class inheritance from `object`
    * Use of set literals
    * Use of f-strings
    * Simpler `super()` calls

* Brunette style rules

In addition, pylint has been added a pre-commit hook. It catches errors but isn't able to auto-fix issues that it finds.

----

If this merges, I _strongly_ recommend enabling [pre-commit.ci](https://pre-commit.ci) for this repository. The pre-commit CI service will keep the pre-commit hook versions up-to-date via automatic PRs, and will add itself as an additional CI check for incoming PRs.

Fixes introduced by the pre-commit hooks will be committed and added to incoming PRs to enforce style rules. No discussion needed! :+1: 

Here's an example of what that looks like from a PR in one of my own repos:

image


cola/decorators.py
```diff
@@ -77,7 +76,7 @@ def interruptable(func, *args, **opts):
     while True:
         try:
             result = func(*args, **opts)
-        except (IOError, OSError) as e:
+        except OSError as e:
```
davvid, line 79: Nice! I didn't know they improved that in python3 (common OSError base class).

cola/themes.py
```diff
@@ -109,371 +108,371 @@ def style_sheet_flat(self):
...
+            field=field,
+            button=color_rgb,
+            darker=darker,
+            lighter=lighter,
+            grayed=grayed,
+            button_text=button_text,
+            field_text=field_text,
+            separator=defs.separator,
+            focus=focus,
+        )
```
davvid, line 475: this is ok, but the stylesheets are arguably nicer without the f-string hammer due to the wide use of `{ }` in qss.

davvid:
I'll try enrolling in pre-commit.ci tonight.

PR #1335, opened 2023-08-22
Disable pylint when running in pre-commit CI

kurtmckee:
`main` is showing pre-commit.ci failures when running the `pylint` hook.

I didn't realize it before, but pylint's documentation states that [pylint can only be used in local pre-commit runs](https://pylint.pycqa.org/en/latest/user_guide/installation/pre-commit-integration.html).

This PR disables pylint in pre-commit.ci but allows it to continue running locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,12 @@
 # Run `pre-commit autoupdate` to update tool versions.
 
+ci:
+  skip:
+    # pylint's pre-commit documentation states it only works when running locally.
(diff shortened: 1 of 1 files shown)


davvid:
It's interesting.. we seem to be able to run pylint in `.github/workflows/ci.yml` (the `garden check` call) but it's all good. I'll take it. Thanks for the fixup

kurtmckee:
I think the discrepancy is how pre-commit sets up the environment that pylint runs in. Good that pylint is still running in CI, though, and good that it's able to run locally!

PR #1336, opened 2023-08-22
starting with sphinx 7.2.0, root is a pathlib.Path not a string anymore

bearsh:
so convert it to a string, the trailing os.sep is not needed. os.path.join which is used to join paths, expects string and inserts the os.sep.

See-also: sphinx-doc/sphinx#11526
See-also: sphinx-doc/sphinx#11608

### DIFF
--- a/extras/sphinxtogithub/sphinxtogithub.py
+++ b/extras/sphinxtogithub/sphinxtogithub.py
@@ -97,7 +97,7 @@ class DirectoryHandler:
     def __init__(self, name, root, renamer):
         self.name = name
         self.new_name = name[1:]
-        self.root = root + os.sep
+        self.root = str(root)
         self.renamer = renamer
 
     def path(self):


davvid:
Thanks! I had just recently pinned sphinx to an older verison because I was running into this bug.

I'll unpin after merging this.

TODO:  we should fix this in the upstream sphinxtogithub ~ https://github.com/michaeljones/sphinx-to-github that way it's not just cola that'll benefit from your fix.

davvid:
Nevermind on that TODO -- the upstream project seems to have archived their repo so that it's read-only now.

I pushed a [similar commit](https://github.com/michaeljones/sphinx-to-github/commit/5915ab289656cf342a873fdd5d3f6ea735d39a91) to my fork ~ https://github.com/davvid/sphinx-to-github

Despite the README saying that it's obsolete, it's still useful when generating docs for a subdirectory of a jekyll site, which is how I'm using it. Thanks for this fix!

extras/sphinxtogithub/sphinxtogithub.py
```diff
@@ -97,7 +97,7 @@ class DirectoryHandler:
     def __init__(self, name, root, renamer):
         self.name = name
         self.new_name = name[1:]
-        self.root = root + os.sep
+        self.root = str(root)
```
davvid, line 100: I fixed this up post-merge to use `str(root) + os.sep` because the code outside of the constructor does things like `path = directory.replace(self.root, "", 1)` (Cf. `def relative_path()`) which currently relies on the trailing `/` being there so that it replaces all the way through the `/`.

It doesn't seem to make a difference in cola's usage, but it doesn't hurt to fix this up in case we end up going down that code path in the future so I tweaked this to keep `+ os.sep` at the end.

PR #1337, opened 2023-08-23
GUI: Stop making newly-created buttons immediately visible

### LINKED ISSUE #1329: A small window flashes rapidly while opening a repo
I have a screen recording of the issue: https://youtu.be/6Wxs39Up7E8

Git Cola version 4.2.1 
- Windows-10-10.0.19041-SP0 
- Python (C:\Users\Owner\AppData\Local\Programs\git-cola\Python\pythonw.exe) 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 18:58:18) [MSC v.1900 64 bit (AMD64)] 
- Git 2.40.1.windows.1 
- Qt 5.14.2 
- QtPy 2.0.1 
- PyQt5 5.14.2 

I looked up a solution for PyQt and found this possible solution:

```Python
mainWidget.setUpdatesEnabled(False)
# do a bunch of operations that would trigger expensive events
# like repaints
mainWidget.setUpdatesEnabled(True)
```
-from <https://stackoverflow.com/a/13781568/4541104>

I realize you use Qt but I had the same problem (tiny window flashing while adding widgets) with Tk on Windows and solved it by hiding the window while adding the widgets. So, if using the `setUpdatesEnabled` calls don't work, maybe hiding the window will work.

kurtmckee:
On Windows, making a newly-created button visible causes a new window to open containing only the new button.

Testing shows that the buttons will be visible in whatever dock they are ultimately assigned to when the main git-cola window opens even if `.setVisible(True)` is not called first.

Therefore, the call to `.setVisible()` seems to be unnecessary.

Fixes #1329

### DIFF
--- a/cola/qtutils.py
+++ b/cola/qtutils.py
@@ -736,7 +736,7 @@ def tool_button():
     return button
 
 
-def create_action_button(tooltip=None, icon=None, visible=True):
+def create_action_button(tooltip=None, icon=None, visible=None):
     """Create a small toolbutton for use in dock title widgets"""
     button = tool_button()
     if tooltip is not None:
@@ -744,7 +744,8 @@ def create_action_button(tooltip=None, icon=None, visible=True):
     if icon is not None:
         button.setIcon(icon)
         button.setIconSize(QtCore.QSize(defs.small_icon, defs.small_icon))
-    button.setVisible(visible)
+    if visible is not None:
+        button.setVisible(visible)
     return button
 
 


davvid:
Thanks for narrowing this down.

I wonder if the real core issue is that these buttons should have a Qt widget parent passed in so that they don't temporarily create their own window. In other words, it seems like maybe we should wire a `parent` parameter into the creator functions so that the buttons always have a valid parent from beginning to end.

Currently they'll only have a valid parent after they get added to a layout, which happens after creation.

I'll have to test this out a bit. The `setVisible(...)` call is really only there for the `False` scenario, so maybe another approach might be to only call setVisible when the value is False. I have to double-check that for the exit-diff-mode button.

A short-term approach might be to keep the setVisible call but only handle the False scenario. The default behavior won't need to call setVisible(True) as you already noted. If we did that then I think this should be okay to merge as a first step and then the `parent` approach can be done afterwards.

What do you think?

kurtmckee:
@davvid I don't have a lot of experience with GUI programming, though I'm aware of the concept of assigning a widget to a parent. I'll have to defer to your knowledge here.

Does visibility have three possible values, like `True`, `False`, and `inherit`? I had assumed that it defaulted to `False`, but now that you've pointed out that only the `False` case is truly important, I'm wondering if there are 3 values.

I'll try another change using a conditional along the lines you suggested, like:

```python
def create_action_button(..., visible=None):
    button = Button()
    if visible is not None:
        button.setVisible(visible)
```

I'll report back when I've tested that, but please also test whether this PR works for the exit-diff-mode button that you mentioned!

davvid:
> Does visibility have three possible values, like `True`, `False`, and `inherit`? I had assumed that it defaulted to `False`, but now that you've pointed out that only the `False` case is truly important, I'm wondering if there are 3 values.

I think we only need to consider True and False. Since `True` is the default then the logic might be something along the lines of...


```
if not visible:
    button.setVisible(visible)
```


> I'll report back when I've tested that, but please also test whether this PR works for the exit-diff-mode button that you mentioned!

Sounds good. I tested this out -- I think we need to handle the `False` scenario as mentioned above. The "exit diff mode" button does actually appear without handling False, but it gets papered over by the StatusWidget.set_mode() logic which goes and hides it after-the-fact.

So we need to handle False so that the button is correctly setup when used in other scenarios.

I've been gradually trying to get a Windows VM up and running so that I can test this stuff myself but it's been slow-coming.

Later on if I take a stab at ensuring that all the widgets have parents I can push that up to a temp branch and maybe you can help me test it out? Thanks again ~ once you have the updates for False ready I'll merge that in right away.

I have a planned v4.3.2 patch release for next week so I'm probably going to cherry-pick your updates into the stable branch as well so that it's included in the next patch release.

kurtmckee:
> `True` is the default [visibility]

I used pdb to call `.isVisible()` on a newly-instantiated button and it returned `False`. On Windows, I think that the default is `False`. This might be a platform difference, so I'm hopeful you can check this in your environment, too.

> The "exit diff mode" button

I cannot find this button in the interface, but I noticed that its icon ("circle_slash_red") sometimes briefly flashed onscreen during testing. It was the only icon that showed up in color, which is what caught my attention. Here's a screenshot of the Status dock in that I think that button is supposed to show up in. This screenshot was taken using git-cola 4.2.1, installed on my computer.

image

Also, during testing I noticed that a button that I took a screenshot of and posted to #1329 didn't appear on my screen. I discovered that it's hidden and doesn't appear to work, though its tooltip still pops up (the screenshot doesn't show my mouse pointer, which was hovering to the right of the Add button). This screenshot was also taken using git-cola 4.2.1, installed on my computer.

image

Would you double-check whether the exit-diff-mode button (and the Favorites delete button) are visible on Linux using currently-released versions of git-cola, or in `main`? I'm curious if you're seeing the same behavior I'm seeing, or if this is a platform-specific difference.

I'll incorporate the feedback re: setting visibility to `False` but I'm unsure that it'll resolve the exit-diff-mode issue you're seeing.

kurtmckee:
My apologies, I finally realized that I had to actually _enter_ diff-mode for the icon to show up. I also finally realized that the exit-diff-mode button was flashing when I ran the code in this branch, even though the value returned by `.isVisible()` is already `False`; setting it explicitly to `False` prevented this brief flashing.

I've incorporated your feedback so the call to `.setVisible()` is guarded by a condition and have force-pushed so the git history is clean.

davvid:
Thanks, after thinking about it I like how you're handling it (tri-state, and opt-in by the caller) better than my initial reaction to make it terse and default to True.

I'll merge this in shortly, this does indeed seem to fix the issue without needing to dive into the parenting stuff just yet as far as I can tell. (I finally managed to get a temporary VM running).

davvid:
👏 PR 1337

kurtmckee:
Haha, didn't even notice this PR had such a leet number! Lucky! 😉

kurtmckee:
@davvid I just realized that I didn't add a note to the CHANGELOG to note this bug fix for Windows. Is that something I should submit a PR for, or do you want to add a note for that?

davvid:
Thanks for the reminder ~ I'll add this to the changelog shortly.

PR #1342, opened 2023-08-31
allow smaller font size for hidpi

sthalik:
Allow using smaller fonts. This can be useful for large HiDPI screens. For example, my Windows machine's screen's pixel density is 105 and the display scaling is set to 150% in 2160p. This makes the size of a 8 pt font too large. It can also vary depending on the Qt backend being used (DirectDraw vs xcb, etc.).

Perhaps there's some sense in lowering it even a bit further but that's enough changes for one day.

As usual, thank you for this wonderful software. The ability to edit diffs before reverting will surely come in handy.


davvid:
Thanks!

PR #1345, opened 2023-09-05
Resolve causes of CI failures

kurtmckee:
This PR should resolve issues caught by pylint and pre-commit.

If the tests against this PR pass, then the issues are resolved. 😀

### DIFF
--- a/cola/widgets/commitmsg.py
+++ b/cola/widgets/commitmsg.py
@@ -487,7 +487,7 @@ class CommitMessageEditor(QtWidgets.QFrame):
         )
         self.context.runtask.start(
             task,
-            finish=lambda commit_task: self.commit_finished.emit(commit_task),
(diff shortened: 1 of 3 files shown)


cola/widgets/commitmsg.py
```diff
@@ -487,7 +487,7 @@ def commit(self):
         )
         self.context.runtask.start(
             task,
-            finish=lambda commit_task: self.commit_finished.emit(commit_task),
+            finish=self.commit_finished.emit,
```
davvid, line 490: Thanks for these fixups! The `commit_finished` signal originally had a different signature but then it became the same and I missed the fact that I could drop the lambda.

kurtmckee: pylint did great work in this regard. I'm honestly astounded that it found this!

PR #1349, opened 2023-09-20
remove translation of git jargon from Czech, Spanish and French

OriShalhon:
changes were made based on [1343](https://github.com/git-cola/git-cola/pull/1343)


davvid:
Thanks!

Issue #1160, opened 2021-09-08
language files should not translate git commands


prof7bit:
git commands like "push", "pull", "commit", "checkout", etc should not be translated into other languages in the user interface. It is the main reason why people are trying to run it in the untranslated version because it completely breaks the mental connection between what I am trying to do with git and how it is called in the localized translation.

I am looking for "push" and then I have to try to reverse the twisted mind of the translator until I find a menu entry that reads "Veröffentlichen" (German) and would actually translate to "publish" which is something entirely different, and there is no way to find out which of these strange commands does what I want, other than clicking it and trying to interpret from the symptoms what it actually did!

please remove all these misleading strings from all translation files.

pilarmolinalopez:
Hello. I think that makes a lot of sense. In fact, if I'm not mistaken, the Spanish translations do not translate those type of strings (strings that correspond to git commands).

atharvapatil123:
Hi @prof7bit, Is the issue still open? If yes, I'd like to take up on it.

jalajk24:
@prof7bit i would like to contribute in this issue can you assign it to me and drop some insight?

davvid:
Thanks @jalajk24 for stepping up. I updated the translation files so that the metadata is up to date.

All of the translations can be found in `cola/i18n/*.po`, for example `cola/i18n/fr.po` contains the French translations.

You should be able to edit the .po files directly and cola will pickup the new translations (there's no "build" step -- the files are consumed as-is).

You can test other languages by launching with a custom $LANG environment variable, eg.:

```
env LANG=fr_FR.UTF-8 ./bin/git-cola
```

davvid:
I took a stab at some of the easy German ones over in https://github.com/git-cola/git-cola/pull/1343

Let me know what you think @prof7bit.  It's not everything but it should be better than what's currently there.

OriShalhon:
where does this issue stand? could I contribute to it?

davvid:
I made some tweaks but pull requests are always accepted. The translation files are located at `cola/i18n/*.po`. This issue should probably be closed by adding a todo note for translators in `CONTRIBUTING.md` noting the preference for leaving git command jargon untranslated.

OriShalhon:
Based on [1343](https://github.com/git-cola/git-cola/pull/1343) I made [1349](https://github.com/git-cola/git-cola/pull/1349) changing Czech, Spanish and French- will continue to additional languages if accepted

OriShalhon:
[...]

PR #1351, opened 2023-09-24
avoid translating git jargon - add test to verify pull anbd push are …

OriShalhon:
remove git jargon translation.
remove test for French - pull and push.
added new test to verify no language translate basic git jargon.

### DIFF
--- a/test/i18n_test.py
+++ b/test/i18n_test.py
@@ -54,10 +54,26 @@ def test_translates_random_english():
     actual = N_('Random')
     assert expect == actual
 
+
 def test_avoid_translation_git_jargon():
-    languages = ['cs_CS', 'de_DE', 'es_ES', 'fr_FR', 'hu_HU', 'id_ID', 'it_IT',
-                 'ja_JP', 'pl_PL', 'pt_BR', 'ru_RU', 'sv_SV', 'tr_TR', 'uk_UK',
-                 'zh_CN', 'zh_TW']
+    languages = [
+        'cs_CS',
+        'de_DE',
+        'es_ES',
+        'fr_FR',
+        'hu_HU',
+        'id_ID',
+        'it_IT',
+        'ja_JP',
+        'pl_PL',
+        'pt_BR',
+        'ru_RU',
+        'sv_SV',
+        'tr_TR',
+        'uk_UK',
+        'zh_CN',
+        'zh_TW',
+    ]
 
     expected_pull = 'Pull'
     expected_push = 'Push'
@@ -65,7 +81,6 @@ def test_avoid_translation_git_jargon():
     failed_languages_push = []
 
     for language in languages:
-
         i18n.install(language)
         if not expected_pull == N_(expected_pull):
             failed_languages_pull.append(language)
@@ -76,12 +91,13 @@ def test_avoid_translation_git_jargon():
         i18n.uninstall()
 
     if failed_languages_pull:
-        print("\nFailed languages pull:", failed_languages_pull)
+        print('\nFailed languages pull:', failed_languages_pull)
     if failed_languages_push:
-        print("Failed languages push:", failed_languages_push)
+        print('Failed languages push:', failed_languages_push)
 
     assert failed_languages_push == []
     assert failed_languages_push == []
+
 
 def test_get_filename_for_locale():
     """Ensure that the appropriate .po files are found"""
--- a/test/i18n_test.py
+++ b/test/i18n_test.py
@@ -95,8 +95,8 @@ def test_avoid_translation_git_jargon():
     if failed_languages_push:
         print('Failed languages push:', failed_languages_push)
 
-    assert failed_languages_push == []
-    assert failed_languages_push == []
+    assert not failed_languages_push
+    assert not failed_languages_pull
 
 
 def test_get_filename_for_locale():
--- a/test/i18n_test.py
+++ b/test/i18n_test.py
@@ -54,17 +54,34 @@ def test_translates_random_english():
     actual = N_('Random')
     assert expect == actual
 
+def test_avoid_translation_git_jargon():
+    languages = ['cs_CS', 'de_DE', 'es_ES', 'fr_FR', 'hu_HU', 'id_ID', 'it_IT',
+                 'ja_JP', 'pl_PL', 'pt_BR', 'ru_RU', 'sv_SV', 'tr_TR', 'uk_UK',
+                 'zh_CN', 'zh_TW']
 
-def test_translate_push_pull_french():
-    i18n.install('fr_FR')
-    expect = 'Pull'
-    actual = N_('Pull')
-    assert expect == actual
+    expected_pull = 'Pull'
+    expected_push = 'Push'
+    failed_languages_pull = []
+    failed_languages_push = []
 
-    expect = 'Push'
-    actual = N_('Push')
-    assert expect == actual
+    for language in languages:
 
+        i18n.install(language)
+        if not expected_pull == N_(expected_pull):
+            failed_languages_pull.append(language)
+
+        if not expected_push == N_(expected_push):
+            failed_languages_push.append(language)
+
+        i18n.uninstall()
+
+    if failed_languages_pull:
+        print("\nFailed languages pull:", failed_languages_pull)
+    if failed_languages_push:
+        print("Failed languages push:", failed_languages_push)
+
+    assert failed_languages_push == []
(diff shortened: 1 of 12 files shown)


test/i18n_test.py
```diff
@@ -55,11 +55,48 @@ def test_translates_random_english():
     assert expect == actual
 
 
-def test_translate_to_french():
-    i18n.install('fr_FR')
-    expect = ' commits avant'
-    actual = N_(' commits ago')
-    assert expect == actual
+def test_avoid_translation_git_jargon():
```
living180, line 58: If these strings should not be translated, wouldn't it be better to update the code to stop marking them as translatable, rather than testing that no translation is provided?

OriShalhon: I agree with you, I wanted it to remain consistent with previous commit [1343](https://github.com/git-cola/git-cola/pull/1343/commits) with similar changes in German.

living180:
And an overall comment:  I don't know if there was a specific decision made regarding translation of these git actions within git-cola, but as a native English speaker who has spent a number of years learning another language, I think changing the text of these buttons and menu items from the translated versions back to English would make git-cola less accessible to non-English speakers.  I just looked at a couple of the translations of the [Git Book](https://git-scm.com/book/en/v2) that I know a little bit, and at least what I saw they translated the git terminology into those languages, unless it was the actual text of a git command.  And since users of git-cola are not trying to use the command line, I think it would be beneficial for them to have non-English terms in the user interface whenever it is reasonably possible.

living180:
Sorry, hit the Close button by accident.

OriShalhon:
@living180 I understand what you're saying and agree,
4 days ago, I created a pull-request that was accepted with Csech, French and Spanish based on previously sent PR 1343
please see [here ](https://github.com/git-cola/git-cola/issues/1160)the discussion, as the previous PR was accepted, I moved on with removing the translation of remaining languages and adding the test. 
I agree that a more robust solution is possible and could work on implementing it if needed.

living180:
@OriShalhon thanks for the further context linked in the #1343 discussion.

davvid:
Yeah, I would generally lean towards, "do the same as core git" so that we're at least consistent. Perhaps we went too far with some of these tweaks?

```bash
./bin/git-dag v4.3.2.. -- cola/i18n
```


I think the "Keep Index" translations that were removed (by me in an earlier commit) can be restored right with this guidance in mind. I'll follow up with those if folks agree.

It does get fuzzy for me when the button says "Push" for a UI action that represents `git push`.. in those cases I've leaned towards not-translating so that users get to gradually familiarize themselves with the command-line interface. @living180 do you think that's a reasonable take on the one-word action buttons -- that we can leave those untranslated, or do we think there's some more nuance and perhaps language-specific?

What do you think about "stash"? Should the word "stash" get translated or should it always be referred to as Stash the concept and command? My tweaks to the translations modified a sentence that referred to a stash, so I'm all ears on whether that went too far and should be restored as well.

(I should probably read the Git Book to see what they're doing for Stash)

@OriShalhon @living180 @prof7bit thoughts? Luckily none of this is in a tagged release yet so I'm keen to get consensus before the upcoming release and adjust accordingly.

OriShalhon:
@davvid I agree with a lot of what you say.

Generally I think the best solution will be to decide upon a set of most used git commands - pull, push, branch, blame ETC. and come up with a solution as part of the code to prevent the translation of these specific commands.

In regards to tralsation of a single word in a sentence I think it is best if we also include a note in CONTRIBUTING.MD where we ask to put under parentheses the translated git command - this in turn will result in the change being gradual as it requires native speakers of each language to make the changes in contrast to the bulk changes in my commits.

living180:
Here's some further thoughts, though again I will defer to translators as I am a native English speaker who only uses git-cola in English:

The original issue that spurred these changes (#1160) was due a user not being able to make a connection between the German word used (Veröffentlichen) and the equivalent git operation (push).  And indeed while the Git Book uses that word in the section on [pushing to remotes](https://git-scm.com/book/de/v2/Git-Branching-Remote-Branches) it does not appear at all in the [German translation strings](https://github.com/git/git/blob/master/po/de.po) for git itself.  It appears that for those translation strings, sometimes the term "push" is left untranslated, and other time alternate terminology is used, but as I don't speak any German, I am not able to identify a specific word that might be appropriate.

Contrast this with Spanish, where in the Spanish translation strings for git, "push" is fairly consistently translated as the verb "empujar".  In my mind, using "Empujar" as a button label in Spanish for "Push" might make sense, but it really depends on the user's level of familiarity with English and the git-specific English terminology.  And not every language may have a readily-agreed-upon term for every English git term, hence the frustration by the German user not immediately understanding what was meant by "Veröffentlichen".  And for some languages, even if there is a clearly corresponding word in that language, users may still tend to use the English terms as loanwords.

Thus I don't know that there is a single right answer for all languages, and I would tend to leave it up to the discretion of the translators, whether they translate a term like "Push" or not.  If I user reports a problem with a translation, as was done in #1160, ideally if the translator still available, the issue could be brought to their attention, but otherwise it would just need to be handled on a case-by-case basis.

I also wonder for the case where git terminology is translated, if it might be possible to have a sort of hybrid solution where the original English term is displayed in a tooltip, but I worry about making the interface feel cluttered.

OriShalhon:
@living180 @davvid 
maybe we should consider it to be up to the user, add the ability via a checkbox or configuration to determine whether the commands should be translated or not.

This in turn requires additional development but could solve the issue.

OriShalhon:
@davvid any thoughts?

davvid:
Sorry for taking a bit to respond and make a decision on this topic. After a lot of consideration here's what I've done. Sorry for leading you astray @OriShalhon since this is a change in direction.

I've basically reverted most of the recent removals of translations. I think allowing them to be removed was a bad decision by my part and even the idea of not-translating single-word action buttons isn't really a good general guidance in the long-term.

I did make some very small changes to the translations to avoid translating *very specific* uses, specifically the Push and Pull labels for the German translations only. This is not a general rule -- every language is different.

Allow translators to apply their best judgement and language-specific preferences is the better long-term position in my opinion. For example, I talked to a native French speaker and they clarified that avoiding translation for the word "stash" specifically was probably a good idea, but that the rest of the translations were just fine.

I also added tooltips  in a few places. I did consider that this could make the UI a little cluttered, but I also think the tooltips are an improvement overall.

For example, the "Push" button in the Actions widget and the main menu now has a tooltip that says, `Push changes to one or more remotes using "git push"`. That tooltip serves two purposes. It allows users to gradually familiarize themselves with the CLI and it also disambiguates any confusing translations that we may be using in languages/contexts that I do not speak.

The first error was when I over-reacted from the original report and removed more translations than I should have. The rest of the PRs were just following along, sorry 'bout that.

I don't really think there's much else to change about the translations at this point beyond adding more of them (rather than removing translations) and possibly adding an additional tooltip here and there after hearing feedback from users that rely on them.

Regarding additional development to disable translations -- we already have that feature with multiple ways of opting-in to a different language. We support the Standard `$LANG` environment variable for replacing the language and we also support a `~/.config/git-cola/language` file that can be used to override the language:
https://git-cola.readthedocs.io/en/latest/git-cola.html#language-settings

The only additional feature I'd be in favor of adding related to language-overrides would be to support a new `$GIT_COLA_LANG` environment variable that'd have higher priority than `$LANG` so that users have a way of targeting just Git Cola with that setting. `$LANG` is global and will affect all tools launched from that environment, which may be undesirable, so `$GIT_COLA_LANG` would give us a way to configure things just for Git Cola. I probably won't implement that unless someone is clamoring for it, though.

Regarding this specific PR -- I closed this PR because it was following the old guidance and really was removing too much. I don't think we'll need these changes with this new stance. I also don't think we should have tests that verify the lack of translations -- that was definitely going a bit too far. Sorry for wasting your time on that stuff @OriShalhon.

Anyways, hopefully we're in a better spot now. #1160 is now closed and we've addressed the original issues by adjusting *just* the slightly-confusing German translations and by adding tooltips that make things much clearer.

Thanks everyone for sharing your thoughts and hopefully this ended up being an improvement overall.

Here's a quick view of everything that changed in the translations. (NOTE: the improved handling for `--` in the "dag" and "diff" commands is a new feature so you'll need to run from the source tree to pickup the latest)

```bash
git diff v4.3.2 -- cola/i18n
./bin/git-cola diff v4.3.2 -- cola/i18n 
./bin/git-dag v4.3.2.. -- cola/i18n
```

PR #1375, opened 2024-01-23
Remarks to commits in rebase sequence editor

laerreal:
When rebasing big enough branches (100+ commits), it's a challenge to
join related commits into groups and to move the groups.
This PR adds a feature into rebase sequence editor that allows to add
simple remarks (digits [0-9]) to commits to visually highlight them.
A remark can be added to single commit or to all commits those touch a file.

Highlighting of commits touching a file requires expensive calls to
underlying `git`.
So, a cache for intermediate results is provided.
The cache is filled in background by a `Thread` to reduce first use GUI freeze
experience.

Russian translation for new menu items is also added.
But, it's is added manually because I have troubles with installing
the `garden` tool.

### DIFF
--- a/cola/gitcmds.py
+++ b/cola/gitcmds.py
@@ -69,7 +69,6 @@ def diff_index_filenames(context, ref):
 
 def diff_filenames(context, *args):
     """Return a list of filenames that have been modified"""
-    git = context.git
     out = diff_tree(context, *args)[STDOUT]
     return _parse_diff_filenames(out)
 
--- a/cola/i18n/ru.po
+++ b/cola/i18n/ru.po
@@ -4207,6 +4207,15 @@ msgstr ""
 msgid "yyyy-MM-dd"
 msgstr ""
 
+msgid "Remarks"
+msgstr "Пометки"
+
+msgid "Toggle remark"
+msgstr "Переключить пометку"
+
+msgid "Toggle remark of touching commits"
+msgstr "Переключить пометку затрагивающих коммитов"
+
 #~ msgid ""
 #~ "\n"
 #~ "\n"
--- a/cola/sequenceeditor.py
+++ b/cola/sequenceeditor.py
@@ -205,6 +205,13 @@ class Editor(QtWidgets.QWidget):
         self.filewidget.files_selected.connect(self.diff.files_selected)
         self.filewidget.remark_toggled.connect(self.remark_toggled_for_files)
 
+        # `git.show` is too expensive.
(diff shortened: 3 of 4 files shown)


laerreal:
How should I fix CI issue about FIXME at `cola/sequenceeditor.py:260:17`.
I am not familiar enough with Qt.
But leaving such comment without a `FIXME`/`TODO` tag may have consequences...

davvid:
Good stuff. The translations don't have to be "built" anymore these days so what you did by editing them directly is all that's needed. I'll take a stab at addressing the FIXMEs this weekend before merging in.

I pushed up a `gitcmds.changed_files(context, oid)` function that can be used to instead of the `git show` parsing here. It might be marginally faster since time won't be spent computing a diffstat. It uses the `git diff-tree` plumbing to get the list of changed files.

I'll commit those tweaks when I merge this into my local tree sometime this weekend. I'll probably also replace the raw thread with a qtutils.Task so that we can leverage the QThreadPool stuff we do elsewhere.

For the FIXME, we should be able to hook into a window close event by overriding `def closeEvent(self)` to shut down the long-running threads. That's what I'll be looking into. This as an almost-complete starting point so it should be pretty straightforward to get that working.

I'll make sure to re-fetch this weekend in case you push anything else in here before then. Thanks for getting this going.

davvid:
Merged! I think I addressed all of the todos now. I changed some of the logic around how toggling works (it not toggles individual items) but I'd be happy to change that back if you think it works better by resetting the state of everything in unison.

I also made it so that cancelling the dialog does not save out an empty file. It's better to leave the file as-is and return exit status 1 since that's how we convey to git that we want to cancel the rebase.

Thanks for the new feature, good stuff.

davvid:
Heads-up @laerreal I added colors to the remarks display to make the remarks easier to see at a glance. I also tweaked the multi-select behavior to behave more intuitively. I hope you dig it.

We now toggle selected items as a group by toggling items towards the majority. For example, if you have 4 items selected with 3 enabled and 1 disabled then the single disabled item will be re-enabled when the toggle action is performed on the selected items. Likewise, if 3 items are disabled and 1 is enabled then the single enabled item will be disabled.

The implementation is slightly more complex to explain but the observed behavior ends up being simpler because the selected items end up in a uniform state. This is closer to what you had originally implemented.

After the initial merge I had changed the behavior to toggle each item individually but treating the selection like a coherent group is definitely better. Thanks again!

PR #1380, opened 2024-02-06
sequenceeditor: fix remarks hiding on item motion (drag & drop)

laerreal:
Looks like Qt drops `QLabel` with remarks on item motion.

### DIFF
--- a/cola/sequenceeditor.py
+++ b/cola/sequenceeditor.py
@@ -689,6 +689,7 @@ class RebaseTreeWidgetItem(QtWidgets.QTreeWidgetItem):
         self.branch = branch
         self.comment_char = comment_char
         self.remarks = remarks
+        self.remarks_label = None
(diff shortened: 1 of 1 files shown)


cola/sequenceeditor.py
```diff
@@ -818,8 +830,6 @@ def set_remarks(self, remarks):
                 ">&nbsp;{remark} </span>
             """
         label.setText(label_text)
-        self._parent.setItemWidget(self, self.REMARKS_COLUMN, label)
-        self._parent.resizeColumnToContents(self.REMARKS_COLUMN)
```
davvid, line 822: Thanks for the fix. I'll merge this shortly but I'll probably restore the call to `resizeColumnsToContents` here. The purpose of this call is that it ensure that the column shrinks when remarks are removed.

PR #1382, opened 2024-02-15
Add AppStream developer name

Mailaender:
https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#name-summary-and-developer-name is now mandatory so all the builds at https://github.com/flathub/com.github.git_cola.git-cola/pulls currently fail.


davvid:
Thanks!

PR #1386, opened 2024-03-08
Fix an AppStream validation error 

Mailaender:
https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#name-summary-and-developer-name


davvid:
Thanks!

PR #1397, opened 2024-05-25
widgets: add an "Unstage" option to the context menu for unmerged files

lefairy:
When merging patches, if you want to cache only a partial change for a file that has already been manually resolved on merge, in previous versions you had to stage the entire file, then find it among the other already staged files. With "Unstage" you can easily find the file in "Modified" then specifically stage just the line or hunk to be patched.


davvid:
Thanks!

PR #1404, opened 2024-06-15
[WIP] Emit notifications when pushing to remotes #350

### LINKED ISSUE #350: Create a system notification when push completes
It would be nice to get a system notification when a push (and maybe similar tasks like pull) completes. This would be use full if the GUI is not in focuse when the task completes.

VishnuSanal:


davvid:
TODO before we can merge this:

- [ ] In unix status code `0` is success, not `200` like it currently uses.
- [ ] `try/except` guards around the `notifypy` dependency in the code so that it's optional and won't break stuff it isn't not importable
- [ ] The message says that X commits were pushed, but the code doesn't know about or count commits. it's not X commits that we know about -- it's Y remotes that were updated that it could possibly calculate. The `all_results` variable does not contain the value of X or Y, you'd have to count inside the loop to get Y.
- [ ] Configuration to make this feature opt-in -- we can use `cola.notifyonpush`. Look at how the other config items are handled in the code and do the same.
- [ ] Read more of the existing code base and follow the existing practices and conventions. All of our user-facing strings are internationalized. The title and message should leverage the `N_(...)` function.
- [ ] The title can include more information. We can say "Y/Z remotes were updated", where Z is just the number of remotes we loop over.
- [ ] "Partially Successful" is not a great title. I think `Y/Z remotes pushed - Git Cola` might be a better title in all scenarios.
- [ ] If we change the title to be more information-dense then the description should follow suite and list the remotes, e.g. `Not pushed: bad-remote.\t\tPushed: origin, other-remote.`
- [ ] You can get an icon using `resources.package_data('icons', 'git-cola.svg')`.
- [ ] If there's an error `resources.package_data('icons', 'circle-slash-red.svg')` can be used instead.
- [ ] `cola/models/prefs.py` and `cola/widgets/prefs.py` would need updates for a new opt-in `cola.notifyonpush` config variable and corresponding widgetry.
- [ ] `setup.cfg` needs updates to add this to the optional dependency to "extras"
- [ ] Documentation and README.md updates to mention notify-py alongside the other extras.
- [ ] Better commit messages ~ read `git log --no-merges` and follow the same style as the longer, more descriptive messages.
- [ ] Rebase and squash in the fixups into the original commits. We shouldn't see any commits from pre-commit bots in your final commits.

Just curious.. was this written using an AI coding assistant? I ask because the `200` vs. `0` mistake seems like the kind of thing an LLM would mess up, but I apologize for the catty question if that's just a :cat: :keyboard:  mistake.

Thanks for the post but I won't be merging this until these items are addressed. This is just a draft currently so I figured you wanted some critical feedback early on. I hope that's helpful as a path towards getting this in an acceptable state.


Extra cherry on top:

- [ ] Add a "red" version of the logo and we'd can show that instead of `circle-slash-red.svg` when there's an error.

That's a lot to ask so let's just focus on the first list of items for now since those are the core issues. Thanks.

VishnuSanal:
hi David, thanks for your extensive review! it is really rare to get a detailed review like this, that too for a draft PR. thanks for helping me figure out! :)

* [x]  In unix status code `0` is success, not `200` like it currently uses.
* [x]  `try/except` guards around the `notifypy` dependency in the code so that it's optional and won't break stuff it isn't not importable
* [x]  The message says that X commits were pushed, but the code doesn't know about or count commits. it's not X commits that we know about -- it's Y remotes that were updated that it could possibly calculate. The `all_results` variable does not contain the value of X or Y, you'd have to count inside the loop to get Y.
* [x]  Read more of the existing code base and follow the existing practices and conventions. All of our user-facing strings are internationalized. The title and message should leverage the `N_(...)` function.
* [x]  The title can include more information. We can say "Y/Z remotes were updated", where Z is just the number of remotes we loop over.
* [x]  "Partially Successful" is not a great title. I think `Y/Z remotes pushed - Git Cola` might be a better title in all scenarios.
* [x]  You can get an icon using `resources.package_data('icons', 'git-cola.svg')`.
* [x]  If there's an error `resources.package_data('icons', 'circle-slash-red.svg')` can be used instead.
* [x]  Add a "red" version of the logo and we'd can show that instead of `circle-slash-red.svg` when there's an error.

TODO (will update soon):

* [ ]  Configuration to make this feature opt-in -- we can use `cola.notifyonpush`. Look at how the other config items are handled in the code and do the same.
* [ ]  If we change the title to be more information-dense then the description should follow suite and list the remotes, e.g. `Not pushed: bad-remote.\t\tPushed: origin, other-remote.`
* [ ]  `cola/models/prefs.py` and `cola/widgets/prefs.py` would need updates for a new opt-in `cola.notifyonpush` config variable and corresponding widgetry.
* [ ]  `setup.cfg` needs updates to add this to the optional dependency to "extras"
* [ ]  Documentation and README.md updates to mention notify-py alongside the other extras.
* [ ]  Better commit messages ~ read `git log --no-merges` and follow the same style as the longer, more descriptive messages.
* [ ]  Rebase and squash in the fixups into the original commits. We shouldn't see any commits from pre-commit bots in your final commits.

> Just curious.. was this written using an AI coding assistant? I ask because the `200` vs. `0` mistake seems like the kind of thing an LLM would mess up, but I apologize for the catty question if that's just a 🐱 ⌨️ mistake.

noooo, it was a mistake!! 🙈 I got `128` as response code, didn't check what it means, assumed it to some HTTP response codes & proceeded to use `200` 🏃🏿‍♂️ (now I [realize](https://stackoverflow.com/a/69631781/9652621) was due to auth issues) : D

README.md
```diff
@@ -64,6 +64,10 @@ Python modules are installed.
 enables cross-platform "Send to Trash" functionality.
 ([source](https://github.com/hsoft/send2trash))
 
+[notify_py](https://pypi.org/project/notify_py/)
+Cross-platform desktop notification library for Python
```
davvid, line 68: sug: instead of "Cross-platform ....," we can say, "enables delivery of desktop notifications."

cola/models/prefs.py
```diff
@@ -117,6 +118,7 @@ class Defaults:
     history_browser = 'gitk'
     icon_theme = 'default'
     inotify = True
+    notifyonpush = True
```
davvid, line 121: This should default to `False`.

cola/widgets/remote.py
```diff
@@ -3,6 +3,8 @@
 import time
 import os
 
+from cola.models import prefs
+
 from cola import resources
```
davvid, line 8: I must've not noticed when these absolute imports snuck in.

We should use `from ..models import prefs` and `from . import resources` instead of mentioning the `cola` package here.

I'll go ahead and update the pre-existing import on the main branch, so you'll have to rebase to pickup that tweak shortly.

cola/widgets/remote.py
```diff
@@ -698,32 +700,33 @@ def push_to_all(self, _remote, *args, **kwargs):
...
+                notification.message = N_(
+                    f'{pushed_remotes}/{total_remotes} remotes updated'
+                )
 
-            notification.send()
-        except (ImportError, ModuleNotFoundError):
-            log.show_remote_messages(self, N_('notifypy dependency not found'))
+                notification.send()
+            except (ImportError, ModuleNotFoundError):
+                log.show_remote_messages(self, N_('notifypy dependency not found'))
```
davvid, line 729: This is too much code under the try/except block. Also, this importError check should be at the top of the file when we try to import from notifypy.

If we hit the ImportError we should set Notify to None.

Then, move all of this logic into a standalone private helper function. Then this part can just have a single line where it calls that function.

That new function, let's call it `def _emit_push_notification(...)`, can check `if Notify is None:` and `return` early.

cola/widgets/remote.py
```diff
@@ -3,10 +3,19 @@
...
+from cola import resources
+
 from qtpy import QtGui
 from qtpy import QtWidgets
 from qtpy.QtCore import Qt
 
+try:
+    from notifypy import Notify
+except (ImportError, ModuleNotFoundError):
+    pass
```
davvid, line 17: Instead of `pass` we should set `Notify = None`.

Better yet, let's not use `from ... import` but instead just use `import notifypy` and in the `except` here we would set `notifypy = None` instead. That's a little more natural.

Then, the `_emit_push_notification(...)` function can check `if notifypy is None:` rather than having to test against a Capitalized name.

davvid: Also, sort the imports. `notifypy` should go before `qtpy` in that block of imports.

cola/widgets/remote.py
davvid: These imports should come later in the internal section, not up here in the stdlib and third-party section.

setup.cfg
```diff
@@ -78,6 +78,7 @@ extras =
 	send2trash
 	# Enables macOS-specific appearance themes.
 	pyobjc; sys_platform == 'darwin'
+	notify-py
```
davvid, line 81: This entry should go before send2trash and can include a "Enables desktop notifications." comment.

cola/widgets/remote.py
```diff
@@ -680,9 +689,45 @@ def push_to_all(self, _remote, *args, **kwargs):
...
+                    )
+                else:
+                    notification.title = N_('Push Failed')
+                    notification.icon = resources.package_data(
+                        'icons', 'git-cola-error.svg'
+                    )
+
+                notification.message = N_(
+                    f'{pushed_remotes}/{total_remotes} remotes updated'
+                )
```
davvid: This should be the title not the message. The message should mention the remote names that are looped over in the `remote` variable as being either failed being "Not pushed" or "Pushed".

If there are no "Not pushed" remotes then we shouldn't see that part of the message and it should only mentioned the "Pushed" repos. Likewise, if there are no "Pushed" remotes then we shouldn't see that part of the message and only the "Not pushed" section should be present.

cola/widgets/remote.py
```diff
@@ -680,9 +689,45 @@ def push_to_all(self, _remote, *args, **kwargs):
...
+
+                notification.send()
+            except (ImportError, ModuleNotFoundError):
+                log.show_remote_messages(self, N_('notifypy dependency not found'))
```
davvid: We don't need to emit a message that the notifypy dependency cannot be found. If it's there it's there, if it's not it's not.

cola/widgets/remote.py
```diff
@@ -116,6 +123,42 @@ def task(self):
...
+            f'Failed to push commits to {total_count - success_count} remotes '
+            f'{", ".join(list(set(selected_remotes) - set(pushed_remotes)))}'
+        )
+    else:
+        notification.title = N_('Push Failed')
+        notification.icon = resources.package_data('icons', 'git-cola-error.svg')
+
+        notification.message = N_(
+            f'No commits pushed to remotes: {", ".join(selected_remotes)}'
+        )
```
davvid, line 157: Thanks! This is looking close enough. I'll probably merge this after making a small fixup commit locally.

The internationalization methods translate fixed strings so we can't pass anything varying to them. The way it's used is to include python string `%(variable)` references in the strings and those references are actually part of the translated string. We then use the old-school python `%` format operator to format the output from `N_(...)` to fill in the variables.

I'll be adjusting the titles and messages. I'm going to make the titles match what I suggested because it's more information dense to include the remote counts in the title and to use a consistent (unchanging) title for all of the scenarios above.

Also, I'm going to pass the pushed a non-pushed remotes as separate variables so that we don't have to do fancy `set(...)` operations. It's better for the code to be simple and dumb.

Lastly, I'm going to adjust the messages to be more succinct with just `Pushed:` and `Not pushed:` notes. The reason is that we want to be as brief as possible (they're notifications) and also because, at least on this desktop environment, the message can only be a single line so `\n` newline separators don't visualize as expected. That's why I had mentioned `\t` originally -- using a tab did provide a visual separator in my testing and was recognized by the notification display.

The other reason why having a consistent title (1) and just the "Pushed" (2) and "Not pushed" (3) notes is that it minimizes the number of strings that need translation. It would only use 3 different strings. The current code above uses 6 different distinct strings that would need to be translated separately.

Hope that makes sense. Thanks again for getting things this far.

I'll be merging shortly after applying this fixup locally.

PR #1445, opened 2025-01-05
widgets: avoid some crashes when cwd is not found

scop:
This is an untested monkey patch to possibly fix some crashes that occur when cola's current working directory gets deleted from underneath it.

I see this every now and then, and these two are the first ones that I managed to reproduce immediately when attempting to reproduce. They might not be the best ways to fix the issues (or fix them in the first place), but if nothing else, they flag where the crashes occurred for me.


davvid:
Later on we might want to find a good place to detect this condition and force the user to open a different repository when this happens, but this is nonetheless an improvement over the current behavior for now. Thanks!

PR #1453, opened 2025-01-18
main: fix incorrect proxy detection in KDE when proxy is disabled globally

### LINKED ISSUE #1452: KDE proxy settings wrongly detected when it is previously set but disabled globally
The following proxy setting in the KDE system settings:

Image

will still make Git Cola try to connect to a previously set proxy as the

```bash
kreadconfig5 --file kioslaverc --group 'Proxy Settings' --key 'httpProxy'
```

command still returns the previous settings:

```text
192.168.49.1 8228
```

### Git

2.43.0

### Git Cola

4.10.2.dev14+g7bc34c3e.d20250118

### OS

Ubuntu 24.04

brlin-tw:
When the proxy is disabled globally in the System Settings in KDE the proxy host and address settings may still be preserved in the kioslaverc/Proxy Settings/*Proxy configuration settings.

Check the kioslaverc/Proxy Settings/httpProxy setting's value and return None when it is not set to "1"(meaning that the "Use manually specified proxy configuration" mode isn't selected).

Fixes #1452.

### DIFF
--- a/cola/models/main.py
+++ b/cola/models/main.py
@@ -716,10 +716,22 @@ def autodetect_proxy_kde(kreadconfig, scheme):
         '--group',
         'Proxy Settings',
         '--key',
-        f'{scheme}Proxy',
+        f'ProxyType',
     ]
     status, out, err = core.run_command(cmd)
-    if status == 0:
-        proxy = out.strip().replace(' ', ':')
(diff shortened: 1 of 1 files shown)


davvid:
thanks!

PR #1503, opened 2025-09-02
cmds: non-interactive git pull action

jm4R:
Most time we do git pull from tracking branch. Let's make the shortcut for such flow without unnecessary additional dialog. I've call it "Sync" but comments are welcome.

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -1028,14 +1028,10 @@ class PullTracking(ContextCommand):
         self.model.update_status()
         if status != 0:
             title = N_('Pull failed')
-            return Interaction.critical(
-                title, err
-            )
+            return Interaction.critical(title, err)
         else:
             title = N_('Pull success')
-            return Interaction.information(
-                title, out
-            )
+            return Interaction.information(title, out)
         return status, out, err
 
 
--- a/cola/widgets/action.py
+++ b/cola/widgets/action.py
@@ -73,7 +73,9 @@ class ActionButtons(QFlowLayoutWidget):
             N_('Pull...'), layout, tooltip=N_('Integrate changes using "git pull"')
         )
         self.sync_button = tooltip_button(
-            N_('Sync'), layout, tooltip=N_('Integrate changes from tracking branch using "git pull"')
+            N_('Sync'),
+            layout,
+            tooltip=N_('Integrate changes from tracking branch using "git pull"'),
         )
         self.stash_button = tooltip_button(
             N_('Stash...'),


.gitignore
```diff
@@ -29,4 +29,6 @@ __pycache__
 *.py[cod]
 *.swp
 *.zip
+*.user
+*.qtcreator
```
davvid, line 33: Please amend to remove these lines. You can setup a global gitignore instead. https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

cola/cmds.py
```diff
@@ -1021,6 +1021,20 @@ def command(self):
         return f'git remote set-url "{self.remote}" "{self.url}"'
 
 
+class PullTracking(ContextCommand):
+    def do(self):
+        status, out, err = self.git.pull()
```
davvid, line 1026: This should be `self.git.pull(ff_only=True)`. If it doesn't fast-forward then we should let it fail and tell the user to rebase or merge. Blind git pull like this creates a gross history when two or more authors are constantly intermixing pulls and local commits.

cola/widgets/action.py
```diff
@@ -72,6 +72,11 @@ def __init__(self, context, parent=None):
         self.pull_button = tooltip_button(
             N_('Pull...'), layout, tooltip=N_('Integrate changes using "git pull"')
         )
+        self.sync_button = tooltip_button(
+            N_('Sync'),
+            layout,
+            tooltip=N_('Integrate changes from tracking branch using "git pull"'),
+        )
```
davvid, line 79: I'd call it "Fast-forward Pull" and change the description accordingly so that we're reusing Git's jargon instead of inventing our own.

davvid: That said.. "Fast-forward Pull" is a lot of text for the Action buttons widget. Maybe the text can just say "Fast fwd" and the tooltip can mention that it's a fast-forward pull.

Or not.. I dunno.. maybe "Sync" is a better name compared to "Fast-fwd"?

"Update" might be more precise, given that we're not syncing out changes upstream. We're just updating our worktree with what has gotten pushed. I'd be ok with "Update" as long as there's a tooltip that says that it's a fast-forward pull or rebase.

If the fast-foward pull fails, and the user happens to have `pull.rebase` or `branch.<name>.rebase` configured, it would be a nice convenience if we did a fallback call to `git pull` to allow the rebase.

Actually, we wouldn't need to fallback at all. I think it'd be more of a simple check -- if they have `pull.rebase` or `branch.<current-branch>.rebase` configured then we can do a blind `git pull`.

If neither of these are configured then we can do a `git pull --ff-only` as noted in the review. The tooltip can explain this nuance.

This seems like it'd be the most convenient for a single-click button.

Is that too complicated.. should we be so opinionated as to just make the tool do a blind `git pull --rebase` unconditionally? I'm slightly resisting that because it could cause havoc for someone who did local merges, so the approach described above seems ok if we were to have an "Update" button.

If we wanted a one-click convenient action for rebasing, we could provide a one-click `Rebase` button that does `git fetch + git rebase`. That would kinda make this "Update" button redundant IMO.

To be clear -- I'm intentionally trying to avoid a blind `git pull` for the scenario where a merge commit would be created, i.e. when a user creates a local commit, and the remote has a new commit as well. The merges created by this scenario get really messy when the number of users pushing/committing >= 2.

-----------------------------------------------------------------------

All that said, what do you think about changing this button to say "Rebase" instead?

We would use `git fetch <remote-for-the-current-branch>` and `git rebase --autostash @{upstream}` when that gets run.

I think this would be better. We're encouraging rebasing w/ an autostash to make it convenient to use in more scenarios. And the button would say "Rebase" so users won't be surprised.

We probably should even include the `git rebase --rebase-merges` option in there in case the user did a local merge.. it seems best to preserve that during the rebase.

We should probably query to make sure the user has an upstream branch (and remote) configured so that we know what to rebase against.

`gitcmds.get_default_remote(context)` can give us which remote to use for `git fetch`. `gitcmds.tracked_branch(context)` can be used to check if we have an upstream branch. It'll basically return what `@{upstream}` would resolve to.

If we don't have an upstream branch configured, we would report an error, but we could also apply another heuristic. It's kinda loose and wild, but if we see that the remote is `origin` and there's only a single remote branch, then we can assume that that's the branch to use.

Only if we get past that check would we report an error asking the user to set an upstream branch using the Branches tool.

jm4R: OK, two things:

* `git pull` with default config and not additional flags is indeed a bad idea. I have `git config --global pull.rebase true` in my environment so it works like with `--rebase` flag all the time. I prefer `--rebase` but fixed `--ff-only` is good enough for me. We could also make new preferences entry to leave user that decision?

* after 3 days of using this from my fork I realized that I've implemented "Sync in" action (using my own invented jargon) and I miss "Sync out" that would push current branch to the tracking branch. Can I add commits with this feature to this PR or make another PR? For me, it's not relevant how the button will be captioned so I am open for all your suggestions. Maybe `Pull...`, `Push...`, `Pull`, `Push` (with and without `...` to mark interactive/not interactive mode) isn't that bad?

davvid: Using "..." is kind of appealing given that it keeps the widgets small, so I like that too.

I also like your "Sync in" and "Sync out" jargon for these new actions. As long as the tooltips and/or `git-cola.rst` explain what they do we're golden. Let's go with what you like better.

Thanks for taking this on.

davvid: Here's one heuristic that might be helpful for the pull scenario.

If the user has `pull.rebase` configured then cola can  `git.pull(rebase=True)` when it's configured to True and `git.pull(rebase_merges=True)` when it's configured to `'merges'`. Otherwise, cola can `git.pull(ff_only=True)` when it's not configured.

That should cover all the bases for me at least, and makes cola follow the user's configuration.

I slightly prefer that over `git.pull()` without any arguments because it's explicit. It does mean that cola has to somewhat shadow the logic that's internal to `git pull`, but that seems fine.

davvid:
I picked this up and made some tweaks to get this merged. This is now merged w/ the extra tweaks described above under a "Sync" button and command. Please give it a spin when you get a chance.

jm4R:
Sorry, I got out of time recently. I was planning to add "Sync out" together with this feature. Thanks for pushing this feature!

PR #1508, opened 2025-10-07
cmds: non-interactive git push action

jm4R:
"Sync out" action based on "Sync".
I was not sure if we should rename "Sync" to "Sync in" so I left it.

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -1066,6 +1066,46 @@ class Sync(ContextCommand):
         return status, out, err
 
 
+class SyncOut(ContextCommand):
+    """Push local changes to the tracking branch"""
+
+    def do(self):
+        current_branch = gitcmds.current_branch(self.context)
+        if not current_branch:
+            title = N_('Sync out failed')
+            message = N_('No current branch')
+            Interaction.critical(title, message=message)
+            return -1, '', message
+        tracked = gitcmds.tracked_branch(self.context, current_branch)
+        if not tracked:
+            title = N_('Sync out failed')
+            message = N_('No tracking branch configured for %s') % current_branch
+            Interaction.critical(title, message=message)
+            return -1, '', message
+        remote, remote_branch = gitcmds.parse_remote_branch(tracked)
+        if not remote or not remote_branch:
+            title = N_('Sync out failed')
+            message = N_('Invalid tracking branch: %s') % tracked
+            Interaction.critical(title, message=message)
+            return -1, '', message
+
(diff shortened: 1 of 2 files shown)


jm4R:
@davvid I've noticed now that you have changed that the action success now doesn't notify it via a window - only log it in console view (and I followed it in this PR). Namely the `Interaction.information` has been replaced by `display.push_notification`.

For me it is not convinient solution – I always must check if it actually worked as I don't have console in my layout. Do you think reverting it is OK? Or inventing another solution like introducing progress icon would be fine?

davvid:
Do you have the inotify2 module installed? `sudo apt install python3-notify2`.

That function doesn't print to the console. It emits a desktop notification which should be a little less obtrusive than having to dismiss the popup dialog while giving users more flexibility in how those notifications are consumed. e.g. their desktop environment might let them silence these notifications, for example.

Try it out w/ notify2 installed and it should be displaying a notification on completion.

cola/cmds.py
```diff
@@ -1066,6 +1066,46 @@ def do(self):
...
+        if not remote or not remote_branch:
+            title = N_('Sync out failed')
+            message = N_('Invalid tracking branch: %s') % tracked
+            Interaction.critical(title, message=message)
+            return -1, '', message
+
+        display_command = f'git push {remote} {current_branch}'
+        status, out, err = self.git.push(remote, current_branch)
+        Interaction.log_status(status, out, err)
+        self.model.update_status()
```
davvid, line 1095: I don't think we need to update status since push only affects the remote tracking branches. I'll fix that up locally when merging. Thanks for adding this.

jm4R:
@davvid I was not aware about that optional dependency and never ever know that git-cola shows desktop notifications. I think many users could have the same.

I am not really into python, can you consider making it mandatory dependency?

davvid:
I'm going to revisit the notifications before the next release. I'm just now realizing that the notify2 package only works on Linux. It's also technically a deprecated project so I'll be exploring other options.

I'm probably going to add a, "Display notifications using a popup" setting that'll default to false.

If we notice that notify2 is not installed then we'll log to the Console, otherwise we'll emit a desktop notification. Likewise, we'll display a popup message box like your change originally did when that setting is set to true.

This should work for non-Linux platforms as well since they'll fall into the "notify2 is not installed" scenario.

davvid:
:loudspeaker: heads-up that you can enable "Display desktop notification using popup dialogs" in the settings, or `git config --global cola.enablepopups true` to opt-in to using popups instead of notifications/console logs.

PR #1536, opened 2026-01-01
metadata: Tidy up

yakushabb:
- Use desktop-application as the component type
- Add bugtracker and vcs-browser URLs

More information: https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines


davvid:
Thanks!

Issue #1537, opened 2026-01-01
RFE: GitHub style diff coloring


scop:
It would be great to have "GitHub style" diff coloring in git-cola. (There's probably a better name for it, but I suppose this conveys the idea.)

For a simple example, https://github.com/git-cola/git-cola/commit/8c89ce1c43e07a4d0feefe07a2b19a5c8fd21343

<img width="1368" height="168" alt="Image" src="https://github.com/user-attachments/assets/39684ff2-e830-48d4-b12e-e4e549cc85b9" />

So instead of plain red/green lines, the changes within the modified lines would also be highlighted

davvid:
I'd like this as well but I will note that this was already discussed in the past and no one stepped up to implement it. I don't keep lingering feature requests open if no one is going to step up to make it happen.

https://github.com/git-cola/git-cola/issues/623

kt-devoss:
@scop   @davvid 
I've been wanting **GitHub-style intra-line highlighting** too!
If it gets implemented, it would be useful and interesting.

I've been hacking on a prototype.
Here’s a screenshot of the current state:


| <img width="" height="400" alt="Image" src="https://github.com/user-attachments/assets/cf6b534d-dcf8-4385-8873-1138d4e2ca66" /> |<img width="" height="400" alt="Image" src="https://github.com/user-attachments/assets/59376b60-e556-4dcc-ad3a-9a4389898c4d" />|
| --- | --- |

davvid:
wow, that's really really cool! If we could make the underlines optional then it'd be just about perfect in my book.

Mohsalam:
Hi! I’m interested in working on this issue and can start investigating right away

kt-devoss:
Update: I’ve added GitHub-style intra-line diff highlighting (overlayed on the existing line-level colors).
It uses 3 highlight colors for deletions/insertions/replacements (red/green/yellow).

PR: #1542 — please take a look and try it if you have time.

>  If we could make the underlines optional then it'd be just about perfect in my book.

I think so, too.
I’ve updated the PR to remove underlines/bold styling.
In practice the underline can be visually distracting (and easy to confuse with `_`), so the intra-line changes are now highlighted using background colors only.

<img width="" height="300" alt="Image" src="https://github.com/user-attachments/assets/1a9a2b5a-fe69-42b3-84da-66acf1cb37cb" />

kt-devoss:
**For anyone interested in this "intra-line diff" feature:**

[...]

Issue #623, opened 2016-11-06
Word diff


nafg:
Please highlight the words or characters that changed.

For example, how long does it take you to tell exactly what did change and what didn't change here?

![image](https://cloud.githubusercontent.com/assets/98384/20042211/efa1ebc6-a443-11e6-8b81-2b40ecdcbb40.png)

Vdragon:
> For example, how long does it take you to tell exactly what did change and what didn't change here?

3 seconds/line

nafg:
> 3 seconds/line

Then I suppose this:

![image](https://cloud.githubusercontent.com/assets/98384/20043429/3464b6dc-a45b-11e6-991c-7b7b04362713.png)

would take you about 50ms ;)

If you have a lot of changes it adds up quickly. Especially if, as often happens, I have a lot of changes, and I need to go through all of them and split them into multiple commits.

So (1) speed matters and it adds up, and (2) I want be certain that the line has only the change I want to commit

Thanks.

richarson:
@nafg Just my $0.02 and maybe I'm too used to unified diffs, but in your second example it actually took me a while to realize what I was looking at and what the differences were.

That said, I undestand your point completely, and if you could keep the "unified diff" look but highlight the modified words I think it would be the best of both worlds.

nafg:
Here's how the same change looks in gitlab (I believe github is basically the same):

![image](https://cloud.githubusercontent.com/assets/98384/20075766/9acf9574-a503-11e6-942d-bd87b83ca911.png)

That would be fine with me and I'm guessing with @richarson too

Epskampie:
I would love this feature too, here is how to do it in git commandline, with a `diff-hightlight` script that might be a useful starting point for an implementation: http://stackoverflow.com/a/15149253/146738

Myz:
Some time ago I implemented this partly and not fully tested in my personal fork of git-cola
https://github.com/Myz/git-cola/commit/9dc5153c7734a9a53b85f7b62fab0d9d4a8a5ecd

This still is missing one local commit but yes, this feature is doable and would be great to have in some situations when you would use the interactive mode from the cli.

Vdragon:
Adding my usecase:

I'm using Git to track several Flat-ODF document, every modification to the file will cause a lot of trivial line changes which simply altering the auto-numbered style name, with this feature it will be easier to determine which line contain the actual changes instead of these trivial changes.

[...]

PR #1542, opened 2026-01-25
diff-ui: add inline (intra-line) diff highlighting to diff viewer

kt-devoss:
### Summary

This PR adds **inline (intra-line)  diff** highlighting to the diff viewer: in addition to the existing line-level add/remove coloring, it highlights the changed substrings within a line.

Issues: #1537, #623
---

### What this changes

* Currently, the diff is rendered at the line level, showing per-line deletions and additions in light red/green.
* This PR **adds inline (intra-line) change highlighting** on top of that.
* Inline highlights use *stronger* emphasis than the baseline line colors to make edits easier to spot.

#### Highlight semantics

* **Deletion side (`-` lines):** emphasized red
* **Addition side (`+` lines):** emphasized green
* **Replacement (changed substring):** emphasized yellow

  * “Replacement” here means a substring that is considered modified (not purely inserted/removed) when comparing the paired `-` and `+` lines.

|  |  |
| --- | --- |
| current | **this PR** |

---

### Limitations / Scope

Inline diff is applied only when all of the following conditions are satisfied:

* Within the same hunk, a contiguous `-` block is immediately followed by a contiguous `+` block, and
* The number of `-` lines and `+` lines are equal (1:1 pairing).

If these conditions are not met (e.g., line count mismatch, larger reshapes, moves/splits/joins), inline diff is skipped and the viewer falls back to the existing line-level rendering.

Inline diff uses difflib.SequenceMatcher, which may not always produce the best-looking match for complex changes.

---

### Performance & Safety

* A guard is in place to skip inline diff for very large diffs / heavy cases to avoid UI stalls.
* Any errors during inline diff computation will safely fall back to the standard line-level diff display.

---

### Implementation notes

* The algorithm is documented step-by-step in code comments marked as `# [STEP] ...`, so the processing flow can be followed directly from the implementation.

---

### Testing

* Manually verified on typical replace/edit cases (e.g., small substring changes) and ensured fallback behavior for non-covered patterns / large diffs.

### DIFF
--- a/cola/widgets/diff.py
+++ b/cola/widgets/diff.py
@@ -1114,7 +1114,7 @@ class Options(QtWidgets.QWidget):
         self.intraline_diff_widget = QtWidgets.QWidget(self)
 
         self.intraline_diff_preset_label = QtWidgets.QLabel(
-            N_('Intra-line diff mode:'), self.intraline_diff_widget
+            N_('Intra-line diff mode'), self.intraline_diff_widget
         )
         intraline_diff_preset_data = [
             (item.label, item.preset_id)
@@ -1156,10 +1156,13 @@ class Options(QtWidgets.QWidget):
             self.set_intraline_diff_timing,
             False,
         )
+        self.intraline_mode_action = QtWidgets.QWidgetAction(self)
+        self.intraline_mode_action.setDefaultWidget(self.intraline_diff_widget)
         # ^^^ intra-line diff widget end
 
         self.menu = menu = qtutils.create_menu(N_('Diff Options'), self.options)
         self.options.setMenu(menu)
+        menu.addAction(self.intraline_mode_action)
         menu.addAction(self.max_diff_action)
         menu.addSeparator()
         menu.addAction(self.ignore_space_at_eol)
@@ -1179,7 +1182,6 @@ class Options(QtWidgets.QWidget):
             defs.no_margin,
             defs.button_spacing,
             self.options,
-            self.intraline_diff_widget,
             self.toggle_image_diff,
             self.filename,
             self.image_mode,
@@ -1191,8 +1193,6 @@ class Options(QtWidgets.QWidget):
         # Policies
         self.image_mode.setFocusPolicy(Qt.NoFocus)
         self.zoom_mode.setFocusPolicy(Qt.NoFocus)
-        self.intraline_diff_preset_combo.setFocusPolicy(Qt.NoFocus)
-        self.intraline_diff_widget.setFocusPolicy(Qt.NoFocus)
         self.options.setFocusPolicy(Qt.NoFocus)
         self.toggle_image_diff.setFocusPolicy(Qt.NoFocus)
         self.setFocusPolicy(Qt.NoFocus)
@@ -1207,7 +1207,6 @@ class Options(QtWidgets.QWidget):
         is_image = diff_type == main.Types.IMAGE
         self.image_mode.setVisible(is_image)
         self.zoom_mode.setVisible(is_image)
-        self.intraline_diff_widget.setVisible(not is_image)
         if is_image:
             self.toggle_image_diff.setIcon(icons.diff())
         else:


davvid:
Thanks, I'll review soon. Please squash the commits from the pre-commit bot into the original commits so that we don't have any bot commits in this PR.

Some of this looks AI generated (specifically the description section in this PR). If the commits themselves were authored with AI assistant tools please mention this in the commit messages so that we have a paper trail of the changes. [1]

Apologies if the commits were not authored w/ AI tooling ~ this is all good if so. I don't mind AI tools being used for PR descriptions since those are auxiliary to the code itself, but it would be good to know whether the code was authored that way for code provenance reasons.



Thanks again! I took a quick look at the code and this is very much on the right track. I do see that there's a variable to enable/disable this feature. That gives me something to run further with -- the next step would be to add a `cola.diffwords` option that can be used to enable/disable this feature.

The next refinement would be that currently the diff colors are configurable:
https://git-cola.readthedocs.io/en/latest/git-cola.html#cola-color-text

We're going to continue respecting this by adding additional configuration values for these new colors that are being introduced. In lieu of ocnfiguration, if there were a simple heuristic (e.g. "10% or 20% brighter") that we could apply to the current (configurable) colors for these "new" colors, then it would let us eschew configuration altogether for these new colors, which might be appealing as a first step since it is advantageous to avoid expanding the configuration space.

Performance-wise, it doesn't *seem* like this would have very much of a performance impact, so I'm feeling pretty positive about this PR.

Let me know what you think about these sugs and whether you'd be game to keep pushing on these last details before we consider merging them. If you'd prefer that I implement these last bits of polish I could find some time to make that happen.

If you have a chance, could you please include a link in the code to the Qt documentation where it states that the text offsets are utf16 code point offsets? This is surprising to me.. even though QString uses utf-16 encoding internally, our external interface is Python unicode text objecte. That part might be a mistake.. so please double-check and share whether or not that's actually needed. I'm hoping it's not actually needed.

Good stuff, and very much appreciated. This is a great start.

[1] I should probably update our CONTRIBUTING.md soon to mention this preference.

davvid:
More references that we should consider:

https://git-scm.com/docs/git-config#Documentation/git-config.txt-diffwordRegex

Core Git has a `diff.wordRegex` [configuration variable](https://git-scm.com/docs/git-config#Documentation/git-config.txt-diffwordRegex).

```
diff.wordRegex

    A POSIX Extended Regular Expression used to determine what is a "word" when performing
    word-by-word difference calculations. Character sequences that match the regular expression
    are "words", all other characters are ignorable whitespace.
```

As far as I can tell, it doesn't seem like we need to strictly honor this variable since we're just using a similarity index.

I reviewed a bit more. These look like they should be configuration variables:


```
_INLINE_DIFF_MAX_LINES = 5000
_INLINE_DIFF_MAX_BLOCK_LINES = 2000
_INLINE_DIFF_MAX_LINE_LENGTH = 1000

# Minimum similarity (0.0–1.0) for enabling inline highlighting:
#   - Increase -> inline highlights appear less often (only for more-similar line pairs),
#       fewer misleading spans, but less detail.
# See: difflib.SequenceMatcher()
_INLINE_DIFF_SQ_RATIO_THRESHOLD = 0.50
```

I'm slightly surprised that we need these limits at all, so it looks like there is a performance issue (contrary to what I stated above before reading the code).

One detail of QSyntaxHighlighter and our current implemention is that I was under the impression that the syntax highlighter would only highlight the blocks that are visible. I thought we fixed that a while ago ~ it should only be highlighting blocks that are actually visible so it'd be a bit of a regression if we're considering stuff that's outside of the view.

Please verify whether or not I'm understanding this correctly, but if this is the case then it's pretty important that we retain this property, especially since I'm going to want to have this feature enabled by default ~ I would like `cola.diffwords` to default to `True` so that it's an opt-out feature. If we do that then we should not need the limit variables at all since the performance should be snappy if we avoid analyzing the full set of diffs when the diff is huge.

kt-devoss:
Thanks for taking a look! I’m honored.
I’ve tried to incorporate the feedback, and I added a small extra improvement as well.

> Please squash the commits from the pre-commit bot into the original commits so that we don't have any bot commits in this PR.

But, I do that, the bot’s changes will end up mixed into this PR instead, I think.

We might be slightly talking past each other here.
About the pre-commit bot commits: it looks like the bot only applied style fixes to `sphinxtogithub.py`. I didn’t touch that file in this PR, so it seems unrelated to my changes. I’m not sure if the bot config changed.

 For now, I rebased so that the bot’s `sphinxtogithub.py` fix becomes the first commit in this PR. However, I think it may be better to move that change to a separate PR, since it’s not part of this PR’s scope and the same issue will likely happen for others on the current `main`. What do you think?

<<edited, snipped>>

I used AI assistance for coding and writing comments (Of course I’m human). I added a note about this to the commit message.

---
Qt

I double-checked this. Qt’s text APIs do use UTF-16 code unit offsets, not Python codepoint indices.
When the text contains surrogate pairs (e.g. emoji), passing Python indices directly to QSyntaxHighlighter.setFormat() results in misaligned spans.

To handle this correctly, I added small helper functions to convert Python codepoint indices to UTF-16 offsets before calling Qt:

_qt_index_from_cp()
_qt_span_from_cp()

I also made a small demo with the code showing the incorrect vs correct behavior with emoji here:
https://gist.github.com/kt-devoss/e43d7f7eef9761746f26c5116ff834ec
image
moji-bake occurred

---

Related to the UTF-16 offset handling, I did a small refactor to make the responsibilities clearer.
I moved the Qt-specific parts (UTF-16 code-unit offsets and QSyntaxHighlighter-related conversions) into diff.py.
I also separated the pure Python logic (inline span computation) into diffinline.py so it has no Qt dependency.

This keeps the Qt coupling localized and makes the core logic easier to test.

---

Color
For the inline highlight colors, I switched from hard-coded values to deriving them from the existing diff background colors.
The idea is to keep the same hue and just make the inline highlights more prominent.

---

On performance: I initially added quite a few guardrails to be conservative. They’re currently disabled (set to unlimited) while I dogfood and measure real usage. 
If anyone’s interested in this feature, I’d love for them to give it a try.

I also added simple timing logs around the span computation (API compute_inline_diff_spans via measure_ms(unit: msec)) so we can see the cost per diff update. 

So far it’s called once per diff update (e.g. when selecting a file), and I’m seeing roughly ~0.01–40 ms. If this stays snappy, I’d prefer to keep it enabled by default and avoid hard limits.

Thanks for the feedback — much appreciated

*the current screenshot*:
| image |image |
|---|---|

davvid:
Thanks, much appreciated! I fixed up the sphinxtogithub formatting on main so once you rebase that should go away. I'll take another look again soon.

kt-devoss:
Thanks! I was able to rebase and push without issues.

Note: the most recent push is rebase-only. The notes below describe the current status and what I’ve been working on since your earlier feedback.

kt-devoss:
## Current status / direction
(Some of the items below are in-progress and not all are pushed yet.)

#### UI responsiveness on huge diffs
- Added lightweight cancellation to the inline diff computation so we can stop early and keep the UI responsive.

#### Clear separation of concerns
- Refactored the core so diffinline.py is stdlib-only and GUI-independent.
- Qt/GUI-specific concerns (timing policy, rendering/styling, UTF-16 indexing) stay on the caller side (diff.py).

## Inline diff core (current scope)
- Tokenization is currently character-level (this is what I initially pushed).
- Pairing logic is intentionally conservative at the moment: strict 1:1 pairing only.

I prototyped word-level diff and it looks promising visually (less visual noise), but it can be problematic for languages without explicit word boundaries (e.g. Japanese).
So I think it’s better to make word-mode selectable(and/or Hybrid).
I'm also prototyping more flexible m:n pairing, which should require relatively minor changes; for now, I am only lightly testing it.

## Cancellation behavior
- diffinline.compute_inline_diff_spans() accepts an optional should_cancel callback.
- diff.py can apply a simple wall-clock time budget; if exceeded, we stop early and return partial spans (or fall back to no inline spans depending on the caller’s policy).
- This avoids fragile per-scenario guardrail constants while preserving interactivity.
- Under normal-sized diffs the callback never triggers and the computation runs to completion; the budget is mainly to prevent UI stalls on very huge diffs.

kt-devoss:
## UI / configuration (in progress)
Longer-term, the core can support any combination of:
- tokenization mode (char / word / …)
- pairing strategy (strict 1:1 vs more flexible m:n variants)

In the core these axes are independent. For the UI, I’m leaning toward starting with a small set of presets
(bundling tokenization + pairing choices) using a stable identifier,
and leaving room for an optional “advanced” configuration later—some power users may appreciate being able to tune tokenization and pairing separately. Also, it’s fun to flip the dropdown and watch the diff update live.

Once we align on the UI approach, it should also make it clearer what the right config surface and key naming should be
(e.g. a simple boolean vs an enum/preset id).

kt-devoss:
At this point, interface changes and internal structure changes are somewhat intertwined.
Before re-pushing a PR, I plan to clean up and simplify the interfaces and overall structure, also to keep the design extensible for future enhancements.

Because of that, it may make sense to do a more thorough code review after those refactorings are done.
I’ll post again once that is in, when the code structure should be easier to review in depth.

davvid:
Thanks for the update, that sounds like a good plan.

I'm curious to see what the cancellation approach looks like. If it gets too complicated then that's something we can live with in the form of a simpler approach. For example, a simpler (non-time) based metric such as # lines and could be used to dynamically enable/disable the feature to account for large diffs.

The ideal would be for the performance to be proportional to just the displayed diff, not the entire diff in the text widget beyond the viewport. That way there wouldn't be a notion of a "large" diff because the highlighting would only be considering the text that is visible. That's the ideal, though. It sounds like it's dependent on the entire diff currently, which is why large diffs would be noticeable.

If that's a limitation of the approach then, yeah, having an upper bound based on the # lines in the diff, or if the time approach is simple enough, then that should do the trick.

As far as UI, I would expect this to be a bool to enable/disable this feature full stop in the Diff widget's options sprocket menu as a starting point.

If we end up having line # limits, then that'd be a spinbox in there. Notably, we already have a setting for maximum diff size, so it would be adjacent to that. If it's time-based, the time limit could be set there.

Setting the # lines (or time limit) to 0 should be equivalent to infinity (no limit).

davvid:
@kt-devoss what are your thoughts on the status of the inline diff feature? Should I hold off on looking into merging this version since you have plans to push a newer version? Thanks again, looking forward to getting this feature in here soon!

kt-devoss:
@davvid Sorry for the delayed reply (Last month was busy, but I'm still alive 🙂 Things should be calmer this month.)

Yes, please hold off for now. I will push a newer version.

davvid:
Thanks for following up. I'll wait for your new version before integrating.

For the record, I was taking a look at the current version and added options to the Diff tool's options menu for enabling and disabling the inline diff and timing features interactively:

https://github.com/davvid/git-cola/commit/ce1147fc279737b5b82f285b990daee663ffa1f3

There's also a parent commit where I moved some of the helper functions around:

https://github.com/davvid/git-cola/commit/e8e8b3b31d1209014bc0800fae39261366b11814


The commits are in the `inline-diff` branch of my davvid fork: https://github.com/davvid/git-cola/tree/inline-diff


To checkout that branch:

```
git remote add davvid https://github.com/davvid/git-cola.git
git fetch davvid
git checkout inline-diff
```

You should be able to cherry pick just these commits into your local branch if you want to pickup just the two commits that I added:

```
git cherry-pick e8e8b3b31d1209014bc0800fae39261366b11814
git cherry-pick ce1147fc279737b5b82f285b990daee663ffa1f3
```

davvid:
I'll also mention that I'm perfectly fine with the current state of this feature as it's sitting in the `inline-diff` branch. 

Adding the toggle controls to make this feature able to be disabled interactively is enough to satisfy the perf overhead concerns.

In other words, if you think this is in a good place I would be happy to merge it in that state.

kt-devoss:
@davvid Thanks for the feedback and for the follow-up patches.

Actually, I reworked the core soon after the previous version to make it easier to try to develop and compare different mode's codes. Because of that, the structure and interfaces changed a lot, so your patches do not apply cleanly to this version as-is.

In this updated version, I also cleaned up the core API. The core and GUI were already separate, but I adjusted the split between the core and `widgets.diff`. I also moved the intra-line widget/helper code out of `widgets.diff`, mainly into `widgets.diff_intraline`, and moved some related small helpers into other files as well.

The new version is no longer limited to strict same-index 1:1 pairing. It now also supports a nearby-greedy pairing mode for locally shifted -/+ blocks, and both char-level and the new word-ish intra-line matching.
These modes are selectable from the diff viewer so they can be compared interactively.

I also picked up the logging/timing control idea, and rebased onto the current `main`.

Before, intra-line diff was just on/off. Now there are multiple mode combinations. Each strategy/granularity has different strengths and weaknesses, so I used a combobox to let you switch modes quickly, including with the mouse wheel, and compare them in the viewer.

Please try it out and let me know whether this structure and UI/config direction make sense.

Feedback is welcome. Thanks!

kt-devoss:
I also split the commits again to make the history simpler, which should make this PR easier to review with less noise.

Files changed and their roles:

- `cola/intraline_diff.py`: pure core logic for intra-line diff
- `cola/widgets/diff.py`: GUI side, including glue code
- `cola/widgets/diff_intraline.py`: helper code used from the GUI side (new file)
- `cola/qtutils.py`: some small related helpers
- `cola/utils.py`: some small related helpers

kt-devoss:
Also, unchanged parts inside changed lines are now shown with slightly less saturated colors, so the changed parts are easier to see.
So now there are 6 derived colors in total.

image

image

cola/widgets/diff.py
```diff
@@ -283,8 +340,119 @@ def set_diff(self, diff):
...
+        if update and hasattr(self, 'options') and self.options is not None:
+            self.options.set_intraline_diff_preset(
+                self._intraline_diff_preset, update=True
+            )
+        self.update_intraline_diff_spans()
+
+    def set_intraline_diff_timing(self, enabled, update=False):
+        """Store the intra-line diff timing option."""
+        self._intraline_diff_timing = enabled
+        if update and hasattr(self, 'options') and self.options is not None:
```
davvid, line 452: I'd suggest minimizing the dict/attr lookups by expressing this as:

```
if update and getattr(self, 'options', None) is not None:
```

cola/widgets/diff.py
```diff
@@ -86,8 +95,23 @@ def __init__(self, context, doc, whitespace=True, is_commit=False):
...
+            diff_intraline.IntralineStyleSet.from_base_colors(
+                text_foreground=self.color_text,
+                added_line_background=self.color_add,
+                removed_line_background=self.color_remove,
+            )
+        )
+
         self.setCurrentBlockState(self.INITIAL_STATE)
 
+    def _set_intraline_spans(self, spans):
```
davvid: Rename this to `def set_intraline_spans(self, spans):` without the leading `_` underscore. The new code calls this method on the highlighter from outside the class, so we should treat it as not-private.

cola/widgets/diff.py
```diff
@@ -30,9 +33,12 @@
 from .text import TextSearchWidget
 from .text import label_selection_timer
 from . import defs
+from . import diff_intraline
 from . import standard
 from . import imageview
 
+ENABLE_INTRALINE_DIFF = True
```
davvid, line 40: Since the enable state is based on the mode, it doesn't seem like we'll need this variable anymore, correct? It seems like this is more of useful aid while debugging the code. Would it be ok to do without it?

cola/widgets/diff.py
davvid: I did this in a follow-up commit.

cola/widgets/diff.py
davvid: I did this in a follow-up commit.

davvid:
I moved the widget into the diff options menu (click on the sprocket icon), addressed a few of the items I mentioned, and merged. Thanks! Please feel free to open a new PR for any follow-up enhancements.

davvid:
I haven't looked into this yet, and something that would be worth fixing, is that the `./bin/git-dag` viewer does not respond to changes to the intra-line diff mode, nor does it remember its settings. I think the settings recall should be simple. I suspect we need to hookup a callback to get it to rehighlight itself on toggle.

If you've got a hotfix for that I'll roll that into the next patch release.

I included this in the minor release I just tagged, btw. Maybe it was a bit rash of me, but this feature seems like it would benefit from more users using it, so I went for it.

davvid:
git-dag has updates for this on the main branch now. Thanks!

cola/widgets/diff.py
kt-devoss: Yes. Not needed.

kt-devoss:
Wow, that was a very quick update!

Very nice UI/UX choice. I did not think of that approach.

At first, I avoided the options menu because I thought quick switching would be harder if I had to keep opening and closing the menu.
But this works very well: I can keep the menu open, use the mouse wheel, and see the diff update right away.
It feels well balanced, smart and less distracting for beginners, too.
It also seems easy to extend later if we add more strategies or tokenization modes.
Thank you!

I am glad that many people may get to use the algorithm I worked on.

kt-devoss:
> I'm still slightly wondering if the colors could be made slightly more subtle (perhaps by trying to keep the original background colors as a starting point?) but I don't think that's a blocker from merging or worth churning on at this stage. It's fine as-is. That's the kind of thing that can/should be tweaked and refined later.

@davvid 
I agree. I also felt the colors can look a bit harsh, especially in dark themes.

image

To handle that, I wrote a small helper to derive the intra-line colors from the original add/remove background colors with HSL-based adjustments in this PR. I wanted it to work reasonably in both light and dark themes, but I could not tune it well enough for both in this PR.

I am still thinking about a better approach that works more consistently across themes.
If I can implement it well, I may send a follow-up PR later.

kt-devoss:
**For anyone interested in this new feature:**

As a quick hack, you can tweak the derived intra-line colors here:
[cola/widgets/diff_intraline.py](https://github.com/git-cola/git-cola/blob/68a6dd793476e420c4c94bc0c799c7d31ba24688/cola/widgets/diff_intraline.py#L67)
Windows:
`C:\Users\<YourLoginID>\AppData\Local\Programs\git-cola\pkgs\cola\widgets\diff_intraline.py`

```
        # Compute derived background colors for inserted, deleted, unchanged,
        # and replaced intra-line spans.

        stronger_added_background = adjust_hsl(
            self.added_line_background,
            sat_ratio=+0.95,    # <---
            light_ratio=-0.10,  # <---
        )
        stronger_removed_background = adjust_hsl(
            self.removed_line_background,
            sat_ratio=+0.95,  # <---
            light_ratio=-0.10,  # <---
        )
        ...
```

The easiest way to adjust them is to change the `sat_ratio` and `light_ratio` values in the `adjust_hsl()` calls.

- `sat_ratio`: saturation, in the `-1.0` to `+1.0` range
- `light_ratio`: lightness, in the `-1.0` to `+1.0` range

For a more subtle look, try lowering `sat_ratio` or moving `light_ratio` closer to `0`.

cf. about saturation, lightness
https://en.wikipedia.org/wiki/HSL_and_HSV#/media/File:Hsl-hsv_models.svg

The intra-line hightlight colors are derived from the existing add/remove line colors, so they are strongly affected by those base colors. So I think the best place to start is by adjusting the existing add/remove colors first.

PR #1545, opened 2026-01-29
Update ru.po

PushKK:
Small update, small fix.

### DIFF
--- a/cola/i18n/ru.po
+++ b/cola/i18n/ru.po
@@ -52,7 +52,7 @@ msgstr ""
 
 #: cola/app.py:603 cola/widgets/startup.py:256
 msgid "Repository Not Found"
-msgstr ""
+msgstr "Репозиторий не найден"
 
 #: cola/app.py:604 cola/widgets/bookmarks.py:486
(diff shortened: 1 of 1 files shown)


cola/i18n/ru.po
```diff
@@ -4509,11 +4509,11 @@ msgstr ""
 #: cola/widgets/submodules.py:47
 #, fuzzy
 msgid "Open Parent"
-msgstr "Открыть последний"
+msgstr "Открыть Родителя"
 
 #: cola/widgets/submodules.py:86
 msgid "git://git.example.com/repo.git"
-msgstr ""
+msgstr "https://github.com/UserName/RepoName"
```
davvid, line 4516: Just a heads-up that I changed this string (and all of the in-app uses of this string) to be `https://git.example.com/user/repo.git`. Nothing to do here -- it's already been updated.

This tool has to be kept neutral and not necessarily advertise one forge over another, which is why I made the change. I did like that your change added a "user" part to the URL, so I incorporated that into the untranslated base value.

PR #1553, opened 2026-02-12
Enable mypy for type checking

p12tic:
At this point it mostly does not do anything. The objective is to have checks running so that any new code with type hints could enable the checks on itself and benefit from automatic checks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,10 +130,11 @@ warn_no_return = false
 warn_unreachable = false
 ignore_missing_imports = true
 disable_error_code = [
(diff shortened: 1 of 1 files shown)


p12tic:
Regarding adding type annotations to existing code - the diff wouldn't be huge. I expect[1] it to be less than 2k lines, which would fit into roughly 5 large PRs. I'm not planning to work on this right now, just looked into how reasonable this change would be.

[1]:
```
grep 'def ' $(git ls-files | grep 'cola/'  | grep -v widgets ) | wc -l
1603
```

PR #1554, opened 2026-02-12
RFC: prefs: Add a setting to control size of previous commit message menus

p12tic:
I didn't open an issue since I thought it would be easy enough to code it to demonstrate what I want. Thus I assume the code can be thrown away :-)

Often I'm working on a large screen with many in-flight commits in a branch - up to 20-30 if counting fixup commits. In such situation the current workflow to commit a fixup commit is not convenient. If I want to fixup a commit from long ago, I need to click on the gear icon, go to "fixup previous commit menu", figure out that the commit I want is not visible, then click on "More", lose focus as an additional window opens, figure out which commit messages I've just read so that I can start searching for the commit I want...

The number of recent repositories is configurable, I think it would be reasonable to configure these menus to reuse a previous commit message as well.

Please see an attached screenshot demonstrating that even at 20 visible commits it occupies reasonable amount of screen real estate. 
image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,7 @@ Usability, bells and whistles
   "Fixup Previous Commit" menus is now configurable through the Preferences
   window.  The ``cola.loadcommitmsgcount`` and ``cola.fixupcommitcount``
   configuration variables control the number of commits displayed in each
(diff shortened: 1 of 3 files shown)


davvid:
Thanks!

PR #1560, opened 2026-03-06
Add more type hints in cola/ directory

Ahrirg:
Implemented more type hints in cola/ directory using [monkeytype](https://monkeytype.readthedocs.io/en/stable/) + manual verification.

### DIFF
--- a/cola/app.py
+++ b/cola/app.py
@@ -75,8 +75,7 @@ if TYPE_CHECKING:
     from qtpy.QtCore import QEvent
     from qtpy.QtGui import QPalette
 
-    from .types import TextType
-    from .widgets.standard import Dialog, MainWindow, Widget
+    from .types import TextType, ViewType
 
 
 def setup_environment() -> None:
@@ -477,7 +476,7 @@ def enforce_single_instance(context: ApplicationContext) -> None:
 
 def application_run(
     context: ApplicationContext,
-    view: Widget | Dialog | MainWindow,
+    view: ViewType,
     start: Callable | None = None,
     stop: Callable | None = None,
 ) -> int:
@@ -496,9 +495,7 @@ def application_run(
     return result
 
 
-def initialize_view(
-    context: ApplicationContext, view: Widget | Dialog | MainWindow
-) -> None:
+def initialize_view(context: ApplicationContext, view: ViewType) -> None:
     """Register the main widget and display it"""
     context.set_view(view)
     view.show()
@@ -512,17 +509,13 @@ def application_start(context, view) -> int:
     return application_run(context, view, start=default_start, stop=default_stop)
 
 
-def default_start(
-    context: ApplicationContext, _view: Widget | Dialog | MainWindow
-) -> None:
+def default_start(context: ApplicationContext, _view: ViewType) -> None:
     """Scan for the first time"""
     QtCore.QTimer.singleShot(0, startup_message)
     QtCore.QTimer.singleShot(0, lambda: async_update(context))
 
 
-def default_stop(
-    _context: ApplicationContext, _view: Widget | Dialog | MainWindow
-) -> None:
+def default_stop(_context: ApplicationContext, _view: ViewType) -> None:
     """All done, cleanup"""
     QtCore.QThreadPool.globalInstance().waitForDone()
 
@@ -756,10 +749,10 @@ class ApplicationContext:
         self.settings: Settings | None = None
         self.selection: selection.SelectionModel | None = None
         self.fsmonitor: fsmonitor._Monitor | None = None
-        self.view: Widget | Dialog | MainWindow | None = None  # QWidget
+        self.view: ViewType | None = None  # QWidget
         self.browser_windows = []  # list of browse.Browser
 
-    def set_view(self, view: Widget | Dialog | MainWindow) -> None:
+    def set_view(self, view: ViewType) -> None:
         """Initialize view-specific members"""
         self.view = view
         self.notifier.setParent(view)
--- a/cola/types.py
+++ b/cola/types.py
@@ -1,7 +1,9 @@
 from typing import Union
 
 from .core import UStr
+from .widgets.standard import Dialog, MainWindow, Widget
 
 
 ConfigValue = Union[bool, str, int]
 TextType = Union[str, UStr]
+ViewType = Union[Dialog, MainWindow, Widget]


cola/actions.py
```diff
@@ -1,37 +1,63 @@
...
+from cola.widgets.diff import DiffEditor
 
 from . import cmds
 from . import difftool
 from . import hotkeys
 from . import icons
 from . import qtutils
 from .i18n import N_
+from PyQt5.QtGui import QIcon
+from PyQt5.QtWidgets import QAction
```
davvid: This looks like a mistake. We should never import from `PyQt*` directly. We should always use the `qtpy` wrapper module.

Also, the imports above are incorrect -- `from cola....` should never show up in the code. We use relative imports everywhere.

I really hope we don't end up causing circular import dependencies. If we start running into those then that's where we'd have to pull back on it.

Last note -- the `from typing ...` import should be much higher, nearby the stdlib imports.

davvid: I now realize that I didn't catch the `from cola ....` imports in the earlier pull request. I should have caught that.

That's going to require some manual fixups. I may fix that up directly on the main branch, so I wouldn't be surprised if that causes conflicts for these changes. Heads-up that I'm going to re-review the previous PRs and I'll likely fix any issues related to these issues (use of absolute cola imports and the misplaced `typing` imports).

davvid: I'm also seeing `from collections.abc import ...` in an earlier PR that was placed in the wrong location. Please make sure to keep the same exact import order as the original code. It's strict PEP-8. stdlib imports go first.

davvid:
I just fixed up the circular import dependencies (!) and the messed-up imports. Please take a look at what I pushed.

If you run into circular import dependencies that means that the type annotations are wrong and should a different approach that doesn't introduce cycles.

Thanks for taking this on. Hopefully this can keep going without too many snags.

Ahrirg:
I am sorry you had to fix my mistakes. I'm not sure how I let those circular dependency errors slip through. I would have fixed the issue myself as quickly as possible if I had known.  I will be more careful next time.

Ahrirg:
I replaced all direct `PyQt*` imports with `qtpy`. All imports are fixed to be relative and import style follows the PEP 8 standard.

Regarding import sorting, would you like me to setup some sort of automation tool like [isort](https://pypi.org/project/isort/) and run it on the codebase as a next PR?

cola/utils.py
```diff
@@ -28,12 +32,15 @@ def asint(obj, default: int = 0) -> int:
     return value
 
 
-def clamp(value, low, high):
+T = TypeVar('T', int, float)
```
davvid: "T" is too generic a name for this concept IMO. I would suggest `NumberT` maybe?

Also, please keep all type declarations at the top of the file after the imports section.

davvid:
Please rebase and squash together the monkeytype changes and the fixup commits that follow them.

cola/actions.py
```diff
@@ -1,4 +1,10 @@
 """QAction creator functions"""
+from __future__ import annotations
+
```
davvid, line 3: `__future__` should be grouped together with the stdlib imports. Let's drop the empty line here.

cola/actions.py
```diff
@@ -7,31 +13,52 @@
 from . import qtutils
 from .i18n import N_
 
+if TYPE_CHECKING:
+    from .app import ApplicationContext
+    from .cmds import LaunchEditor, LaunchEditorAtLine, StageOrUnstage
+    from .widgets.commitmsg import CommitMessageEditor
+    from .widgets.diff import DiffEditor
+    from .widgets.status import StatusTreeWidget
+
```
davvid, line 22: Thanks, `if TYPE_CHECKING:` is a nice solution for the circular dependencies issue.

cola/app.py
```diff
@@ -69,6 +72,9 @@
 from . import utils
 from . import version
 
+if TYPE_CHECKING:
+    from .widgets.main import MainView
+
```
davvid, line 79: Hmm.. not quite. `app.application_start(...)` is also called by `cola/dag.py` and a different class is passed in. Semantically, this really just requires a QWidget, and does not need to be associated with the concrete MainView class.

`cola/main.py` also has a bunch of calls with other classes. If we wanted to be a little more precise, we could say that this expects a `.widgets.standard.WidgetMixin`, but that loses all of the Qt methods from being visible, since that's a mixin class.

Practically, we could define a type that is a union over `.widgets.standard.Widget`, `.widgets.standard.Dialog` and `.widgets.standard.MainWindow`. That's probably the most semantically valid set of types that the `view` variables expect.

cola/app.py
```diff
@@ -265,22 +276,22 @@ def stop(self) -> None:
...
-    def __init__(self, context, argv) -> None:
+    def __init__(self, context: ApplicationContext, argv: list[str]) -> None:
         super().__init__(argv)
         self.context = context
         # Make icons sharp in HiDPI screen
         if hasattr(Qt, 'AA_UseHighDpiPixmaps'):
             self.setAttribute(Qt.AA_UseHighDpiPixmaps, True)
 
-    def event(self, e):
+    def event(self, e: QEvent | QTimerEvent | QChildEvent) -> bool:
```
davvid: I think these are supposed to be `QEvent` only?

cola/interaction.py
```diff
@@ -11,14 +13,28 @@ class Interaction:
...
 
     @classmethod
-    def command(cls, title, cmd, status, out, err) -> None:
+    def command(
+        cls,
+        title: str | core.UStr,
+        cmd: list[str | core.UStr] | str,
+        status: int,
+        out: str | core.UStr,
+        err: str | core.UStr,
```
davvid: Is there a way to define a type alias for `str | core.UStr` so that we can avoid all of this duplication?

cola/sequenceeditor.py
```diff
@@ -86,28 +90,28 @@ def parse_args():
...
 
-    def __init__(self, context, parent=None) -> None:
+    def __init__(self, context: ApplicationContext, parent: Any | None = None) -> None:
         super().__init__(parent)
         self.context = context
         self.status = 1
         # If the user closes the window without confirmation it's considered cancelled.
         self.cancelled = True
-        self.editor = None
+        self.editor: Any | None = None
```
davvid: Is `Any | None` redundant, or is that best practice? In other words, if we see `Any`, does that not also mean that `None` is a valid argument?

I think it'd be best to just see `Any` here. It seems like this is how it's intended to be.

cola/themes.py
```diff
@@ -28,27 +29,27 @@ class EStylesheet:
...
         self.is_dark = is_dark
-        self.is_palette_dark = None
+        self.is_palette_dark: bool | None = None
         self.style_sheet = style_sheet
         self.main_color = main_color
         self.macos_appearance = macos_appearance
         self.disabled_text_color: str | None = None
         self.text_color: str | None = None
         self.highlight_color: str | None = None
         self.background_color: str | None = None
```
davvid, line 51: For common things such as `str | None` can we define a type alias maybe called `MaybeStr`?

Likewise `MaybeInt` and `MaybeBool` possibly.

cola/actions.py
```diff
@@ -1,4 +1,10 @@
 """QAction creator functions"""
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from qtpy.QtGui import QIcon
+from qtpy.QtWidgets import QAction
```
davvid: Since a lot of these imports are just for `TYPE_CHECKING` then it'd be better to keep the imports inside `if TYPE_CHECKING:` when they're not used in the actual code.

cola/gitcfg.py
```diff
@@ -207,10 +221,10 @@ def _get(self, src, key, default, func=None, cached: bool = True):
             if func:
                 value = func()
             else:
-                value = default
+                value: bool | str | int | None = default
```
davvid: This might be a good place for a `ConfigValue` type alias as well.

cola/gitcfg.py
```diff
@@ -207,10 +221,10 @@ def _get(self, src, key, default, func=None, cached: bool = True):
...
         return value
 
-    def _get_value(self, src, key):
+    def _get_value(self, src: dict[str, str | int | bool], key: str) -> Any:
```
davvid: We'll want to reuse these type aliases across modules. Carving out a new `cola/types.py` ~ `cola.types` python namespace for these aliases would make sense.

cola/app.py
```diff
@@ -705,32 +725,32 @@ def __init__(self) -> None:
...
         self.timer: Timer | None = None  # Timer
-        self.runtask: qtutils.RunTask | None = None  # qtutils.RunTask
-        self.settings = None  # settings.Settings
+        self.runtask: qtutils.RunTask | None = None
+        self.settings: Settings | None = None
         self.selection: selection.SelectionModel | None = None
-        self.fsmonitor = None  # fsmonitor
-        self.view = None  # QWidget
+        self.fsmonitor: fsmonitor._Monitor | None = None
+        self.view: MainView | None = None  # QWidget
```
davvid: This and all other `view` / `_view` variables and function parameters would need updating for the new type alias over the `.widgets.standard.{Dialog,Widget,MainWindow}` base classes.

cola/core.py
```diff
@@ -152,40 +163,42 @@ def write(path, contents, encoding=None, append: bool = False):
...
-    cwd=None,
-    add_env=None,
+    cmd: list[UStr | str],
+    cwd: UStr | None = None,
+    add_env: None = None,
     universal_newlines: bool = False,
-    stdin=subprocess.PIPE,
-    stdout=subprocess.PIPE,
+    stdin: int | None = subprocess.PIPE,
+    stdout: BufferedWriter | int = subprocess.PIPE,
```
davvid: Our usage only uses these as `MaybeInt`, and `sys.stdout` is not a `io.BufferedWriter`, so it might be better to drop BufferedWriter here.

cola/version.py
```diff
@@ -78,51 +80,51 @@ def version():
...
+def version_to_list(value: str) -> list[int | str]:
     """Convert a version string to a list of numbers or strings"""
     ver_list = []
     for part in value.split('.'):
         try:
-            number = int(part)
+            number: int | str = int(part)
         except ValueError:
-            number = part
+            number: int | str = part
```
davvid: I think number really should just be `int`.

Slight, but correct, change in behavior is that we shoulddo `number = 0` here instead of `number = part`.

This code path doesn't hurt things in practice if we ever reach it, so it seems better to simplify things by keeping the return type as `list[int]` instead of the theoretical possibility of `list[int | str]`.

cola/sequenceeditor.py
```diff
@@ -639,11 +647,11 @@ def move(self, src_idxs, dst_idx) -> None:
 
     # Qt events
 
-    def dropEvent(self, event) -> None:
+    def dropEvent(self, event: QtGui.QDropEvent) -> None:
         super().dropEvent(event)
         self.validate()
 
-    def contextMenuEvent(self, event) -> None:
+    def contextMenuEvent(self, event: QtGui.QDropEvent) -> None:
```
davvid: I believe this is `QContextMenuEvent`, not `QDropEvent`.

davvid (changes requested):

p12tic:
> Please rebase and squash together the monkeytype changes and the fixup commits that follow them.

I think it is useful to have monkeytype commits separately, because it records type information on runtime. In case we discover weird typing later on then it's useful to know that a particular type was in particular place on runtime and that weird typing is not just human error.

cola/themes.py
p12tic: I would recommend against that. `str | None` is standard pattern in Python. In the past we even had `Optional[str]` and Python migrated away from that.

cola/sequenceeditor.py
p12tic: It's redundant indeed

cola/actions.py
p12tic: I believe `__future__` is considered a separate import group. Both ruff and isort adds a empty line after `from __future__ import annotations`.

cola/interaction.py
p12tic: What could it be named as? AnyStr?

davvid: We could call it `Text` or `TextType`, perhaps?

davvid:
One reason I wanted to squash the monkeytype commits w/ the fixups is because this is getting to be a large series with 61 commits in it already.

I think it might be worth squashing all of the monkeytype commits together into one commit and all of the human fixup commits into one commit. That'll cut down on the vast majority of commits, and we'll keep just the later ones where type aliases and other adjustments are made as separate commits.

While you're in there, please adjust the commit messages. Take a look at `git log origin/main` and follow the same capitalization style. The subject line should be lowercase after the `<category>:` prefix.

Also, please sign-off your commits. `git commit --amend -s` or the `ctrl-i` hotkey in the commit message editor should add your sign-off.

This is getting pretty close. I don't like my `git log` getting overloaded with all of these small commits, so I appreciate the rebase hygiene.

davvid:
Cherry on top ~ the squashed together monkeytype commit message would benefit from including details in the extended description section about which specific commands were run, a note about which version was used, and a link to monkeytype's project page ~ that kinda thing.

The fixup commit should aim to explain why the fixups are being done so that a future editor can apply the same criteria when updating other modules, for example.

p12tic:
Hmm, I no longer see our discussion about MaybeInt for some reason. Maybe a force-push has hidden it somehow. For the record - our discussion was about replacing `int | None` with `MaybeInt` and I recommended against that because `int | None` is standard python typing practice.

davvid:
> Hmm, I no longer see our discussion about MaybeInt for some reason. Maybe a force-push has hidden it somehow. For the record - our discussion was about replacing `int | None` with `MaybeInt` and I recommended against that because `int | None` is standard python typing practice.

Not sure why the discussion went away but `int | None` is fine.

Ahrirg:
As requested, I squashed the commits, updated the commit message format, and added additional details about the MonkeyType tool that was used.

Please let me know if any further changes are required.

cola/actions.py
davvid: Just noting -- we don't use ruff or isort. We're using cercis, which is fine with this line being removed. These blank lines are being dropped in the main branch.

If we later switch to ruff then we can let the tool add these. I don't like having a section "before" the stdlib. In that sense, future is part of the stdlib in my simple grug brain so I just like it better that way.

Issue #1561, opened 2026-03-09
Option for setting branch tracking when pushing a new branch


KayEss:
I keep forgetting to set the checkbox, and then later on when pulling code using the command line (often via scripts) things break because the tracking isn't set up. Now I have to go and try to remember the correct command for setting the tracking up.

Is it possible to have it at least remember the checkbox between using the push dialog box?

It does appear that it ignores any options for this that are set in the git config. Maybe it shouldn't do that?

Thanks

PR #1562, opened 2026-03-13
Add optional preference to remember upstream checkbox in Push dialog

legasik21:
Fixes issue #1561

This PR adds an optional preference that allows Git Cola to remember the
"Set upstream tracking" checkbox state in the Push dialog.

Behavior:
- Default behavior remains unchanged.
- When the preference is enabled, Git Cola remembers the last used
  upstream checkbox state and restores it when reopening the Push dialog.

Changes:
- Added new preference key in prefs model
- Added UI toggle in Preferences
- Implemented push dialog logic to persist and restore checkbox state
- Added tests for prefs and remote push behavior

Tests:
11 tests passing locally.

p.s. 

These are some of my first issues. If it needs anything else, please let me know. Thanks!


davvid:
I believe we already resolved this when we merged the changes in 4b666d08

PR #1563, opened 2026-03-14
push: confirm upstream persistence when enabling tracking

legasik21:
Prompt when enabling ‘Set upstream’ so users can choose whether it applies once or is remembered for future pushes. Persist remembered choice in Push dialog state and restore it on reopen. Clear remembered behavior when users disable the checkbox so re-enabling requires confirmation again

### DIFF
--- a/cola/widgets/remote.py
+++ b/cola/widgets/remote.py
@@ -931,11 +931,9 @@ class Fetch(RemoteActionDialog):
 class Push(RemoteActionDialog):
     """Push to remote repositories"""
 
-    REMEMBER_UPSTREAM_STATE_KEY = 'remember_upstream'
-    UPSTREAM_STATE_KEY = 'upstream'
-
     def __init__(self, context, parent=None):
         super().__init__(context, PUSH, N_('Push'), parent=parent, icon=icons.push())
+        self.remember_upstream_checkbox_state = False
         self.upstream_checkbox.toggled.connect(self.upstream_checkbox_toggled)
 
     def upstream_checkbox_toggled(self, enabled):
@@ -944,9 +942,9 @@ class Push(RemoteActionDialog):
         Choose between one-time and remembered behavior.
         """
         if not enabled:
-            self._remember_upstream_checkbox = False
+            self.remember_upstream_checkbox_state = False
             return
-        if getattr(self, '_remember_upstream_checkbox', False):
+        if self.remember_upstream_checkbox_state:
             return
         title = N_('Set upstream')
         msg = N_(
@@ -958,7 +956,7 @@ class Push(RemoteActionDialog):
         )
         ok_text = N_('Yes, remember this setting for future pushes')
         cancel_text = N_('No, enable this setting just once for the next push')
-        self._remember_upstream_checkbox = Interaction.confirm(
+        self.remember_upstream_checkbox_state = Interaction.confirm(
             title,
             msg,
             info_txt,
@@ -974,10 +972,10 @@ class Push(RemoteActionDialog):
         state['force'] = get(self.force_checkbox)
         state['prompt'] = get(self.prompt_checkbox)
         state['tags'] = get(self.tags_checkbox)
-        remember_upstream = bool(getattr(self, '_remember_upstream_checkbox', False))
-        state[self.REMEMBER_UPSTREAM_STATE_KEY] = remember_upstream
-        if remember_upstream:
-            state['upstream'] = get(self.upstream_checkbox)
+        if self.remember_upstream_checkbox_state:
+            state['set_upstream_branch'] = get(self.upstream_checkbox)
+        else:
+            state.pop('set_upstream_branch', None)
         return state
 
     def apply_state(self, state):
@@ -992,13 +990,10 @@ class Push(RemoteActionDialog):
         # Restore the "tags" checkbox
         tags = bool(state.get('tags', False))
         self.tags_checkbox.setChecked(tags)
-        self._remember_upstream_checkbox = bool(
-            state.get(self.REMEMBER_UPSTREAM_STATE_KEY, False)
-        )
-        if self._remember_upstream_checkbox and self.UPSTREAM_STATE_KEY in state:
-            upstream = bool(state.get(self.UPSTREAM_STATE_KEY, False))
-            with qtutils.BlockSignals(self.upstream_checkbox):
-                self.upstream_checkbox.setChecked(bool(upstream))
+        set_upstream_branch = bool(state.get('set_upstream_branch', False))
+        self.remember_upstream_checkbox_state = set_upstream_branch
+        with qtutils.BlockSignals(self.upstream_checkbox):
+            self.upstream_checkbox.setChecked(set_upstream_branch)
         return result
 
 
--- a/test/remote_test.py
+++ b/test/remote_test.py
@@ -23,9 +23,7 @@ class FakeCheckbox:
 
 class DummyPushDialog:
     def __init__(self):
-        self.REMEMBER_UPSTREAM_STATE_KEY = 'remember_upstream'
-        self.UPSTREAM_STATE_KEY = 'upstream'
-        self._remember_upstream_checkbox = False
+        self.remember_upstream_checkbox_state = False
         self.context = Mock()
         self.force_checkbox = FakeCheckbox(False)
         self.prompt_checkbox = FakeCheckbox(True)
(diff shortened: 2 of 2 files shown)


cola/widgets/remote.py
```diff
@@ -931,15 +931,53 @@ def apply_state(self, state):
...
+    def upstream_checkbox_toggled(self, enabled):
+        """Prompt once when enabling upstream.
+
+        Choose between one-time and remembered behavior.
+        """
+        if not enabled:
+            self._remember_upstream_checkbox = False
+            return
+        if getattr(self, '_remember_upstream_checkbox', False):
+            return
```
davvid, line 948: Thanks for addressing my notes from the earlier version.

`self._remember_upstream_checkbox = False` should be initialized in the constructor.

After construction, `apply_state()` gets called, so we'll be able to flip this value to `True` during `apply_state`.

That should allow us to get the desired behavior without needing to store the this extra state in the `export/apply_state()` dict.

cola/widgets/remote.py
```diff
@@ -931,15 +931,53 @@ def apply_state(self, state):
...
+            'Do you want this setting to be remembered and apply to future pushes?'
+        )
+        ok_text = N_('Yes, remember this setting for future pushes')
+        cancel_text = N_('No, enable this setting just once for the next push')
+        self._remember_upstream_checkbox = Interaction.confirm(
+            title,
+            msg,
+            info_txt,
+            ok_text,
+            default=False,
```
davvid, line 964: Thanks for making the default value here `False`. If it were `True` then it would be slightly easier since users could just hit `Enter` to confirm, but since this behavior is opt-in having a default of `False` makes more sense.

cola/widgets/remote.py
```diff
@@ -931,15 +931,53 @@ def apply_state(self, state):
...
+        )
 
     def export_state(self):
         """Export persistent settings"""
         state = RemoteActionDialog.export_state(self)
         state['force'] = get(self.force_checkbox)
         state['prompt'] = get(self.prompt_checkbox)
         state['tags'] = get(self.tags_checkbox)
+        remember_upstream = bool(getattr(self, '_remember_upstream_checkbox', False))
+        state[self.REMEMBER_UPSTREAM_STATE_KEY] = remember_upstream
```
davvid: I want to believe that we can get the expected behavior without needing to store the `'remember'` key.

In other words.. if we have the `'upstream'` key set to `True` when applying settings then we know that they must have chosen to persist the setting.

cola/widgets/remote.py
```diff
@@ -931,15 +931,53 @@ def apply_state(self, state):
...
+        if remember_upstream:
+            state['upstream'] = get(self.upstream_checkbox)
```
davvid: I forget whether the dict is retained across runs, but if we wanted to handle the `else:` situation here (i.e. `remember_upstream` is False) then we could do `state.pop('upstream', None)` in an `else:` block here to ensure that the field is cleared when exporting.

cola/widgets/remote.py
```diff
@@ -931,15 +931,53 @@ def apply_state(self, state):
...
+        msg = N_(
+            'The upstream branch will be configured when branches are pushed.\n'
+            'This can affect all subsequent pushes.'
+        )
+        info_txt = N_(
+            'Do you want this setting to be remembered and apply to future pushes?'
+        )
+        ok_text = N_('Yes, remember this setting for future pushes')
+        cancel_text = N_('No, enable this setting just once for the next push')
+        self._remember_upstream_checkbox = Interaction.confirm(
```
davvid: This part here should be enough to get the expected behavior without needing to store a `'remember'` key. The class instance will know.

cola/widgets/remote.py
```diff
@@ -954,6 +992,13 @@ def apply_state(self, state):
         # Restore the "tags" checkbox
         tags = bool(state.get('tags', False))
         self.tags_checkbox.setChecked(tags)
+        self._remember_upstream_checkbox = bool(
+            state.get(self.REMEMBER_UPSTREAM_STATE_KEY, False)
+        )
+        if self._remember_upstream_checkbox and self.UPSTREAM_STATE_KEY in state:
+            upstream = bool(state.get(self.UPSTREAM_STATE_KEY, False))
+            with qtutils.BlockSignals(self.upstream_checkbox):
+                self.upstream_checkbox.setChecked(bool(upstream))
```
davvid: I believe this can be slightly simpler, but note that this method has to be involved with the `remember` state in order to have the choice persist.

Assume that we don't have the `'remember'` state in the dict, but we do have the `'upstream'` state set to True. We should be able to make it work with something along these lines without needing to write out a separate `'remember'` setting to disk:

```python
set_upstream_branch = bool(state.get('set_upstream_branch', False))
if set_upstream_branch: 
    # Ensure that `set_upstream_branch` will be re-exported.
    self._remember_upstream_checkbox = True

with qtutils.BlockSignals(self.upstream_checkbox):
    self.upstream_checkbox.setChecked(set_upstream_branch))
```

This example uses `set_upstream_branch` as the key name since that's a bit more explicit and closer to what's displayed in the UI. I'd suggest using that longer name.

BTW, it's okay to just use the key name directly rather than having a constant for it. We only use it twice (in `apply_state()` and `export_state()`) and that's what we do for the rest of the fields. We don't expect to use this key name outside of these specific methods or this class so it's fine IMO.

test/remote_test.py
```diff
@@ -0,0 +1,118 @@
+from unittest.mock import Mock
+from unittest.mock import patch
+
+from cola.widgets import remote
```
davvid, line 4: Thank you for adding a unit test for this feature!!

test/remote_test.py
```diff
@@ -0,0 +1,118 @@
...
+    def blockSignals(self, value):
+        previous = self.blocked
+        self.blocked = value
+        return previous
+
+
+class DummyPushDialog:
+    def __init__(self):
+        self.REMEMBER_UPSTREAM_STATE_KEY = 'remember_upstream'
+        self.UPSTREAM_STATE_KEY = 'upstream'
```
davvid: Ah, I hadn't seen that the tests also use these key names. Call me silly, but I still think it'd be okay. These tests are rather tightly coupled to the implementation, so I'm fine with this minimal amount of duplication in the name of simplicity.

test/remote_test.py
```diff
@@ -0,0 +1,118 @@
...
+        self.context = Mock()
+        self.force_checkbox = FakeCheckbox(False)
+        self.prompt_checkbox = FakeCheckbox(True)
+        self.tags_checkbox = FakeCheckbox(False)
+        self.upstream_checkbox = FakeCheckbox(False)
+
+
+def test_push_export_state_remembers_upstream_when_enabled():
+    dialog = DummyPushDialog()
+    dialog._remember_upstream_checkbox = True
```
davvid: Since the tests are banging on this boolean I would suggest just making it a "public" field and calling it `self.remember_upstream_checkbox_state` so that the tests don't look like they're diving into a "private" field. There's no harm in having it be "public".

test/remote_test.py
```diff
@@ -0,0 +1,118 @@
...
+    dialog.upstream_checkbox.setChecked(True)
+    with patch.object(remote.RemoteActionDialog, 'export_state', return_value={}):
+        state = remote.Push.export_state(dialog)
+    assert state['remember_upstream'] is True
+    assert state['upstream'] is True
```
davvid: I would almost suggest dropping the `is True` part and letting it be truthy. `assert <expr>` already ensures that `<expr>` is truthy.

davvid (changes requested):
Thanks for this new version. It's very very close now!! I really appreciate you taking this over the last mile. Very nicely done! This looks good to merge as soon as these last little details are ironed out.

legasik21:
Addressed all requested changes in eadf037. I simplified the persisted state to use set_upstream_branch, initialized the remember state in the constructor, and updated tests accordingly. Could you take another look?

cola/widgets/remote.py
```diff
@@ -932,14 +932,50 @@ class Push(RemoteActionDialog):
     """Push to remote repositories"""
 
     def __init__(self, context, parent=None):
+        self.remember_upstream_checkbox_state = False
```
davvid: Typically, all initialization should happen after calling the parent's init constructor.

I'll squash that in locally before merging. You'll have to `git reset --hard origin/main` post-merge to pickup the rewritten commit.

davvid:
The commits were merged -- github gets confused if I merge a rewritten commit before pushing the rewritten commit. I should have done it in the other order. It's merged.

PR #1565, opened 2026-03-16
models: reload GIT_COLA_MSG on each refresh cycle

kurellajunior:
As you think, this is a good idea, I went ahead and tried to add it. I used
the help of such an agent, but reviewed and adapted it carefully. I tried to
keep it as non-invasive as possible. If I am going into the wrong direction,
please tell me.

## Summary

- Add mtime-based detection in `_update_commitmsg`: on each refresh,
  `_update_cola_msg` reads the file when its mtime has changed and updates
  the commit message editor – unless the user has manually edited the field
  since the last auto-load

## Use case

External tools (coding agents) or scripts can write a suggested commit
message to `.git/GIT_COLA_MSG` at any time and have it appear in the open
editor on the next refresh (`Ctrl+R`)

### DIFF
--- a/cola/models/main.py
+++ b/cola/models/main.py
@@ -114,6 +114,8 @@ class MainModel(QtCore.QObject):
         self.commitmsg = ''  # current commit message
         self._auto_commitmsg = ''  # e.g. .git/MERGE_MSG
         self._prev_commitmsg = ''  # saved here when clobbered by .git/MERGE_MSG
+        self._cola_commitmsg = ''  # last content loaded from .git/GIT_COLA_MSG
+        self._cola_commitmsg_mtime: float | None = None  # .git/GIT_COLA_MSG mtime on last read
         self.object_format = 'sha1'  # repository object format variables.
         self.oid_len = git.OID_LENGTH_SHA1
         self.empty_tree_oid = git.EMPTY_TREE_SHA1
@@ -447,6 +449,36 @@ class MainModel(QtCore.QObject):
         elif self._auto_commitmsg and self._auto_commitmsg == self.commitmsg:
             self._auto_commitmsg = ''
             self.set_commitmsg(self._prev_commitmsg)
+
+        elif not self._auto_commitmsg:
+            self._update_cola_commitmsg()
+
+    def _update_cola_commitmsg(self) -> None:
+        """Reload the commit message when .git/GIT_COLA_MSG changes on disk
+
+        Updates the commit message only when the user has not modified the
+        previously loaded content, so manual edits are never clobbered.
+        """
+        path = gitcmds.commit_message_path(self.context)
+        if not path:
+            self._cola_commitmsg_mtime = None
+            return
+        try:
+            mtime = os.path.getmtime(path)
+        except OSError:
+            return
+        if mtime == self._cola_commitmsg_mtime:
+            return
+        msg = core.read(path)
+        self._cola_commitmsg_mtime = mtime
+        if msg != self._cola_commitmsg:
+            # needs stripping, as the empty commitmsg is saved as \n during close but initialized as ''
+            if (
+                not self.commitmsg.strip()
+                or self.commitmsg.strip() == self._cola_commitmsg.strip()
+            ):
+                self.set_commitmsg(msg)
+            self._cola_commitmsg = msg
 
     def update_submodules_list(self) -> None:
         self.submodules_list = gitcmds.list_submodule(self.context)
--- a/cola/models/main.py
+++ b/cola/models/main.py
@@ -222,7 +222,6 @@ class MainModel(QtCore.QObject):
 
     def set_commitmsg(self, msg: str, notify: bool = True) -> None:
         self.commitmsg = msg
-        self._auto_commitmsg = ''
         if notify:
             self.commit_message_changed.emit(msg)
 
--- a/docs/git-cola.rst
+++ b/docs/git-cola.rst
@@ -445,6 +445,13 @@ The `Ctrl+i` keyboard shortcut adds a standard "Signed-off-by: " line,
 and `Ctrl+Enter` creates a new commit using the commit message and
 staged content.
 
+The commit message is saved to `.git/GIT_COLA_MSG` when `git cola` closes and
+restored from it on the next startup. In addition this file is checked on updates (Ctrl+R)
+and if modified by any tool and not conflicting with manual changes the commit
+message will be updated on the next refresh.
+This feature is useful as it allows scripts or agents to prepare a message for manual
+review.
+
 Sign Off
 --------
 
--- a/test/main_model_test.py
+++ b/test/main_model_test.py
@@ -295,6 +295,30 @@ def test_cola_msg_suppressed_during_merge(app_context):
     os.unlink(merge_path)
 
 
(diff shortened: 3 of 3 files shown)


test/main_model_test.py
```diff
@@ -258,6 +258,32 @@ def test_remote_args_rebase_only(mock_context):
...
+def test_cola_msg_updated_when_not_manually_edited(app_context):
+    """GIT_COLA_MSG is loaded into the commit message on refresh
+
+    GIT_COLA_MSG update is applied when the user has not edited the message"""
+    msg_path = app_context.git.git_path('GIT_COLA_MSG')
+    helper.write_file(msg_path, 'first message\n')
+    app_context.model._update_commitmsg()
+    assert app_context.model.commitmsg == 'first message\n'
+    helper.write_file(msg_path, 'updated message\n')
+    os.utime(msg_path, (os.path.getmtime(msg_path) + 1,) * 2)  # ensure mtime differs on coarse-grained filesystems
```
kurellajunior, line 270: If that is not necessary, I am happily removing this line. The pipeline should run on modern hardware, right?

davvid: Correct, `os.stat(filename).st_mtime` granularity should be reliable.

cola/models/main.py
```diff
@@ -114,6 +114,8 @@ def __init__(self, context, cwd: str | None = None) -> None:
         self.commitmsg = ''  # current commit message
         self._auto_commitmsg = ''  # e.g. .git/MERGE_MSG
         self._prev_commitmsg = ''  # saved here when clobbered by .git/MERGE_MSG
+        self._cola_commitmsg = ''  # last content loaded from .git/GIT_COLA_MSG
+        self._cola_msg_mtime: float | None = None  # .git/GIT_COLA_MSG mtime on last read
```
davvid: sug: for consistency we might want to call this `self._commitmsg_mtime` or `self._cola_commitmsg_mtime`

cola/models/main.py
```diff
@@ -448,6 +450,34 @@ def _update_commitmsg(self) -> None:
...
+        except OSError:
+            return
+        if mtime == self._cola_msg_mtime:
+            return
+        msg = core.read(path)
+        self._cola_msg_mtime = mtime
+        if msg != self._cola_commitmsg:
+            # needs stripping, as the empty commitmsg is saved as \n during close but initialized as ''
+            if self.commitmsg.strip() == self._cola_commitmsg.strip():
+                self.set_commitmsg(msg)
```
davvid, line 479: I wonder if we should allow for another affordance. What if the user changed the message, and they want to force the message to reload?

Maybe we should allow the user to empty the commit message and then hit Ctrl-R and that should also trigger a reload.

In order to do that, we'd just need to extend this conditional to be `if not self.commitmsg.strip() or self.commitmsg.strip() == self._cola_commitmsg.strip():`

cola/models/main.py
```diff
@@ -448,6 +450,34 @@ def _update_commitmsg(self) -> None:
...
+            self._cola_commitmsg = msg
+
```
davvid, line 481: In order to keep things tidy, it *seems* like we could move this `self._cola_commitmsg = msg` action into the `set_commitmsg()` method.

That said, this PR is also changing the merge message loading code path, and perhaps that code path needs adjustment to play nice with this idea since in theory it could reset the message twice. I'll leave a comment on that section for you to consider.

cola/models/main.py
```diff
@@ -448,6 +450,34 @@ def _update_commitmsg(self) -> None:
             self._auto_commitmsg = ''
             self.set_commitmsg(self._prev_commitmsg)
 
+        # Check for externally updated .git/GIT_COLA_MSG (e.g. written by agents/tools)
+        if not merge_msg_path and not self._auto_commitmsg:
+            self._update_cola_msg()
```
davvid: Since the two blocks above could potentially change the commit message, I wonder if we should keep these three blocks as being mutually exclusive.

In other words, instead of `if not merge_msg_path` we would use an `elif` here instead, so that we know that the code will only ever trigger for one of the three conditions above.

I haven't tested that out, but I believe that should still retain the desired behavior.

docs/git-cola.rst
```diff
@@ -445,6 +445,12 @@ The `Ctrl+i` keyboard shortcut adds a standard "Signed-off-by: " line,
 and `Ctrl+Enter` creates a new commit using the commit message and
 staged content.
 
+The commit message is saved to `.git/GIT_COLA_MSG` when `git cola` closes and
+restored from it on the next startup. In addition this file is checked on updates (Ctrl+R)
+and if modified by any tool and not conflicting with manual changes the commit
+message will be updated on the next refresh. (useful to have scripts or agents to
+prepare a message for manual review)
```
davvid: I suggest dropping the `(...)` parentheses and instead spell this out as `This feature is useful as it allows scripts or agents to prepare a message for manual review.`

davvid (changes requested):
Thanks, this is nicely done so I just have a few minor sugs before we can merge this.

While you're here, please squash the updates into the original commit and please adjust the commit message to explain why this change is being done (rather than the bullet-point style).

For the commit message conventions, I would suggest mimicking other commits. For example, the message in `git show 11b9b7bae479edde54dce650f6b8507764effb3c` is a good example of how to structure them (imperative tone, etc.) as described in `CONTRIBUTING.md`.

davvid:
I forgot to mention, please also sign-off your commits. That's `git commit -s` or the `ctrl-i` shortcut in the commit message editor.

The sign-off conveys that you adhere to the Developer's Certificate of Origin:
https://developercertificate.org/

Thanks!

davvid:
Also, please note that it's okay to  not really care about the github PR description. Those are ephemeral and do not end up in the permanent `git log` history so I try to discourage folks from wasting effort putting stuff in there that would be better conveyed via the commit messages directly.

The "CHANGES" section, for example, is not really needed because we can figure that out by looking at the diff itself.

Having good messages on the commits themselves is the way to go. They are what really matters, and they should stand on their own by explaining the "why" behind the changes (rather than the "what", which the diffs already convey).

cola/models/main.py
kurellajunior: yeah, that is a lovely idea

test/main_model_test.py
kurellajunior: unfortunately, at least on my system it is not 🤣 I had to leave it in order to run the tests 🤷

cola/models/main.py
kurellajunior: I don’t see how that could work. The `self._cola_commitmsg` should hold the last know content of the file `GIT_COLA_MSG`. If we would set that somehow during `set_commitmsg` it would loose that connection, would it not?

kurellajunior:
@davvid I addressed all but one comment (see remark)
While adding a test to see how the feature works with `_auto_codingmsg` I stumbled upon the fact that this is reset on any `set_commitmsg`. Which failed my test as the second path in autocoding was never reached. Was this a bug or did I misunderstand the code’s intention there?

I added it as a separate commit

cola/models/main.py
davvid: Makes sense, thanks for verifying and fixing that up.

PR #1580, opened 2026-04-16
prefs: per-repository GUI and icon themes

### LINKED ISSUE #1578: Feature Request: Per-Repository Theme Settings
Git-cola has per-repository settings pages, but not for appearance. Would it be possible to implement this?

I frequently open multiple instances of git-cola, and per-repo color schemes would be useful for quickly telling them apart and avoiding mistakes. I already do something like this with accent colors in obsidian.

laothrs:
Read cola.theme and cola.icontheme from the repository git config, add controls on Preferences > Current Repository, refresh appearance when preferences or worktree change, and coerce unknown theme names to default.

Document theme precedence and add unit tests for coercion and config hooks.

Closes https://github.com/git-cola/git-cola/issues/1578

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,25 @@ Git Cola is a powerful Git GUI with a slick and intuitive user interface.
 
 * [Contributing guidelines](CONTRIBUTING.md)
 
+## Themes and appearance
+
+The GUI theme is read from the git config entry `cola.theme`. Set a default under
+**Preferences → Appearance** (stored in your user config), or override it for one
+repository only via **Preferences → Current Repository** or with
+`git config cola.theme <theme-name>` inside that repository (local `.git/config`).
+Git applies the usual config precedence, so a repository value overrides your global
+default when both are set.
+
+Passing **`git cola --theme <theme-name>`** (or **`git cola gui --theme ...`**) forces
+that theme for the entire process; preference changes cannot override it until you
+exit and restart without `--theme`.
+
+Icon themes are assembled from the `GIT_COLA_ICON_THEME` environment variable
+(`:`-separated list, if set) and from one or more `cola.icontheme` config entries;
+see `get_icon_themes()` in `cola/app.py` for details. In the main Git Cola window,
+GUI theme, icon theme, and bold-fonts preference changes apply immediately; font
+family, font size, and high DPI still require a restart.
+
 
 # Requirements
 
--- a/cola/app.py
+++ b/cola/app.py
@@ -32,6 +32,7 @@ On a Debian/Ubuntu system you can install these modules using apt:
     )
     sys.exit(1)  # core.EXIT_FAILURE
 
+from qtpy import QtGui
 from qtpy import QtWidgets
 from qtpy.QtCore import Qt
 from qtpy.QtCore import Signal
@@ -66,6 +67,7 @@ from .cmd import CommandBus
 from .i18n import N_
 from .interaction import Interaction
 from .models import main
+from .models import prefs
 from .models import selection
 from .settings import Session
 from .settings import Settings
@@ -182,11 +184,17 @@ def get_icon_themes(context: ApplicationContext) -> list[str]:
 
     icon_themes_env = core.getenv('GIT_COLA_ICON_THEME')
     if icon_themes_env:
-        result.extend([x for x in icon_themes_env.split(':') if x])
+        for x in icon_themes_env.split(':'):
+            x = x.strip()
+            if x:
+                result.append(x)
 
     icon_themes_cfg = list(reversed(context.cfg.get_all('cola.icontheme')))
-    if icon_themes_cfg:
-        result.extend(icon_themes_cfg)
+    for entry in icon_themes_cfg:
+        if isinstance(entry, str):
+            s = entry.strip()
+            if s:
+                result.append(s)
 
     if not result:
         result.append('light')
@@ -199,6 +207,12 @@ class ColaApplication:
     """The main cola application
 
     ColaApplication handles i18n of user-visible data
+
+    Process model: one OS process owns a single :class:`QtWidgets.QApplication`.
+    Each shell invocation of ``git cola`` (or ``git cola --repo …``) is therefore
+    isolated from other processes. Per-repository ``cola.theme`` only affects
+    that process while its Git worktree points at that repository; switching
+    repos in the same window reapplies the merged config for the new worktree.
     """
 
     def __init__(
@@ -207,7 +221,6 @@ class ColaApplication:
         argv: list[str],
         locale: None = None,
         icon_themes: list[Any] | None = None,
-        gui_theme: None = None,
     ) -> None:
         cfgactions.install()
         i18n.install(locale)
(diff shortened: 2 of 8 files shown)


cola/app.py
```diff
@@ -182,11 +184,17 @@ def get_icon_themes(context: ApplicationContext) -> list[str]:
 
     icon_themes_env = core.getenv('GIT_COLA_ICON_THEME')
     if icon_themes_env:
-        result.extend([x for x in icon_themes_env.split(':') if x])
+        for x in icon_themes_env.split(':'):
```
davvid, line 187: `x` is a bit of a non-descript name. Maybe this should be called `icon_them`?

cola/app.py
```diff
@@ -182,11 +184,17 @@ def get_icon_themes(context: ApplicationContext) -> list[str]:
...
+                result.append(x)
 
     icon_themes_cfg = list(reversed(context.cfg.get_all('cola.icontheme')))
-    if icon_themes_cfg:
-        result.extend(icon_themes_cfg)
+    for entry in icon_themes_cfg:
+        if isinstance(entry, str):
+            s = entry.strip()
+            if s:
+                result.append(s)
```
davvid, line 197: Ditto for `s` -- can we just re-assign that to `entry`?

cola/models/prefs.py
```diff
@@ -247,6 +247,24 @@ def bold_headers(context) -> bool:
...
+    Use :func:`GitConfig.get_repo` with ``THEME`` when you need the value stored
+    only in the current repository (missing key means no repo override).
+    Use :meth:`GitConfig.get_user_or_system` for the user/system default without
+    a repo-specific entry.
+
+    Unknown or mistyped values are coerced to a registered name so removing a
+    repo-specific ``cola.theme`` or deleting a custom theme file cannot leave the
+    UI without a valid stylesheet.
+    """
+    from .. import themes
```
davvid, line 262: Please keep all imports at the top-level. If the imports cause circular imports then that's a design problem.

It seems like this import would causes a circular import dependency. This will need to be refactored to keep things tidy.

cola/models/prefs.py
```diff
@@ -247,6 +247,24 @@ def bold_headers(context) -> bool:
...
+
+    raw = context.cfg.get(THEME, default=Defaults.theme)
+    return themes.coerce_gui_theme_name(raw)
+
+
```
davvid, line 267: I haven't tried it, but it *seems* like we can solve the circular dependency by pushing this entire `def gui_theme(context)` function into the `cola.themes` module.

There's only a single call site from `cola.app` and it already imports the `themes` module so that seems like it should work.

cola/widgets/prefs.py
```diff
@@ -16,6 +16,24 @@
...
+    except IndexError:
+        return
+    icon_theme = icon_theme_widget.value()
+    if theme.name == 'default':
+        if icon_theme in ('light', 'dark'):
+            icon_theme_widget.set_value('default')
+    elif theme.is_dark:
+        if icon_theme in ('default', 'light'):
+            icon_theme_widget.set_value('dark')
+    elif not theme.is_dark:
```
davvid, line 32: should this just be an `else:` since it's a tautological check? In other words, if we get this far we already know that `not theme.is_dark` due to the check in the block above.

cola/widgets/prefs.py
```diff
@@ -243,7 +274,21 @@ def __init__(self, context, model, parent, source):
...
+
+        repo_config = {}
+        if source == 'local':
+            repo_config = {
+                prefs.THEME: (self.theme, Defaults.theme),
+                prefs.ICON_THEME: (self.icon_theme, Defaults.icon_theme),
+            }
+
         self.set_config({
+            **repo_config,
```
davvid, line 291: Please don't use `**repo_config` here. It's a bit messy, especially since we can simply call `self.set_config(..)` twice.

It seems simpler to just change the logic to use a direct call in the block above.. that also lets us avoid having to initialize en empty `repo_config`. e.g.

```python
        if source == 'local':
            self.set_config({
                prefs.THEME: ...,
                prefs.ICON_THEME: ...,
            })
```

cola/widgets/prefs.py
```diff
@@ -546,7 +582,13 @@ def __init__(self, form, parent):
         self.form = form
         self.label = QtWidgets.QLabel(
             '<center><b>'
-            + N_('Restart the application after changing appearance settings.')
+            + N_(
+                'GUI theme, icon theme, and bold-fonts apply immediately. '
+                'Restart after changing fonts, font size, or high DPI. '
```
davvid, line 587: sug: Restart *the application* after changing ...

cola/widgets/prefs.py
```diff
@@ -546,7 +582,13 @@ def __init__(self, form, parent):
...
+                'Per-repository GUI and icon themes are under '
+                'Preferences → Current Repository. '
```
davvid: Instead of the unicode arrow let's just use a `>`  sign here.

cola/app.py
```diff
@@ -220,26 +233,46 @@ def __init__(
...
+        # In normal operation this is the only QApplication; tests may replace it.
+        inst = QtWidgets.QApplication.instance()
+        if inst is not None and inst is not self._app:
+            return
+        icons.clear_icon_cache()
+        icons.install(get_icon_themes(self.context))
+        self._install_style()
 
-    def _install_style(self, theme_str: None) -> None:
+    def _install_style(self, theme_str: str | None = None) -> None:
```
davvid, line 252: This signature doesn't really make any sense anymore.

This commit rewires things so that `gui_theme=args.theme` is no longer passed to the ColaApplication constructor, and thus we are no longer calling `install_style(...)` with an argument.

We know that the `theme_str` is never specified anymore so the signature can be simplified and likewise the method can be simplified further.

cola/app.py
```diff
@@ -220,26 +233,46 @@ def __init__(
...
-        theme_stylesheet = theme.build_style_sheet(self._app.palette(), bold_fonts)
-        self._app.setStyleSheet(theme_stylesheet)
-
         is_macos_theme = theme_str.startswith('macos-')
         if is_macos_theme:
             themes.apply_platform_theme(theme_str)
         elif theme_str != 'default':
             self._app.setPalette(theme.build_palette(self._app.palette()))
+        else:
+            self._app.setPalette(QtGui.QPalette(self._baseline_palette))
```
davvid, line 271: `self._baseline_palette` is already a QPalette. Do we strictly need to make another copy here, or can we just pass `self._baseline_palette` directly?

cola/app.py
```diff
@@ -220,26 +233,46 @@ def __init__(
...
+        icons.install(get_icon_themes(self.context))
+        self._install_style()
 
-    def _install_style(self, theme_str: None) -> None:
+    def _install_style(self, theme_str: str | None = None) -> None:
         """Generate and apply a stylesheet to the app"""
         if theme_str is None:
-            theme_str = self.context.cfg.get('cola.theme', default='default')
+            argv_theme = self.context.args.theme
+            if argv_theme is not None:
```
davvid, line 256: `if argv_theme is not None:` allows empty strings to fall into this code block.

It'd be better to treat empty values as equivalent to missing by checking `if argv_theme:` instead. Empty strings evaluate to False so omitting the `is not None` clause ensures that empty strings go down the `else:` code path.

README.md
```diff
@@ -26,6 +26,25 @@ Git Cola is a powerful Git GUI with a slick and intuitive user interface.
 
 * [Contributing guidelines](CONTRIBUTING.md)
 
+## Themes and appearance
+
+The GUI theme is read from the git config entry `cola.theme`. Set a default under
+**Preferences → Appearance** (stored in your user config), or override it for one
+repository only via **Preferences → Current Repository** or with
+`git config cola.theme <theme-name>` inside that repository (local `.git/config`).
```
davvid, line 34: sug: "... inside that repository's local `.git/config`."

cola/widgets/prefs.py
```diff
@@ -243,7 +274,21 @@ def __init__(self, context, model, parent, source):
         )
         self.add_row(N_('HTTP Proxy URL'), self.http_proxy)
 
+        if source == 'local':
+            self.add_row('', QtWidgets.QLabel())
+            self.add_row(N_('GUI theme (this repository)'), self.theme)
+            self.add_row(N_('Icon theme (this repository)'), self.icon_theme)
```
davvid, line 280: Let's omit the ` (this repository)` suffixes here. It's already clear that all of these settings only apply to the current repository since this tab is already marked as being so, and it'd be weird to call out *just* these values.

Another reason we want to avoid new strings is that these strings don't have any translations (`cola/i18n/*.po`) so they won't be localized for users that have another language enabled.

cola/widgets/prefs.py
```diff
@@ -201,6 +219,19 @@ def __init__(self, context, model, parent, source):
...
 
+        if source == 'local':
+            self.themes = themes.get_all_themes()
+            self.theme = qtutils.combo_mapped(themes.options(themes=self.themes))
+            self.icon_theme = qtutils.combo_mapped(icons.icon_themes())
+            repo_theme_tip = N_(
+                'Override appearance for this repository only. '
+                'Matches the effective theme until you change it here. '
+                'GUI and icon themes apply immediately in the main window. '
+                'The command-line --theme option overrides all config when set.'
```
davvid, line 230: sug: "The command-line --theme option overrides all configuration when specified."

cola/widgets/prefs.py
```diff
@@ -546,7 +582,13 @@ def __init__(self, form, parent):
         self.form = form
         self.label = QtWidgets.QLabel(
             '<center><b>'
-            + N_('Restart the application after changing appearance settings.')
```
davvid, line 549: I would suggest keeping the original string, "Restart the application after changing appearance settings" and then the additional details below can be specified as a separate `N_(...)` string with a space (`' '`) in-between.

The string below can omit the, "Restart after changing fonts, font size, or high DPI." sentence since it will already have been covered by the (restored) string above. It can instead call out the fact that GUI/icon theme and bold-fonts apply immediately without needing to call out all of the ones that don't.

The plus side of keeping the original string in there is that users will at least get a partially translated UI rather than one that's completely untranslated.

cola/widgets/prefs.py
```diff
@@ -546,7 +582,13 @@ def __init__(self, form, parent):
...
+            + N_(
+                'GUI theme, icon theme, and bold-fonts apply immediately. '
+                'Restart after changing fonts, font size, or high DPI. '
+                'Per-repository GUI and icon themes are under '
+                'Preferences → Current Repository. '
+                'The --theme command-line flag overrides all theme config when used.'
```
davvid, line 590: Same sug here for this line: "The --theme command-line option overrides all configuration when specified."

Also note that this string was referring to "flag" whereas the other strings were referring to it as an "option." The sug is more consistent by referring to it as an "option."

davvid (changes requested):
Please address these notes and squash the pre-commit style tweaks into the commit so that there aren't any bot commits in here.

This is looking like it's on the right path and just needs a few tweaks before we can merge. Thanks

davvid:
This was implemented in #1623

PR #1581, opened 2026-04-18
dag: Add branch labels to inline graph

p12tic:
Before:

Screenshot_20260418_163534

After:

Screenshot_20260418_163433

The style is similar to gitk which I found rather convenient over the years.

Didn't update changelog because this is part of not yet released dag graph work, so existing release note covers it well.

Remaining work in this area:
 - solve the issue of a branch with long name being present locally and on one or more remotes. If all these refs point to the same commit, there's no space for commit message. Solved this with grouping related refs and collapsing branch names to `...` in all except one. Will create a new PR.
 - make graph itself more tidy
 - HEAD should change commit circle color instead of being a label
 - hide long lines in the graph like gitk does
 - fix color opaqueness noise in the graph lines (most likely due to line overlap)
 - make the commit list more condensed to improve information density. Will create new PR.

### DIFF
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -2,6 +2,7 @@ import collections
 import itertools
 import math
 from functools import partial
+from typing import Optional
 
 from qtpy import QtCore
 from qtpy import QtGui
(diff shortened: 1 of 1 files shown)


p12tic:
I'm sorry, I didn't notice that tests are broken. Fixed now

davvid:
Thanks! This is awesome. I'm planning on merging this soon.

davvid:
One issue I'm seeing is that when you do `View -> Display Inline Graph` it hides the entire summary rather than just the graph itself.

This is because the code is setting the column as hidden, and that's hiding the summary as well as the custom graph rendering. We'll need to get that working before merging.

davvid:
Got it -- I just had to update the code to toggle use of the `self.graph_delegate` instead of hiding the entire column. Works great now!

p12tic:
Thanks! I'm sorry for not noticing the problem myself

PR #1585, opened 2026-04-23
dag: Convert HEAD label to a commit circle color

p12tic:
Before:

After:

### DIFF
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -25,6 +25,7 @@ from ..i18n import N_
 from ..models import dag
 from ..models import main
 from ..models import prefs
+from ..models.graph import GraphRowColor
 from ..models.graph import build_graph
 from ..qtutils import get
 from . import archive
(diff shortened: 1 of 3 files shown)


cola/widgets/dag.py
```diff
@@ -1102,16 +1101,20 @@ def add_commits(self, commits):
...
         items = []
+        head_oid = None
         for c in reversed(commits):
             item = CommitTreeWidgetItem(c)
             items.append(item)
             self.oidmap[c.oid] = item
             for tag in c.tags:
                 self.oidmap[tag] = item
+            if 'HEAD' in c.tags:
+                head_oid = c.oid
```
davvid, line 1112: This is starting to get expensive because we're now looping over `c.tags` twice.

There's a loop immediately above where we're already looping over all of the "tags" markers. How about adding a check to that loop?:

```python
        for tag in c.tags:
            self.oidmap[tag] = item
            if tag == 'HEAD':
                head_oid = c.oid
```

davvid (changes requested):
This looks great. I had one small note so I just went ahead and fixed that up in 8d537ffa096e06b91f6c4db0753e55511b015692.

davvid:
Maybe because I use dark mode (and it's evening so the Night Mode red-shift display color is active) but the yellow color still seems a bit subtle for denoting the HEAD commit.

What do you think about making it green as well? That would also make it match the branch label color (for better or worse). It seems like a green circle might be a little easier to spot visually.

i'm curious about your opinion on that so I won't change that just yet...

p12tic:
I use dark mode as well :-) Maybe let's use red? That would be completely unambiguous.

davvid:
For now I'll just leave it as-is.. it's growing on me. I do like that it's consistent with the rectangular `HEAD` label. Red sometimes implies an error, but the gold color is actually pretty nice considering the alternatives (light blue, green, etc).

PR #1589, opened 2026-04-26
dag: Reduce spacing to increase information density

p12tic:
Before:
before

After:
after


davvid:
There was minor merge conflict in `cola/widgets/dag.py` after merging #1588 but it seems to have been resolved correctly. Thanks!

PR #1590, opened 2026-04-28
graph: improve handling for graphs with gaps

### LINKED ISSUE #1586: misrendered DAG graph with files in command-line
This happens when there's a filter containing filenames in the text input box above "Summary".

Image

davvid:
Running the DAG over the history of a file, e.g. "git dag garden.yaml", can result in a graph where every node allocates a new column. The root cause is that we will receive a list of commits that contain gaps, as the history has been simplified and is truncated due to the pathspec filter.

Detect when commits exist whose parents are not present in the graph. Reset the commit columns for these commits. This has the net effect of preventing the issue where the graph continually grows to the right.

Closes: #1586

### DIFF
--- a/cola/models/graph.py
+++ b/cola/models/graph.py
@@ -45,14 +45,23 @@ def build_graph(
     rows: list[GraphRow] = []
     max_columns = 0
 
+    all_oids = {commit_and_parents[0] for commit_and_parents in commits}
+
     # The graph is built top-to-bottom (newest first), so the input is reversed.
     for oid, parent_oids in reversed(commits):
+        # Is this a terminal commit without any parents?
+        terminal_commit = False
         # Find the commit in active_lanes or allocate a new lane.
         try:
             commit_column = active_lanes.index(oid)
         except ValueError:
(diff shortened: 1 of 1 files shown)


davvid:
Heads-up @p12tic ~ it'd be great to get your opinion on whether this is a viable implementation strategy for handling #1586.

To reproduce the issue, try these invocations:

```
./bin/git-dag garden.yaml
./bin/git-dag Makefile
```

p12tic (approved):

p12tic:
> it'd be great to get your opinion on whether this is a viable implementation strategy for handling

I will need to think more about this, because this use case seems to be a major oversight by me. Surely it's good enough as a workaround.

By the way, there build_graph has unit tests, so it makes sense to add more of them for this case

davvid:
Thanks for reviewing @p12tic. I'll proceed with the merge and I'll definitely keep an eye out for whether you have a better deeper solution in the future. In the very least, this stopgap workaround should be good enough for the initial release and then we'll keep refining it over time.

PR #1596, opened 2026-05-18
Fix clipboard clearing on MacOS when closing git-cola

joshuataylor:
Qt's QClipboard on MacOS does not always hand a persistent copy of the string to the system pasteboard, so anything copied through Copy Path / Copy Relative Path / etc. is lost when git-cola exits. The existing persist_clipboard() helper only addresses the X11 case.

Write the string through [AppKit's NSPasteboard](https://developer.apple.com/documentation/appkit/nspasteboard) on MacOS so the pasteboard
server retains its own copy and the clipboard survives app exit.

Tried to copy the same approach from cola/themes.py

### DIFF
--- a/cola/qtutils.py
+++ b/cola/qtutils.py
@@ -19,6 +19,11 @@ from qtpy import compat
 from qtpy.QtCore import Qt
 from qtpy.QtCore import Signal
 
+try:
+    import AppKit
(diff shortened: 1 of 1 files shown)


davvid:
Thanks!

PR #1599, opened 2026-05-19
Fix the macOS menu bar to show "Git Cola" instead of "python3"

joshuataylor:
When git-cola is launched from a non-bundled Python (running the `git-cola` script via a system `python3`, or via Homebrew's keg-only `python`, rather than from a `.app` bundle), the macOS menu bar shows the interpreter's executable name — `python3`, `python`, or `python3.14` — in the Apple menu, in the Hide / Quit items, and in the dock tile. Cocoa derives all of those from the running process when no bundle is present.

It's amazing how weird Apple is for running apps, I thought this would be a simple fix.. GitHub code search got a workout with this one to see how others fix things...

Anyway, three things need to be done BEFORE `QApplication` is constructed so Qt's Cocoa plugin uses the right name when it builds the menu:

- [`QCoreApplication.applicationName`](https://doc.qt.io/qt-6/qcoreapplication.html) — used for Qt menus.
- [`CFBundleName`](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundlename) / `CFBundleDisplayName` in `[[NSBundle mainBundle] infoDictionary]` — read by Qt to populate the macOS menu bar (title slot, Hide / Quit items).
- [`NSProcessInfo.processName`](https://developer.apple.com/documentation/foundation/processinfo/processname) — used for the dock tile and Apple-menu app title.

Verified locally on macOS 15.7.2 / arm64 / Qt 6.11.0 / PyQt6 6.11.0: after this change the menu bar reads "Git Cola", the Apple menu reads "About git-cola" / "Hide Git Cola" / "Quit Git Cola", and the dock tile reads "Git Cola" on hover.

For homebrew git-cola, before:

image

After:

image

(I'll fix the Quit Menu item, which is already not working, later).

### DIFF
--- a/cola/app.py
+++ b/cola/app.py
@@ -194,6 +194,60 @@ def get_icon_themes(context: ApplicationContext) -> list[str]:
     return result
 
 
+def _set_application_name() -> None:
+    """Tell Qt and Cocoa the user-visible application name is "Git Cola".
+
+    Without this, a non-bundled launch on macOS (running ``git-cola`` via the
+    system ``python3``) shows ``python3`` in the menu bar, the Apple menu
+    "Hide" / "Quit" items, and in the dock, because Cocoa derives all of
+    those from the executable name.
+
+    Cocoa reads three different sources for the user-visible app name and
+    they have to be patched in lock-step:
+
+    * Qt menus: ``QCoreApplication.applicationName``
+    * macOS menu bar (title slot, Hide/Quit items): the ``CFBundleName`` key
+      of ``[[NSBundle mainBundle] infoDictionary]``, which Qt's Cocoa plugin
+      reads when it builds the menu.
+    * Dock tile / ``NSRunningApplication.localizedName``: the Cocoa process
+      name, set via ``-[NSProcessInfo setProcessName:]``.
+
+    Must run before ``QApplication`` is constructed as Qt snapshots the bundle dictionary the first time it builds the
+    Cocoa menu.
+
+    References:
+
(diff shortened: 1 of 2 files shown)


joshuataylor:
Sorry, this should have been a draft, just finishing tests/manual testing :)

cola/app.py
```diff
@@ -194,6 +194,60 @@ def get_icon_themes(context: ApplicationContext) -> list[str]:
...
+        # nil; replace it via the private _infoDictionary ivar fallback only
+        # if the public mutator is unavailable.
+        bundle = NSBundle.mainBundle()
+        info = bundle.localizedInfoDictionary() or bundle.infoDictionary()
+        if info is not None:
+            # NSBundle's info dict is an NSMutableDictionary in practice;
+            # setObject_forKey_ works even when the public API claims it is
+            # read-only.
+            info.setObject_forKey_('Git Cola', 'CFBundleName')
+            info.setObject_forKey_('Git Cola', 'CFBundleDisplayName')
```
davvid, line 245: Nicely done! I'd suggest passing `context` to this function so that we can use `context.app_name` to get the `Git Cola` string.

When running git-dag, context.app_name is set to `Git DAG`.

joshuataylor: Amazing, didn't know this! Will take a look.

davvid:
I did some testing and applied the sug in 3e83cc96808a383e10bea2dac998eb45a17a9b9 and it's looking good to me so I went ahead and merged. Thanks again, really appreciated these enhancements.

PR #1600, opened 2026-05-19
Fix maximized window state on restore on macOS

joshuataylor:
On macOS 15, calling `QMainWindow.restoreGeometry()` with a blob whose "maximized" flag is set sends Qt into a layout-feedback loop on relaunch: the window grinds creating and destroying tab bars and dock-widget children until it eventually settles or hangs.

How to reproduce on MacOS:

1. Launch git-cola in a fresh profile (e.g move $HOME/.config/git-cola somewhere else, so it doesn't break your current settings)
2. Make the window maximised (NOT FULL SCREEN!)
3. Close git-cola
4. Reopen git-cola, and it should now "freeze"

With a clean `~/.config/git-cola` the relaunch seems to be stuck in widget construction, using `py-spy dump` shows the GUI thread in `QApplication.exec_()`, with `qt.widgets.showhide` events showing.

This is the latest in a long-running upstream chain: [QTBUG-4397](https://bugreports.qt.io/browse/QTBUG-4397), [QTBUG-21371](https://bugreports.qt.io/browse/QTBUG-21371), [QTBUG-106678](https://bugreports.qt.io/browse/QTBUG-106678), [QTBUG-123335](https://bugreports.qt.io/browse/QTBUG-123335) (`save/restoreGeometry behaves incorrectly when application is maximized`, still open and affecting macOS + Windows).

So, this fixes it for me, but I would love to get some more testing, especially around the ridiculous flag bytes.

In `cola/widgets/standard.py`, we parse the `saveGeometry()` header, stripping the maximized flag byte (offset 44 in the [documented binary format](https://doc.qt.io/qt-6/qwidget.html#saveGeometry)), and re-apply `Qt.WindowMaximized` via `QTimer.singleShot(0, ...)` so it lands AFTER Cocoa has placed the window and the event loop is running. This breaks the feedback loop while preserving the user's maximized state across launches.

New unit tests in `test/widgets_standard_test.py` cover the strip helper across real captured `saveGeometry()` blobs (maximized and non-maximized), plus edge cases (junk bytes, truncated blob, the observed-in-practice `0x02` flag value that isn't documented as the canonical "true"). I have very little experience with Qt, so please feel free to change anything.

### DIFF
--- a/cola/widgets/standard.py
+++ b/cola/widgets/standard.py
@@ -22,6 +22,40 @@ from ..settings import Settings
 from ..settings import mklist
 from . import defs
 
+# Magic header emitted by QWidget::saveGeometry() (Qt 5/6).
+_QWIDGET_GEOMETRY_MAGIC = b'\x01\xd9\xd0\xcb'
+# Offset of the "maximized" flag byte inside a saveGeometry() blob:
+#   4 magic + 2 major + 2 minor + 16 frame QRect + 16 normal QRect + 4 screen
+_QWIDGET_GEOMETRY_MAXIMIZED_OFFSET = 44
+
+
+def _strip_maximized_geometry_flag(blob):
+    """Return (blob_without_maximized_flag, was_maximized).
+
+    QWidget::saveGeometry() encodes the maximized state as a single byte at a
+    well-known offset. Strip it so restoreGeometry() does not re-apply the
+    maximized state during widget construction; the caller is responsible for
+    re-applying it after show() has been called.
+
+    Cf. https://doc.qt.io/qt-6/qwidget.html#saveGeometry
+    """
+    # The QByteArray returned by fromBase64 supports the bytes-like protocol,
+    # so bytes(blob) gives us a plain Python bytes object we can index into.
+    raw = bytes(blob)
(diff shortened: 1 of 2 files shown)


davvid:
Nice, this is hacky, but definitely worth it. To make things slightly safer, I'd suggest adding an early-out for non-darwin platforms. That way we'll only do this on the affected macOS platform.

joshuataylor:
Hacky is an understatement. 😂

I wonder if we can make a note somewhere to follow up on the Qt issues, I made a note of them here for future maintainers (hello!).

davvid:
I went ahead and merged this as-is since it does seem to at least improve the behavior for now.

I did notice issues with the Toolbars while testing. The toolbars would disappear when shutting down and restoring in the fullscreen state.

It turns out that this was a pre-existing issue, though, and 479ab8ebbda7d73213e60fe4d2a6d62ca513a0eb seems to have prevented that issue.

I then relegated the workaround to just macOS in 7e71026c50e1b03cab5c276f90b5f24353c9036a which will at least limit the blast radius in case the workaround has adverse effects on other platforms.

Thanks!

PR #1601, opened 2026-05-23
fix: auto-detect system dark/light theme for default icon selection

aruzaphoenix:


cola/app.py
```diff
@@ -176,6 +176,37 @@ def _get_askpass() -> TextType:
...
 
 
+def _detect_system_theme() -> str:
+    """Detect if the system is using a dark or light theme.
+
+    Uses Qt 6.5+ colorScheme() if available, falls back to
+    QPalette lightness comparison for older Qt versions.
+    """
+    from qtpy.QtWidgets import QApplication
+    from qtpy.QtGui import QPalette
```
ElementW, line 186: Already imported at the top

cola/app.py
```diff
@@ -189,11 +220,65 @@ def get_icon_themes(context: ApplicationContext) -> list[str]:
         result.extend(icon_themes_cfg)
 
     if not result:
-        result.append('light')
+        result.append(_detect_system_theme())  # auto-detect instead of hardcoding 'light'
 
     return result
 
 
+def _detect_system_theme() -> str:
```
ElementW, line 228: Defined twice?! And the implementations differ

cola/app.py
```diff
@@ -176,6 +176,37 @@ def _get_askpass() -> TextType:
...
+    """Detect if the system is using a dark or light theme.
+
+    Uses Qt 6.5+ colorScheme() if available, falls back to
+    QPalette lightness comparison for older Qt versions.
+    """
+    from qtpy.QtWidgets import QApplication
+    from qtpy.QtGui import QPalette
+
+    app = QApplication.instance()
+    if app is not None:
```
ElementW, line 189: 1. Precondition is this is called from a context where it's not `None`, let it `TypeError` away
2. Nice way to indent a whole block for no reason when the alternate path is a `return`

ElementW (commented):
What in the vibecoded slopfest is this

Do people really not use their brains any more nowadays?

PR #1604, opened 2026-06-07
Display remote URL alongside remote name in push/fetch/pull dialog

### LINKED ISSUE #1602: Suggest: optimize the remote display of push
As shown in figure, replace 'origin' to 'origin (git@github.com:user/repo.git)'.
Image

suchismitab0511:
Fixes #1602

When using the Push, Fetch, or Pull dialog, the remote list only 
showed the remote name (e.g. "origin"). This made it hard to 
identify which remote corresponds to which repository, especially 
when multiple remotes are configured.

This change updates the remote list to show the remote name along 
with its URL in parentheses.

Before: origin
After: origin (https://github.com/user/repo.git)

The URL is fetched using the existing `model.remote_url()` method, 
so no new logic was needed.
Screenshot 2026-06-07 210503

### DIFF
--- a/cola/widgets/remote.py
+++ b/cola/widgets/remote.py
@@ -181,7 +181,11 @@ class RemoteActionDialog(standard.Dialog):
         if action == PUSH:
             mode = QtWidgets.QAbstractItemView.ExtendedSelection
             self.remotes.setSelectionMode(mode)
-        self.remotes.addItems(model.remotes)
+        displayed_remotes = []
+        for remote_name in model.remotes:
+            url = model.remote_url(remote_name, action)
+            displayed_remotes.append(f'{remote_name} ({url})')
+        self.remotes.addItems(displayed_remotes)
 
         self.remote_branch_label = QtWidgets.QLabel()
         self.remote_branch_label.setText(N_('Remote Branch'))


davvid:
One downside to this approach is that it makes the display very busy and noisy, especially when you have dozens of remotes. Overall, I think this actually decreases usability due to the information overload.

The remote *names* are actually intended to be the human-friendly side of the interface. That's why we register URLs behind remote names -- names such as `upstream` are supposed to convey more meaning than just the URL.

Does this need to be front and center?

A related question -- have you considered giving your remotes better names so that you don't need to look at the URLs?


Alternatives to consider:

- When remotes are selected, we could display a short table after the example `git push` / `git fetch` commands for all of the selected remotes. This table could basically be a list of  `remote: <url>` lines that are displayed.

- Another idea would be that, instead of increasing the visual noise for everyone all the time, we can add tooltips to the items. That way if someone is unsure of what the remote implies then they can hover their mouse over the remote and it'll display the output of `git remote show -n <remote>` in a tooltip.


Would a tooltip work in your scenario?

davvid:
This was not merged but the spirit of #1602 was resolved by adding a `Remotes` section to the command display and by adding tooltips with details from `git remote show -n <remote>`.

PR #1609, opened 2026-06-26
Support copying multiple files at once

joshuataylor:
Copying multiple files at once in the status window only copies the first file name, and I've been annoyed long enough by this quirk that I decided to fix it. :-)

Happy for this to be changed however you want. Not fussed, just would like a solution :P.

Users can select multiple files at once, but trying to copy multiple files via either a shortcut (CTRL+C/CMD+C) or via Right Click -> `Copy` -> `Copy Path to Clipboard`/`Copy Relative Path to Clipboard`/`Copy Leading Path to Clipboard`/`Copy Basename to Clipboard` only copies the first file.

image

Instead, we can use `context.selection.group()` to get all the files in copy_path/copy_basename/copy_leading_path/copy_format, and split by new line (`\n`).

> Single-file selections still produce a single line with no trailing newline.

### DIFF
--- a/cola/widgets/main.py
+++ b/cola/widgets/main.py
@@ -731,6 +731,7 @@ class MainView(standard.MainWindow):
             self.diffeditor,
             self.bookmarkswidget.tree,
             self.recentwidget.tree,
+            self.statuswidget.tree,
         )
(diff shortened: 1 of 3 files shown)


davvid:
Thanks!

PR #1610, opened 2026-06-26
Fix focus always going to the start of the line instead of the end

joshuataylor:
When opening git-cola with a commit message already filled, when pressing CMD+L to go to the commit title, the cursor goes to the start of the line instead of the end, but only for the first time you do this.

Going back to status window for example, then pressing CMD+L again, the cursor goes to the end of the line.

Screen recording:
https://github.com/user-attachments/assets/3b560384-eba6-48aa-9bd7-ec63f23838a1

### DIFF
--- a/cola/widgets/spellcheck.py
+++ b/cola/widgets/spellcheck.py
@@ -105,12 +105,21 @@ class SpellCheckLineEdit(SpellCheckTextEdit):
         """Select text when entering with a tab to mimic QLineEdit"""
         super().focusInEvent(event)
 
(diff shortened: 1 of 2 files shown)


joshuataylor:
I've added some tests as well, I'm not 100% across Qt/PyQt6, but I read the documentation and checked the other tests. Happy for those to be changed as you want.

I read: https://doc.qt.io/qt-6/qfocusevent.html
https://www.riverbankcomputing.com/static/Docs/PyQt6/index.html

PR #1612, opened 2026-06-29
Improve Gravatar (use sha256, fix resizing of avatars)

joshuataylor:
Scrolling through commits in `DAG` would show a delay and "flicker" when Gravatar is enabled.

I have been doing a fair bit of git history spelunking in the last week or so, and this was annoying me :-).

Before:

https://github.com/user-attachments/assets/7f81c899-46a5-48bc-831c-5203de8de7f3

After:

https://github.com/user-attachments/assets/59e52c73-d4a5-4992-992f-db61bd8feee5

--

After going down this new rabbithole, I also found that Gravatar changed their avatar lookup to SHA256, which means we can now simplify the lookup, see: https://docs.gravatar.com/rest/hash .

I then started testing across various repos, finding weird bugs, mostly due to pixmap misunderstandings, so I tried to create a test for every example I found.

---

I then was getting annoyed by the rendering of the avatars, and the "flickering"/loading, so I learnt more about pixmaps, and use that as the default.

As a result, scrolling through the DAG view is now much nicer, IMHO, as it doesn't nudge about the layout.

---

I've split the work into three commits, happy for this to be split into smaller work, as always - take whatever you think is useful for the commits (none, some, all, I don't mind), rework how you see fit.

I'm going to continue testing this to ensure I haven't created any weird bugs, hence the draft.

### DIFF
--- a/cola/gravatar.py
+++ b/cola/gravatar.py
@@ -8,6 +8,7 @@ from qtpy import QtCore
 from qtpy import QtGui
 from qtpy import QtNetwork
 from qtpy import QtWidgets
+from qtpy.QtCore import Qt
 
 from . import core
 from . import icons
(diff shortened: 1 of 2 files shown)


davvid:
This has been working great in my testing so I went ahead and merged all of the pending changes. If no other topics come in over the weekend I'll plan to release the next version early in the week.

Thanks!

PR #1613, opened 2026-06-29
diff: debounce DAG diff loading and drop superseded results

joshuataylor:
"Scrolling" (e.g using up/down arrow keys) commit-to-commit in the DAG felt sluggish because every selection change synchronously kicked off "git log %b" + "git diff" for the commit passed over, even while holding an arrow key through dozens of commits.

CommitDiffWidget now updates the metadata labels immediately (so the selection stays responsive) but defers the diff load behind a 100ms single-shot timer, so only the commit the selection settles on triggers git.

Each started diff task is stamped with a monotonic token; set_diff drops a result whose token is no longer current, so a slow diff finishing after the selection has moved on cannot overwrite the diff for the current commit. (Empty selection cancels any pending load.)

---

Before:

https://github.com/user-attachments/assets/94b67679-c806-46bb-bcd7-abd11b7ba49b

After:

https://github.com/user-attachments/assets/61ebdeac-4bf1-4387-862e-3e4ab3a388f5

IMHO, the debounce could be a setting, as it's really noticeable for larger repositories.

### DIFF
--- a/cola/widgets/diff.py
+++ b/cola/widgets/diff.py
@@ -1947,6 +1947,11 @@ class AuthorLabel(PlainTextLabel):
 class CommitDiffWidget(QtWidgets.QWidget):
     """Display commit metadata and text diffs"""
 
+    # Delay before a selected commit's diff is loaded. Holding an arrow key
+    # in the DAG steps through commits faster than this, so intermediate
+    # commits never trigger a "git diff"; only the commit we land on does.
+    DIFF_DEBOUNCE_MSEC = 100
+
     def __init__(self, context, parent, is_commit=False, options=None):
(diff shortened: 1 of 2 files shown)


davvid:
Thanks, this is a really nice strategy with the diff tokens. This looks pretty good for merging. I'll plan to take a look in about 24 hours in case there's other small details you might change in the meantime.

The 100 ms time could be configurable per-repo; 100 or 125 does seem like a good default, though. Once this is merged I could look into adding a `cola.dag.diffinterval` variable for configuring it for repos where large diffs are common.

Part of me wants to believe that it doesn't need to be configurable, though. We could try releasing it without adding configuration and then see if it ever becomes an issue in practice. I'm slightly inclined to go down that route.

Issue #1540, opened 2026-01-12
Feature request / discussion: Working with repositories on another machine


p12tic:
Thanks for amazing piece of software, I've been using it for more than a decade!

I have the following use case that I would like to discuss and, if approved, implement.

git-cola currently can work only on repositories that reside on local machine. It cannot connect to a different machine via ssh, run git commands there and view the results on local GUI.

This feature is useful when doing remote development, that is, connecting from a weak local machine to a big powerful server, having the code there and using the server for running all the development commands such as tests. IDEs such as vscode/vscodium supports this approach quite well. Unfortunately vscode works by running everything on the server, not even having the code locally. This makes it very hard to impossible to use git-cola. The repository is too large for sshfs to be fast enough for local git operations.

I would like to propose git-cola support this approach as well. git-cola could connect to the machine via ssh and run git commands and any needed file access there.

Using separate ssh connections for each interaction would be very slow, so while this could be a good first iteration, the final implementation should be different. One solution would be to do something like vscode does. git-cola could upload small python script which could listen for instructions on stdin and return results over stdout. The instructions would be whatever operation git-cola needs to run locally right now.

Would this be something that maintainers think is useful to have in git-cola?

davvid:
I do think this feature would be a pretty killer feature in general. I do like the idea of initiating the connection over ssh to minimize the security concerns. The remote worktree could also provide a REST API, but that brings along a bunch of security issues that we can offload if we kept the ssh connection alive and used it as a bidirectional pipe.

A REST API on the remote side could work as well if we locked it down to require a secret random key that is generated when the initial ssh connection tunnels in and launches a cola server. All the usual aspects of distributed systems would come into play (gracefully handling network timeouts, disconnections, etc).

This would definitely fall in the camp of being a large / huge feature given how much of the code assumes that it can just access the filesystem and git directly. Not impossible; it's just a lot of interactions that would need to be adapted.

This is where using something like Rust for the server component seems appealing simply because we could ship a lot of functionality over to the remote host using a single binary. We could use Python as well if we were to put the burden of ensuring that all of the python modules are installed and available on the remote side, but Python doesn't really shine in that area given all of the complications around needing to make a Python interpreter, all of the dependencies, and all of the cola server code available on the remote. A single binary really is nice in that respect.

[...]

PR #1617, opened 2026-07-14
cola: implement basic server/client mode

Ahrirg:
This PR adds basic client/server support to `git-cola` using WebSockets. [Feature request](https://github.com/git-cola/git-cola/issues/1540)

## Usage

Start the server:

```bash
python -m cola.server
```

Connect a client to a remote repository:

```bash
python -m cola connect <ip:port> </path/to/repo>
```

## Current Limitations
- `fsmonitor.py` is not yet implemented for remote mode. As a result, any features or UI actions that depend on the `fsmonitor.py` are currently disabled when connected to a remote server.

### DIFF
--- a/cola/app.py
+++ b/cola/app.py
@@ -159,8 +159,8 @@ def _get_askpass(ops: operations.IOperations) -> TextType:
     if sys.platform == 'darwin':
         return resources.package_command('ssh-askpass-darwin')
 
-    kde_askpass = core.find_executable('ksshaskpass')
-    gnome_askpass = core.find_executable('gnome-ssh-askpass')
+    kde_askpass = ops.find_executable('ksshaskpass')
+    gnome_askpass = ops.find_executable('gnome-ssh-askpass')
     if gnome_askpass is None:
         gnome_askpass = '/usr/lib/openssh/gnome-ssh-askpass'
 
--- a/cola/app.py
+++ b/cola/app.py
@@ -90,10 +90,12 @@ def setup_environment() -> None:
     # Spoof an X11 display for SSH
     os.environ.setdefault('DISPLAY', ':0')
 
+    ops = operations.LocalOperations()
+
     if not core.getenv('SHELL', ''):
         for shell in ('/bin/zsh', '/bin/bash', '/bin/sh'):
             if os.path.exists(shell):
-                compat.setenv('SHELL', shell)
+                compat.setenv(ops, 'SHELL', shell)
                 break
 
     # Setup the path so that git finds us when we run 'git cola'
@@ -101,19 +103,19 @@ def setup_environment() -> None:
     bindir = core.decode(os.path.dirname(sys_argv0))
     path_entries.append(bindir)
     path = os.pathsep.join(path_entries)
-    compat.setenv('PATH', path)
+    compat.setenv(ops, 'PATH', path)
 
     # We don't ever want a pager
-    compat.setenv('GIT_PAGER', '')
+    compat.setenv(ops, 'GIT_PAGER', '')
 
     # Setup the openssh askpass credentials helper.
-    askpass = _get_askpass(operations.LocalOperations())
-    compat.setenv('GIT_ASKPASS', askpass)
-    compat.setenv('SSH_ASKPASS', askpass)
+    askpass = _get_askpass(ops)
+    compat.setenv(ops, 'GIT_ASKPASS', askpass)
+    compat.setenv(ops, 'SSH_ASKPASS', askpass)
 
     # Avoid taking optional locks for read-only operations.
     # We assume that proper state-modifying operations continue to take locks.
-    compat.setenv('GIT_OPTIONAL_LOCKS', '0')
+    compat.setenv(ops, 'GIT_OPTIONAL_LOCKS', '0')
 
     # --- >8 --- >8 ---
     # Git v1.7.10 Release Notes
@@ -145,7 +147,7 @@ def setup_environment() -> None:
     # --- >8 --- >8 ---
     # Longer-term: Use `git merge --no-commit` so that we always
     # have a chance to explain our merges.
-    compat.setenv('GIT_MERGE_AUTOEDIT', 'no')
+    compat.setenv(ops, 'GIT_MERGE_AUTOEDIT', 'no')
 
 
 def _get_askpass(ops: operations.IOperations) -> TextType:
@@ -159,8 +161,8 @@ def _get_askpass(ops: operations.IOperations) -> TextType:
     if sys.platform == 'darwin':
         return resources.package_command('ssh-askpass-darwin')
 
-    kde_askpass = ops.find_executable('ksshaskpass')
-    gnome_askpass = ops.find_executable('gnome-ssh-askpass')
+    kde_askpass = core.find_executable('ksshaskpass')
+    gnome_askpass = core.find_executable('gnome-ssh-askpass')
     if gnome_askpass is None:
         gnome_askpass = '/usr/lib/openssh/gnome-ssh-askpass'
 
@@ -401,7 +403,7 @@ def process_args(args: argparse.Namespace, ops, setup_repo: bool = False) -> Non
         sys.exit(core.EXIT_SUCCESS)
 
     # Handle session management
-    restore_session(ops, args)
+    restore_session(args)
 
     # Initialize args.repo. If a repository was specified as a
     # positional argument then we will use that value.
@@ -428,7 +430,7 @@ def process_args(args: argparse.Namespace, ops, setup_repo: bool = False) -> Non
         sys.exit(core.EXIT_USAGE)
 
 
-def restore_session(ops, args: argparse.Namespace) -> None:
+def restore_session(args: argparse.Namespace) -> None:
     """Load a session based on the window-manager provided arguments"""
     # args.settings is provided when restoring from a session.
     args.settings = None
@@ -747,7 +749,7 @@ def initialize() -> str:
     ops: operations.IOperations = operations.LocalOperations()
     git_path = find_git(ops)
     if git_path:
-        prepend_path(git_path)
+        prepend_path(ops, git_path)
 
     # The current directory may have been deleted while we are still
     # in that directory.  We rectify this situation by walking up the
@@ -926,13 +928,13 @@ def find_git(ops) -> str | None:
     return None
 
 
-def prepend_path(path) -> None:
+def prepend_path(path, ops) -> None:
     """Adds git to the PATH.  This is needed on Windows."""
     path = core.decode(path)
     path_entries = core.getenv('PATH', '').split(os.pathsep)
     if path not in path_entries:
         path_entries.insert(0, path)
-        compat.setenv('PATH', os.pathsep.join(path_entries))
+        compat.setenv(ops, 'PATH', os.pathsep.join(path_entries))
 
 
 def detect_system_theme() -> str:
--- a/cola/app.py
+++ b/cola/app.py
@@ -390,12 +390,16 @@ class ColaQApplication(QtWidgets.QApplication):
         sid = session_mgr.sessionId()
         skey = session_mgr.sessionKey()
         session_id = f'{sid}_{skey}'
-        session = Session(session_id, repo=core.getcwd())
+        session = Session(session_id, repo=self.context.ops.getcwd())
         session.update()
         view.save_state(settings=session)
 
 
-def process_args(args: argparse.Namespace, ops, setup_repo: bool = False) -> None:
+def process_args(
+    ops: operations.IOperations,
+    args: argparse.Namespace,
+    setup_repo: bool = False,
+) -> None:
     """Process and verify command-line arguments"""
     if args.version:
         # Accept 'git cola --version' or 'git cola version'
@@ -403,7 +407,7 @@ def process_args(args: argparse.Namespace, ops, setup_repo: bool = False) -> Non
         sys.exit(core.EXIT_SUCCESS)
 
     # Handle session management
-    restore_session(args)
+    restore_session(ops, args)
 
     # Initialize args.repo. If a repository was specified as a
     # positional argument then we will use that value.
@@ -426,17 +430,17 @@ def process_args(args: argparse.Namespace, ops, setup_repo: bool = False) -> Non
             )
             % repo
         )
-        core.print_stderr(errmsg)
+        ops.print_stderr(errmsg)
         sys.exit(core.EXIT_USAGE)
 
 
-def restore_session(args: argparse.Namespace) -> None:
+def restore_session(ops: operations.IOperations, args: argparse.Namespace) -> None:
     """Load a session based on the window-manager provided arguments"""
     # args.settings is provided when restoring from a session.
     args.settings = None
     if args.session is None:
         return
-    session = Session(args.session)
+    session = Session(args.session, ops)
     if session.load():
         args.settings = session
         args.repo = session.repo
@@ -902,7 +906,7 @@ class Notifier(QtCore.QObject):
         self.emit_log(f'[git] {message}')
 
 
-def find_git(ops) -> str | None:
+def find_git(ops: operations.IOperations) -> str | None:
     """Return the path of git.exe, or None if we can't find it."""
     if not utils.is_win32():
         return None  # UNIX systems have git in their $PATH
--- a/cola/app.py
+++ b/cola/app.py
@@ -55,6 +55,7 @@ from . import guicmds
 from . import hidpi
 from . import i18n
 from . import icons
+from . import operations
 from . import qtcompat
 from . import qtutils
 from . import resources
@@ -106,7 +107,7 @@ def setup_environment() -> None:
     compat.setenv('GIT_PAGER', '')
 
     # Setup the openssh askpass credentials helper.
-    askpass = _get_askpass()
+    askpass = _get_askpass(operations.LocalOperations())
     compat.setenv('GIT_ASKPASS', askpass)
     compat.setenv('SSH_ASKPASS', askpass)
 
@@ -147,7 +148,7 @@ def setup_environment() -> None:
     compat.setenv('GIT_MERGE_AUTOEDIT', 'no')
 
 
-def _get_askpass() -> TextType:
+def _get_askpass(ops: operations.IOperations) -> TextType:
     """Get a default askpass program appropriate for the current environment"""
     git_askpass = core.getenv('GIT_ASKPASS')
     ssh_askpass = core.getenv('SSH_ASKPASS')
@@ -170,7 +171,7 @@ def _get_askpass() -> TextType:
         order = (kde_askpass, gnome_askpass)
 
     for askpass in order:
-        if askpass and os.path.exists(askpass):
+        if askpass and ops.exists(askpass):
             return askpass
 
     return resources.package_command('ssh-askpass')
@@ -392,7 +393,7 @@ class ColaQApplication(QtWidgets.QApplication):
         view.save_state(settings=session)
 
 
-def process_args(args: argparse.Namespace, setup_repo: bool = False) -> None:
+def process_args(args: argparse.Namespace, ops, setup_repo: bool = False) -> None:
     """Process and verify command-line arguments"""
     if args.version:
         # Accept 'git cola --version' or 'git cola version'
@@ -400,22 +401,22 @@ def process_args(args: argparse.Namespace, setup_repo: bool = False) -> None:
         sys.exit(core.EXIT_SUCCESS)
 
     # Handle session management
-    restore_session(args)
+    restore_session(ops, args)
 
     # Initialize args.repo. If a repository was specified as a
     # positional argument then we will use that value.
     # If unspecified, the current directory is used.
     if setup_repo and not args.repo:
-        if args.args and git.is_git_repository(args.args[0]):
+        if args.args and git.is_git_repository(ops, args.args[0]):
             args.repo = args.args.pop(0)
     if not args.repo:
-        args.repo = core.getcwd()
+        args.repo = ops.getcwd()
 
     # Bail out if --repo is not a directory
     repo: TextType = core.decode(args.repo)
     repo = repo.removeprefix('file:')
-    repo = core.realpath(repo)
-    if not core.isdir(repo):
+    repo = ops.realpath(repo)
+    if not ops.isdir(repo):
         errmsg = (
             N_(
                 'fatal: "%s" is not a directory.  '
@@ -427,7 +428,7 @@ def process_args(args: argparse.Namespace, setup_repo: bool = False) -> None:
         sys.exit(core.EXIT_USAGE)
 
 
-def restore_session(args: argparse.Namespace) -> None:
+def restore_session(ops, args: argparse.Namespace) -> None:
     """Load a session based on the window-manager provided arguments"""
     # args.settings is provided when restoring from a session.
     args.settings = None
@@ -449,10 +450,12 @@ def application_init(
     """Parses the command-line arguments and starts git-cola"""
     # Ensure that we're working in a valid git repository.
     # If not, try to find one.  When found, chdir there.
-    setup_environment()
-    process_args(args, setup_repo=setup_repo)
+    ops = operations.LocalOperations()
 
-    context = new_context(args, app_name=app_name)
+    setup_environment()
+    process_args(ops, args, setup_repo=setup_repo)
+
+    context = new_context(ops, args, app_name=app_name)
     enforce_single_instance(context)
 
     timer = context.timer
@@ -470,13 +473,16 @@ def application_init(
 
 
 def new_context(
-    args: argparse.Namespace, app_name: str = 'Git Cola'
+    ops: operations.IOperations,
+    args: argparse.Namespace,
+    app_name: str = 'Git Cola',
 ) -> ApplicationContext:
     """Create top-level ApplicationContext objects"""
     context = ApplicationContext(args)
     context.timestamp = time.time()
     context.settings = args.settings or Settings.read()
-    context.git = git.create()
+    context.ops = ops
+    context.git = git.create(context.ops)
     context.cfg = gitcfg.create(context)
     context.fsmonitor = fsmonitor.create(context)
     context.selection = selection.create()
@@ -490,8 +496,9 @@ def new_context(
 
 def create_context() -> ApplicationContext:
     """Create a one-off context from the current directory"""
-    args = null_args()
-    return new_context(args)  # type: ignore[arg-type]
+    ops = operations._create_ops()
+    args = null_args(ops)
+    return new_context(ops, args)  # type: ignore[arg-type]
 
 
 def enforce_single_instance(context: ApplicationContext) -> None:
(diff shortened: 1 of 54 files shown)


cola/models/main.py
```diff
@@ -673,7 +678,7 @@ def run_remote_action(
             cmd_args.append('pull')
         cmd_args.extend(transform_kwargs(**kwargs))
         cmd_args.extend(args)
-        context.notifier.git_cmd(core.list2cmdline(cmd_args))
+        context.notifier.git_cmd(context.ops.list2cmdline(cmd_args))
```
davvid: These are going to be an ongoing incremental series so I know there'll be more to come. In the meantime I'll comment on some minor tweaks so we can move this forward.

I'd suggest keeping `core.list2cmdline` as a purely local operation. Technically we're currently using subprocess.list2cmdline under the hood, and that function may not be platform-specific in behavior if I'm remember correctly.

We *mostly* use these values for display purposes, but for some scenarios (such as the DAG query field) we strongly assume and only support unix shell quoting behavior in the input field.

For that reason, I think we should keep this as local function. One thing I might change about that function these days would be to try to use `shlex.join(...)` as the implementation guts instead. The only thing that would need to be verified is whether it correctly handles non-ascii. We might still want a `core.shell_join` wrapper for that new function (so that we can throw either str or bytes at it and have it consistently return str) but we'd still want to not have it be part of the `operations` suite.

cola/models/stash.py
```diff
@@ -70,7 +68,7 @@ def do(self) -> None:
             Interaction.log_status(status, out, err)
         else:
             title = N_('Error')
-            cmdargs = core.list2cmdline(args)
+            cmdargs = self.context.ops.list2cmdline(args)
```
davvid: `list2cmdline` ~ the usage here is for display purposes only so local-only is better

cola/models/stash.py
```diff
@@ -112,7 +110,7 @@ def do(self) -> None:
...
```
davvid: same

cola/models/stash.py
```diff
@@ -138,7 +136,7 @@ def do(self) -> None:
                 Interaction.log_status(status, out, err)
             else:
                 title = N_('Error')
-                cmdargs = core.list2cmdline(args)
+                cmdargs = self.context.ops.list2cmdline(args)
```
davvid: same

cola/models/stash.py
```diff
@@ -209,8 +207,8 @@ def do(self) -> None:
         # stash ref, so now we have to remove the staged changes from the
         # worktree.  We do this by applying a reverse diff of the staged
         # changes.  The diff from stash->HEAD is a reverse diff of the stash.
-        patch = utils.tmp_filename('stash')
-        with core.xopen(patch, mode='wb') as patch_fd:
+        patch = self.context.ops.tmp_filename('stash')
+        with self.context.ops.xopen(patch, mode='wb') as patch_fd:
```
davvid: (I haven't read the whole diff yet so maybe this is answered below) ~ I'm curious how the `patch_fd` context manager is handled given that will now be writing to a remote file.

cola/widgets/commitmsg.py
```diff
@@ -79,6 +79,9 @@ def __init__(self, context, parent):
 
         self.launch_editor = actions.launch_editor_at_line(context, self)
         self.launch_difftool = actions.launch_difftool(context, self)
+        if isinstance(context.ops, operations.RemoteOperations):
```
davvid: It does tie the abstraction to the implementation a bit, but it would be nice if we could use a dedicated method, e.g. `if context.ops.is_remote():` here instead of `isinstance(...)`.

cola/widgets/diff.py
```diff
@@ -2215,24 +2218,24 @@ def new_apply_patches(context, patches=None, parent=None):
...
 
 
-def get_patches_from_paths(paths):
+def get_patches_from_paths(paths, ops: operations.IOperations):
     """Returns all patches beneath a given path"""
     paths = [core.decode(p) for p in paths]
-    patches = [p for p in paths if core.isfile(p) and p.endswith(('.patch', '.mbox'))]
-    dirs = [p for p in paths if core.isdir(p)]
+    patches = [p for p in paths if ops.isfile(p) and p.endswith(('.patch', '.mbox'))]
+    dirs = [p for p in paths if ops.isdir(p)]
```
davvid: Design note: the per-path `isfile` and `isdir` checks were cheaper before because we weren't going over the network for each one. This code seems like it would become quite inefficient if operating on a remote instance.

These kinds of patterns would benefit from batch operations so that we can process all of the `paths` entries in a single call (versus each individual `opts.isfile(...)` etc. operation).

Just a note for later.. I wouldn't expect that in the initial proof of concept, but that is the level of redesign that we should expect as this topic matures.

cola/widgets/diff.py
```diff
@@ -2354,33 +2357,33 @@ def add_files(self):
         if not files:
             return
         self.curdir = os.path.dirname(files[0])
-        self.add_paths([core.relpath(f) for f in files])
+        self.add_paths([self.context.ops.relpath(f) for f in files])
```
davvid: Same as above: these scenarios will benefit from batch operations.

cola/widgets/main.py
```diff
@@ -705,9 +708,21 @@ def __init__(self, context, parent=None):
         self.patches_menu.addAction(self.apply_patches_skip_action)
         self.patches_menu.addAction(self.apply_patches_abort_action)
 
+        if isinstance(context.ops, operations.RemoteOperations):
```
davvid: This is another `isinstance(...)` checking for whether this is a remote instance so having `context.ops.is_remote_io()` would help here as well.

cola/widgets/main.py
```diff
@@ -626,7 +626,9 @@ def __init__(self, context, parent=None):
         # We can do this because can_rebase == not is_rebasing
         self.rebase_start_action_proxy = utils.Proxy(
             self.rebase_start_action,
-            setEnabled=lambda x: self.rebase_start_action.setEnabled(not x),
+            setEnabled=lambda x: self.rebase_start_action.setEnabled(
+                False if isinstance(context.ops, operations.RemoteOperations) else not x
```
davvid: With the dedicated method this conditional would look like: `not context.ops.is_remote_io() and not x`

cola/widgets/finder.py
```diff
@@ -266,7 +265,7 @@ def search_for(self, txt):
         self.focus_input()
 
     def process_result(self, filenames):
-        self.tree.set_filenames(filenames, select=True)
+        self.tree.set_filenames(filenames, self.context.ops, select=True)
```
davvid: This part made me realize that we're adjusting a fair number of method signatures in order to pass in the `self.context.ops`.

For here (and maybe a few others), it seems like a better long-term play would be to adjust the constructor so that it accepts `context`, just like all of the other widgets, dialogs and misc. code that accepts `context`.

If we do that for the FileTree constructor, for example, then we wouldn't need to change the `set_filenames()` signature. The method could instead operate on the `self.context.ops` that were specified during construction via the `context` object.

When in doubt, making things take a `context` parameter is closer in spirit to the current design.

cola/widgets/remote.py
```diff
@@ -489,7 +487,7 @@ def update_command_display(self):
             args, kwargs = main.remote_args(self.context, remote, self.action, **kwargs)
             cmd.extend(git.transform_kwargs(**kwargs))
             cmd.extend(args)
-            commands.append(core.list2cmdline(cmd))
+            commands.append(self.context.ops.list2cmdline(cmd))
```
davvid: note: `list2cmdline` is used purely for display here

cola/widgets/remote.py
```diff
@@ -948,7 +946,7 @@ def _prune_selected_remotes(self):
         """Prune stale worktrees from the persistent selected_remotes_by_worktree"""
         worktrees = list(self.selected_remotes_by_worktree.keys())
         for worktree in worktrees:
-            if not os.path.exists(worktree):
+            if not self.context.ops.exists(worktree):
```
davvid: todo for later: `for worktree in worktrees` ls a loop with guts that could be batched to minimize network request round trips.

cola/widgets/standard.py
```diff
@@ -918,9 +919,10 @@ def set_value(self, value):
 class DirectoryPathLineEdit(QtWidgets.QWidget):
     """A combined line edit and file browser button"""
 
-    def __init__(self, path, parent):
+    def __init__(self, path, parent, ops: operations.IOperations):
```
davvid: This is an example where passing-in and storing the `context` (instead of `ops`) is more consistent with the existing code.

cola/widgets/editremotes.py
```diff
@@ -513,6 +513,6 @@ def validate(self, _text_or_index=''):
     def open_repo(self):
         """Set the URL from a repository on disk"""
         git = self.context.git
-        repo = qtutils.opendir_dialog(N_('Open Git Repository'), core.getcwd())
+        repo = qtutils.opendir_dialog(N_('Open Git Repository'), core.getcwd())  # TODO make this work on remote
```
davvid, line 516: This will be a bit of a challenge. We're using Qt's builtin file dialogs. We'd have to do something very custom (implement our own remote filesystem model, more or less) in order to do it.

`./bin/git-cola browse` (implementation: `cola/widgets/browse.py` and the `GitTreeModel` which ultimately derives from `QtGui.QStandardItemModel`) is an example file browser-ish widget that demonstrates how someone might go about doing that.

cola/widgets/filetree.py
davvid: This is one of the constructors that should be updated to take a `context` object.

cola/widgets/status.py
```diff
@@ -936,11 +939,13 @@ def _create_unstaged_context_menu(self, menu, s):
...
             menu.addAction(self.delete_untracked_files_action)
             menu.addSeparator()
-            menu.addAction(
+            action = menu.addAction(
                 icons.edit(),
                 N_('Ignore...'),
                 partial(gitignore.gitignore_view, self.context),
             )
+            if isinstance(self.context.ops, operations.RemoteOperations):
+                action.setEnabled(False)
```
davvid, line 967: `if self.context.ops.is_remote_io():`

cola/app.py
```diff
@@ -142,10 +144,10 @@ def setup_environment() -> None:
     # --- >8 --- >8 ---
     # Longer-term: Use `git merge --no-commit` so that we always
     # have a chance to explain our merges.
-    compat.setenv('GIT_MERGE_AUTOEDIT', 'no')
+    compat.setenv('GIT_MERGE_AUTOEDIT', 'no', operations.LocalOperations())
```
davvid: It seems like we should create a single `LocalOperaitons` instance and reuse it throughout this function instead of creating a new one for each call.

cola/cmds.py
```diff
@@ -895,10 +898,10 @@ def do(self) -> tuple[int, str, str]:
                     **kwargs,
                 )
             )
-            self.context.notifier.git_cmd(core.list2cmdline(cmd_args))
+            self.context.notifier.git_cmd(self.context.ops.list2cmdline(cmd_args))
```
davvid: note: purely cosmetic `list2cmdline` usage

cola/cmds.py
```diff
@@ -960,16 +963,16 @@ def do(self) -> None:
...
-        core.write(filename, new_additions)
+        self.context.ops.write_file(filename, new_additions)
         Interaction.log_status(0, f'Added to {filename}:\n{for_status}', '')
         self.model.update_file_status()
 
 
-def file_summary(files: list[TextType]) -> str:
-    txt = core.list2cmdline(files)
+def file_summary(files: list[TextType], ops: operations.IOperations) -> str:
+    txt = ops.list2cmdline(files)
```
davvid: note: also purely cosmetic `list2cmdline`

cola/cmds.py
```diff
@@ -1254,7 +1259,7 @@ def do(self) -> None:
 
         title = N_('Delete Files?')
         msg = N_('The following files will be deleted:') + '\n\n'
-        msg += file_summary(files)
+        msg += file_summary(files, self.context.ops)
```
davvid: I would expect these diffs to disappear once we stop using remote ops for list2cmdline. Please make sure to squash any updates into your original commit.

cola/cmds.py
```diff
@@ -2112,8 +2117,8 @@ def do(self) -> None:
         if not dirnames:
             return
         # An empty dirname defaults to to the current directory.
-        dirs = [(dirname or core.getcwd()) for dirname in dirnames]
-        utils.launch_default_app(dirs)
+        dirs = [(dirname or self.context.ops.getcwd()) for dirname in dirnames]
+        utils.launch_default_app(dirs, self.context.ops)
```
davvid: A more consistent signature for `launch_default_app` would be `utils.launch_default_app(self.context, dirs)`.

There's a number of places where we try to make the `context` the 1st arg. Ditto in the diff hunk above this hunk as well.

cola/cmds.py
```diff
@@ -2312,13 +2317,14 @@ class SequenceEditorEnvironment:
     def __init__(self, context: ApplicationContext, **kwargs) -> None:
         self.env = {
             'GIT_EDITOR': prefs.editor(context),
-            'GIT_SEQUENCE_EDITOR': sequence_editor(),
+            'GIT_SEQUENCE_EDITOR': sequence_editor(context.ops),
```
davvid: This is going to be especially tricky, and perhaps something we won't be able to support in Remote Control mode.

The way `git rebase` handles a custom "sequence" editor is that we set the `GIT_SEQUENCE_EDITOR` environment variable and then we run `git rebase --interactive`.

`git` itself is then in charge of executing the `GIT_SEQUENCE_EDITOR` directly -- we don't have control over that. The tricky part is that it's going to be executing that over there on the remote machine.

The difficulty is that the sequence editor `./bin/git-cola-sequence-editor` is the command that's getting executed, and that command is a fully interactive Qt GUI application. The problem is that it's running over there, on the remote side, via the remote `git rebase`.

I would suggest punting on that feature.

cola/cmds.py
```diff
@@ -3469,7 +3477,7 @@ def get_args(self) -> list[str]:
         return cmd
 
 
-def launch_history_browser(argv: list[str]) -> None:
+def launch_history_browser(argv: list[str], ops: operations.IOperations) -> None:
```
davvid: I'd recommend `def launch_history_browser(context, argv):` with the `context` as the 1st argument.

cola/git.py
```diff
@@ -327,20 +333,20 @@ def execute(
...
                     % (elapsed_time, ' '.join(command), status, out, err)
                 )
             else:
-                core.print_stderr(
+                ops.print_stderr(
                     '# %.3fs: %s -> %d' % (elapsed_time, ' '.join(command), status)
                 )
         elif cola_trace:
-            core.print_stderr('# {:.3f}s: {}'.format(elapsed_time, ' '.join(command)))
+            ops.print_stderr('# {:.3f}s: {}'.format(elapsed_time, ' '.join(command)))
```
davvid: Hmm... it seems like these debug `print_stderr(...)` prints should be a local-only operation.

cola/gitcfg.py
```diff
@@ -405,10 +409,10 @@ def get_author(self) -> tuple[str, str]:
         if _use_pwd:
             user = pwd.getpwuid(os.getuid()).pw_name
         else:
-            user = os.getenv('USER', 'unknown')
+            user = self.context.ops.getenv('USER', 'unknown')
 
         name: str = self.get('user.name', user)
-        email: str = self.get('user.email', f'{user}@{core.node()}')
+        email: str = self.get('user.email', f'{user}@{self.context.ops.node()}')
```
davvid: This is fallback code that we typically don't hit, but I wonder if we should use the local client node name or hte remote node name. I'd personally lean towards local, especially if it avoids a request round-trip.

cola/gitcmds.py
```diff
@@ -213,7 +213,7 @@ def current_branch(context: ApplicationContext) -> core.UStr:
     else:
         stat_file = context.git.git_path('HEAD')
     try:
-        key = core.stat(stat_file).st_mtime
+        key = context.ops.stat(stat_file).st_mtime
```
davvid: I haven't looked too closely, but if we can send `struct stat` and its `st_mtime` over the wire that'd be great. I was under the impression that this struct is platform-specific, but we can probably get away with just marshaling the common subset of bits we need.

cola/hidpi.py
```diff
@@ -29,11 +30,13 @@ def apply_choice(value: str) -> None:
...
-            compat.unsetenv('QT_SCALE_FACTOR')
+            compat.setenv(
+                'QT_AUTO_SCREEN_SCALE_FACTOR', '1', ops=operations.LocalOperations()
+            )
+            compat.unsetenv('QT_SCALE_FACTOR', ops=operations.LocalOperations())
     elif value and value != Option.DISABLE:
-        compat.unsetenv('QT_AUTO_SCREEN_SCALE_FACTOR')
-        compat.setenv('QT_SCALE_FACTOR', value)
+        compat.unsetenv('QT_AUTO_SCREEN_SCALE_FACTOR', ops=operations.LocalOperations())
+        compat.setenv('QT_SCALE_FACTOR', value, ops=operations.LocalOperations())
```
davvid: Use a single `LocalOperations` instance here.

cola/main.py
```diff
@@ -13,11 +13,12 @@
 from .widgets.main import MainView
 
 if TYPE_CHECKING:
+    from . import server
     from .app import ApplicationContext
 
 
 def main(argv: list[str] | None = None) -> int:
-    app.initialize()
+    app.initialize(None)
```
davvid: If we have to pass `None` here then we should make this an optional argument that defaults to `None`.

cola/main.py
```diff
@@ -788,6 +799,16 @@ def cmd_tag(args: argparse.Namespace) -> int:
     return app.application_start(context, view)
 
 
+def cmd_connect(args: argparse.Namespace):
+    from . import server
```
davvid: Try to avoid inline imports. Prefer these near the top-level with all the rest of them. If this is causing a circular import error, though, then that's a design issue that'll need to be addressed.

cola/operations.py
```diff
@@ -0,0 +1,793 @@
...
+        self.seq_number = 0
+
+    def _send_op(self, data):
+        received = self.client.send_message_msgpack(data)
+        if received.get('seq_number') != data.get('seq_number'):
+            raise Exception('seq_number was not correct')
+        self.seq_number += 1
+
+        if received.get('is_exception', False):
+            raise Exception(f"Error was found: {received.get('result')}")
```
davvid: Should we be using `IOError` instead of `Exception` for these?

I'd suggest dropping the "Error was found:" part -- it'll likely be noise around the actual details from `received.get('result')`.

cola/operations.py
```diff
@@ -0,0 +1,793 @@
...
+            'op': 'expanduser',
+            'args': [],
+            'kwargs': {'path': path},
+        }
+        return self._send_op(data)
+
+    def run_command(self, cmd, *args, **kwargs):
+        safe_kwargs = {}
+
+        # Only allow safe fields
```
davvid: Please elaborate this comment with respect to what constitutes a "safe" vs. a non-"safe" field. The `isinstance` checks, for example, are slightly surprising, so it'd be good to understand what's up there.

cola/sequenceeditor.py
```diff
@@ -377,7 +377,7 @@ def save(self, string=None):
             string = '\n'.join(lines) + '\n'
 
         try:
-            core.write(self.filename, string)
+            self.context.ops.write_file(self.filename, string)
```
davvid: This is the `GIT_SEQUENCE_EDITOR` part that I mentioned above.

cola/server.py
```diff
@@ -0,0 +1,152 @@
...
+        ):
+            print(f'Server running on ws://0.0.0.0:{self.port}')
+            await asyncio.Future()  # run forever
+
+    def run(self):
+        asyncio.run(self.run_async())
+
+
+class SocketClient:
+    def __init__(self, ip: str, port: int = 8000, protocol: str = 'ws'):
```
davvid: Port 8000 is very common, and may clash by default with other things the user may be running.

If we wanted to be clever we could use port `0xc01a` by default (`49178`).

test/git_test.py
```diff
@@ -74,9 +76,9 @@ def test_find_git_dir_found_right_away(is_git_dir):
 def test_find_git_does_discovery(is_git_dir):
     git_dir = str(pathlib.Path('/the/root/.git').resolve())
     worktree = str(pathlib.Path('/the/root').resolve())
-    is_git_dir.side_effect = lambda x: x == git_dir
+    is_git_dir.side_effect = lambda path, ops: path == git_dir
 
-    paths = git.find_git_directory('/the/root/sub/dir')
+    paths = git.find_git_directory('/the/root/sub/dir', operations.LocalOperations())
```
davvid: Here and elsewhere -- the `ops` argument should probably be the very first argument, that way there's a regular shape to all of the functions.

test/resources_test.py
```diff
@@ -13,11 +14,11 @@ def test_command_unix(mock_prefix, mock_compat):
     mock_prefix.return_value = helper.fixture()
 
     expect = helper.fixture('bin', 'bare-cmd')
-    actual = resources.command('bare-cmd')
+    actual = resources.command('bare-cmd', operations.LocalOperations())
     assert expect == actual
 
     expect = helper.fixture('bin', 'exe-cmd')
-    actual = resources.command('exe-cmd')
+    actual = resources.command('exe-cmd', operations.LocalOperations())
```
davvid: Reuse a single LocalOperations instance for the duration of each test.

test/server_test.py
```diff
@@ -0,0 +1,312 @@
...
+class create_test_server:
+    def __init__(self):
+        app = server.SocketServer()
+
+        self.server_thread = multiprocessing.Process(target=app.run, daemon=True)
+        self.server_thread.start()
+        time.sleep(0.4)
+
+    def __enter__(self):
+        self.socket = server.SocketClient(ip='127.0.0.1', port=8000)
```
davvid: Since this is a test with a hard-coded port number then we should really be careful to use an uncommon port. That said, if we run into actual issues with a hard-coded port number then we consider letting it be overridden via an environment variable.

test/server_test.py
```diff
@@ -0,0 +1,312 @@
...
+            './testfile_remote'
+        )
+
+
+def test_server_remove():
+    with create_test_server() as s:
+        for _ in range(2):
+            with open('test.file', 'w') as f:
+                f.close()
+            if _ == 0:
```
davvid: If we're going to use the variable then it should have a name (not `_`).

test/server_test.py
```diff
@@ -0,0 +1,312 @@
...
+        assert results_local == results_remote
+
+    os.remove('test.file')
+
+
+def test_server_run_command():
+    with create_test_server() as s:
+        assert s.ops_local.run_command(
+            ['echo', 'hello world']
+        ) == s.ops_remote.run_command(['echo', 'hello world'])
```
davvid: I'd recommend adjusting the `hello world` string with one that contains non-ascii unicode. That way we'll be sure that we're correctly handling unicode.

test/server_test.py
```diff
@@ -0,0 +1,312 @@
...
+
+def test_server_get_environ():
+    with create_test_server() as s:
+        s.ops_local.environ_setvalue('key_test', 'test_value')
+        s.ops_remote.environ_setvalue('key_test', 'test_value')
+
+        assert (
+            s.ops_local.get_environ()['key_test']
+            == s.ops_remote.get_environ()['key_test']
+        )
```
davvid, line 268: Same for the rest of these tests -- it'd be good to validate this using non-ascii unicode values.

davvid (changes requested):
Thanks for seeing this through. These are relatively minor notes (with only a few that require actual changes) so please take a look, squash any updates into the existing commit, and then we can probably start merging this first pass.

I'm aware that there will be more to come later, but this is on the right track for establishing the core functionality.

I didn't see any code related to handling dropped/reset connections, re-connection, or other error scenarios, so I'm assuming that those last mile concerns will be addressed later once we're past the initial proof of concept.

Thanks

cola/operations.py
```diff
@@ -0,0 +1,793 @@
...
+
+
+class RemoteOperations(IOperations):
+    def __init__(self, socket_client: server.SocketClient):
+        from . import server
+
+        self.client = server.SyncSocketClient(socket_client)
+        self.seq_number = 0
+
+    def _send_op(self, data):
```
p12tic: Add types to _send_op. It's weird that mypy isn't complaining about this.

cola/operations.py
```diff
@@ -0,0 +1,793 @@
...
+        received = self.client.send_message_msgpack(data)
```
p12tic, line 355: I think it would be better to make callers of _send_op not care about `seq_number` at all. Move addition of `seq_number` to data to this function.

p12tic (approved):
LGTM except for nits I noticed and existing comments.

cola/cmds.py
Ahrirg: I’ve reverted these changes for now.

cola/models/main.py
Ahrirg: I’ve reverted these changes for now.

cola/operations.py
Ahrirg: From my understanding, `IOError` was merged into `OSError` in Python 3. `garden check` is automatically replacing `IOError` with `OSError` as a result. Should i use `OSError` or something else?

davvid: OSError should be fine. Either that, or we should derive our own subclass(es) from Exception and raise that instead.

cola/widgets/diff.py
Ahrirg: Added a comment for now. I'll implement this in a later PR.

cola/widgets/editremotes.py
Ahrirg: We already require the repository path to be provided via the CLI. The current alternative is running `git-cola` over something like `Waypipe`, which also requires specifying the repository path, so this wouldn't really be a worse experience. I will add remote file dialog later if needed.

cola/models/stash.py
Ahrirg: xopen currently supports only read operations and doesn't support write operations yet. I'll implement write support in a later PR. Reverted the change, my bad.

p12tic: xopen with wb is used only to capture command output in fd. The plan is to enhance command runner to be able to collect results to a file all by itself. This way xopen problem will disappear.

One way would be to have a dataclass `CollectOutputToFile` with path member and pass it to `_stdout` param. Command runner would understand what this means and do what's needed on both local and remote sides.

davvid:
This is looking good to me for an initial merge. I'm aware that this is a very large change and is very prone to merge conflicts as we continue to evolve the code.

To make things easier, I'm going to merge a batch of changes tonight and then I'll keep the tree relatively stable for about a week.

@Ahrirg if you could please rebase after tonight's merges I'll prioritize merging this change since I don't want to keep moving the goal posts.

Ahrirg:
just a friendly ping, this is ready for merge if nothing else needs to be changed.

PR #1618, opened 2026-07-14
git: skip optional index locks on read-only commands

joshuataylor:
git-cola runs a lot of read-only git commands in the background -- `status`, `diff`, `log`.

Those already avoid git-cola's own in-process index lock (they pass `_readonly=True`), but git itself still takes the optional `.git/index.lock` to refresh the index stat cache while it reads.

That means the background reads do disk I/O and can contend with each other, or with any external git process touching the same repository -- e.g. if you also have an IDE open like JetBrains PyCharm doing its own git work in the same repo.

git has a top-level flag for this: [`--no-optional-locks`](https://git-scm.com/docs/git#Documentation/git.txt---no-optional-locks).

Passing it for the read-only commands lets git skip the stat-cache refresh and the lock that goes with it, reducing disk I/O and lock contention, stale lock files and other issues.

Because it's a top-level option (not a subcommand option), it's inserted BEFORE the subcommand, and only when `_readonly` is set -- commands that may write are unaffected:

`git -c ... --no-optional-locks <subcommand> ...`

`--no-optional-locks` has been available since git 2.15.2 (September 2017), well within the git versions git-cola already targets, so there's no new version requirement.

Here's a video example, where I added logging to show where it's being used, it's easier to explain:

https://github.com/user-attachments/assets/dc4350be-595c-4c3f-b6b0-b79792bbbca2

> I also re-read CONTRIBUTING.md, and have tried to adhere to the commit title conventions etc, let me know if you need me to do anything that'll be helpful when reviewing, merging, etc. I have a few other random changes I'd like to propose, so I want to make sure I'm actually not causing any dramas when reviewing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,15 @@
+.. _v4.20.0:
+
+v4.20.0
+=======
+
+Usability, bells and whistles
+-----------------------------
+* Git Cola now avoids taking optional locks that can interact with
+  external ``git`` processes.
+  (`#1618 <https://github.com/git-cola/git-cola/pull/1618>`_)
+
+
 .. _v4.19.0:
 
 v4.19.0
--- a/cola/app.py
+++ b/cola/app.py
@@ -112,6 +112,10 @@ def setup_environment() -> None:
     compat.setenv('GIT_ASKPASS', askpass)
     compat.setenv('SSH_ASKPASS', askpass)
 
+    # Avoid taking optional locks for read-only operations.
+    # We assume that proper state-modifying operations continue to take locks.
+    compat.setenv('GIT_OPTIONAL_LOCKS', '0')
+
     # --- >8 --- >8 ---
     # Git v1.7.10 Release Notes
     # =========================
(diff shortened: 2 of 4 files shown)


davvid (changes requested):
This is a good idea.

Since the `git` option only affects *optional* locks, then it seems like it's harmless to have this mode enabled at all times. If that's the case, I noticed this note in `git help git`:

```
     --no-optional-locks
         Do not perform optional operations that require locks.
         This is equivalent to setting the GIT_OPTIONAL_LOCKS to 0.
```

So, instead of new machinery to conditionally include the option, I would lean towards setting `GIT_OPTIONAL_LOCKS` to `0` in `cola/app.py`'s `setup_environment()` initialization step instead.

All of the state-mutating commands ( `git commit`, `git update-index`, etc) should continue to operate as before with the environment variable set, so it *seems* simpler to always enable that mode.

If that doesn't end up working out in practice then the changes in this PR seem like they could be merged, but please try the environment variable first.

If that works then we can eschew the tests too since there'll be not much to test.

joshuataylor:
Sorry! Forgot to respond!

I was first hesitant about the global usage of no-optional-locks, but I've been using it for the last couple of weeks locally with 0 dramas.

davvid:
Thanks for the confirmation @joshuataylor. I've only been using it today so hearing that you've been using it for a couple of weeks is great. Much appreciated.

PR #1619, opened 2026-07-14
stage: add undo support for Stage and Unstage commands

### LINKED ISSUE #1550: Feature request: Undo-able stage and unstage operations
* Thank you for the great tool. I like Git Cola’s **line-based (un)stage UI/UX.** I have an idea to make it even more appealing, so I would like to request a feature. This issue is related to #1549. Discussing them together may be effective, but I opened separate issues to design each final UI/UX well. I think these issues are independent, and each can be discussed separately.
* This issue's feature is the **most important feature for me**, because I want to **split patches into small, precise commits.**

## Motivation

* Git Cola supports line-based staging to create atomic commits.
* When carefully staging changes in small hunks, it is easy to accidentally stage or unstage a whole file or a larger range.
* In such cases, the previously crafted staging state is lost and cannot be recovered.
* This often discourages users from using fine-grained staging, or forces excessive caution.

---

## Problem

* Stage and unstage operations can irreversibly overwrite the current index (line-based) state.
* There is no confirmation dialog (for good reason), and no undo mechanism.
* As a result, a single miss-type or miss-click **can destroy a carefully prepared staging state**.

---

## UX Goal

* Avoid intrusive confirmation dialogs for frequent actions.
* Make stage and unstage operations safely reversible.
* Allow users to work aggressively and confidently with fine-grained staging.

---

## Proposal

* Make stage and unstage operations undo-able.
* Provide an immediate Undo action after each stage/unstage operation.
* Treat each operation as a single undo-able unit.

---

## Behavior / UX sketch

* After a stage or unstage operation (line-based (un)staging or file-based), show a brief notification with an **Undo** action.
* Undo applies to the last operation only (stack-based).
* Optionally, provide a small history view showing recent operations (file count, line count).

Image
brief notification with an **Undo** action; small history view showing recent operations

Image
undo history window

## Relation to undoable revert

* This proposal follows the same design philosophy as **Undo-able Revert**.
* The difference is the affected area:

  * Revert Undo: working tree and index
  * Stage/Unstage Undo: index only
* Both aim to prevent irreversible loss caused by a single action.

---

## Technical notes (example approach)

* Capture index changes before each stage/unstage operation.
* Store undo data separately from `git stash`.
* Restore index state using cached patches.
* Keep undo history short and local.

see detail Git CLI commands:
https://github.com/git-cola/git-cola/issues/1549#issuecomment-3848077759

---

## Non-goals

* No confirmation dialogs for stage/unstage.
* No partial undo within a single operation.
* Not a general history or time-travel feature.

Thank you for considering this idea.
Do you think this feature would be useful?
I would be happy to hear your thoughts.

AFRADOX:
## Summary

Fixes #1550

Adds undo support to the `Stage` and `Unstage` commands. Previously,
staging or unstaging a file (whole-file or line-based/hunk staging)
had no way to be undone — a single accidental click could discard
carefully prepared partial staging with no recovery path.

## Approach

Before staging/unstaging, the command snapshots whatever was already
staged for the affected paths using `git diff --cached`. On undo, the
paths are reset and the snapshot is re-applied via `git apply --cached`,
restoring the exact prior state — whether that was nothing staged,
fully staged, or partially staged via hunks.

This matches the approach discussed in the issue: storing a patch
(rather than a full index snapshot) keeps undo fast and avoids writing
extra objects into the repository.

## Testing

Added two new tests in `test/cmds_test.py` that exercise undo against
a real temporary git repository (not mocks):

- `test_stage_undo_restores_untracked_state` — stages a new file, then
  undoes it, confirming the file returns to untracked.
- `test_unstage_undo_restores_staged_state` — unstages a modified file,
  then undoes it, confirming it returns to staged.

Full test suite passes (274 passed; 2 pre-existing unrelated failures
on Windows due to CRLF line-ending differences in commit message tests).

## Notes

This covers whole-file and line-based staging via `Stage`/`Unstage`.
Wiring this into the UI (e.g. an "Undo" button/shortcut after staging)
is a natural follow-up but is out of scope for this PR, which focuses
on the underlying undo capability.

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -2900,6 +2900,8 @@ def should_stage_conflicts(path) -> bool:
 class Stage(ContextCommand):
     """Stage a set of paths."""
 
+    UNDOABLE = True
+
     @staticmethod
     def name() -> str:
         return N_('Stage')
@@ -2907,11 +2909,36 @@ class Stage(ContextCommand):
     def __init__(self, context: ApplicationContext, paths: list[str]) -> None:
         super().__init__(context)
         self.paths = paths
+        self._old_diff = ''
 
     def do(self) -> tuple[int, str, str]:
+        if not super().do():
+            return (0, '', '')
+        if self.paths:
+            status, out, _ = self.git.diff(
+                '--', *self.paths, cached=True, _readonly=True
+            )
+            if status == 0:
+                self._old_diff = out
         msg = N_('Staging: %s') % (', '.join(self.paths))
         Interaction.log(msg)
         return self.stage_paths()
+
+    def undo(self) -> None:
+        super().undo()
+        if not self.paths:
+            return
+        self.git.reset('--', *self.paths)
+        if self._old_diff:
+            tmp_file = utils.tmp_filename('undo-stage')
+            core.write(tmp_file, self._old_diff)
+            try:
+                status, out, err = gitcmds.apply_diff(self.context, tmp_file)
+            finally:
+                core.unlink(tmp_file)
+            if status != 0:
+                Interaction.log_status(status, out, err)
+        self.model.update_files(emit=True)
 
     def stage_paths(self) -> tuple[int, str, str]:
         """Stages add/removals to git."""
@@ -3174,6 +3201,8 @@ class Tag(ContextCommand):
 class Unstage(ContextCommand):
     """Unstage a set of paths."""
 
+    UNDOABLE = True
+
     @staticmethod
     def name() -> str:
         return N_('Unstage')
@@ -3181,12 +3210,20 @@ class Unstage(ContextCommand):
     def __init__(self, context: ApplicationContext, paths: list[str]) -> None:
         super().__init__(context)
         self.paths = paths
+        self._old_diff = ''
 
     def do(self) -> tuple[int, str, str]:
         """Unstage paths"""
+        if not super().do():
+            return (0, '', '')
         context = self.context
         head = self.model.head
         paths = self.paths
+
+        if paths:
+            status, out, _ = self.git.diff('--', *paths, cached=True, _readonly=True)
+            if status == 0:
+                self._old_diff = out
 
         msg = N_('Unstaging: %s') % (', '.join(paths))
         Interaction.log(msg)
@@ -3196,6 +3233,20 @@ class Unstage(ContextCommand):
         Interaction.command(N_('Error'), 'git reset', status, out, err)
         self.model.update_file_status()
         return (status, out, err)
+
+    def undo(self) -> None:
+        super().undo()
+        if not self.paths or not self._old_diff:
+            return
+        tmp_file = utils.tmp_filename('undo-unstage')
+        core.write(tmp_file, self._old_diff)
+        try:
+            status, out, err = gitcmds.apply_diff(self.context, tmp_file)
+        finally:
+            core.unlink(tmp_file)
+        if status != 0:
+            Interaction.log_status(status, out, err)
+        self.model.update_file_status()
 
 
 class UnstageAll(ContextCommand):
--- a/test/cmds_test.py
+++ b/test/cmds_test.py
@@ -220,3 +220,53 @@ def test_undo_last_commit_confirms_action(prefs, interaction):
     assert cmd.confirm()
     context.model.is_commit_published.assert_called_once()
     interaction.confirm.assert_called_once()
+
+
+from cola import core
+from cola import gitcmds
+
+from . import helper
+from .helper import app_context
+
+# Prevent unused imports lint errors.
+assert app_context is not None
+
+
+def test_stage_undo_restores_untracked_state(app_context):
+    """Undoing a Stage on a new file returns it to untracked"""
+    app_context.timestamp = time.time()
+    model = app_context.model
+    helper.touch('new_file.txt')
+    model.update_status()
+    assert 'new_file.txt' in model.untracked
+
+    cmd = cmds.Stage(app_context, ['new_file.txt'])
+    cmd.do()
+    model.update_status()
+    assert 'new_file.txt' in model.staged
+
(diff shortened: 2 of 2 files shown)


cola/cmds.py
```diff
@@ -2900,19 +2900,42 @@ def should_stage_conflicts(path) -> bool:
...
+        self.git.reset('--', *self.paths)
+        if self._old_diff:
+            tmp_file = utils.tmp_filename('undo-stage')
+            try:
+                core.write(tmp_file, self._old_diff)
+                gitcmds.apply_diff(self.context, tmp_file)
+            finally:
+                core.unlink(tmp_file)
+        self.model.update_files(emit=True)
+
```
davvid, line 2942: Please narrow the `try/finally`. I don't think `apply_diff` should be inside of that exception block.

Also, please make sure to check the `(status, out, err)` return values from apply_diff, specifically the status code. It should be 0 on success.

cola/cmds.py
```diff
@@ -2900,19 +2900,42 @@ def should_stage_conflicts(path) -> bool:
...
         self.paths = paths
+        self._old_diff = ''
 
     def do(self) -> tuple[int, str, str]:
+        if not super().do():
+            return (0, '', '')
+        if self.paths:
+            self._old_diff = self.git.diff(
+                '--cached', '--', *self.paths, _readonly=True
+            )[STDOUT]
```
davvid: It'd be better to capture the `status, out, _ = self.git.diff(...)` separately and only set old_diff when `status == 0`.

davvid (changes requested):
Please address these notes and squash all updates (including the pre-commit formatting/etc. fixes) into a single commit.

This does look more or less correct. We will be accumulating more memory when running a long session, but that's a fair tradeoff and something that can be addressed elsewhere. One thing we might want to do later is have a way to flush the undo/redo queue, but that's unrelated to these changes.

To be clear -- this does implement the low-level undo operations, but it does not expose the undo/redo functionality in the main UI.

You may have noticed that the code is already structured to support undo/redo via Command objects.

Now that we have a few that actually support undo then I'll have to follow up and actually start exposing the undo functionality in the UI.

Thanks

AFRADOX:
Hi davvid,
Thanks for the review! I've addressed all three points:

Captured status, out, _ from git diff separately and only set self._old_diff when status == 0, for both Stage.do() and Unstage.do().
Narrowed the try/finally blocks in both undo() methods so only core.unlink(tmp_file) is in finally — apply_diff is called outside the write, with its own explicit status check.
Checked the (status, out, err) return from apply_diff and log the failure via Interaction.log_status() if it's non-zero.

Also squashed everything (including the pre-commit.ci formatting commits) into a single commit as requested.

cola/cmds.py
```diff
@@ -2930,19 +2900,46 @@ def should_stage_conflicts(path) -> bool:
...
         self.paths = paths
+        self._old_diff = ''
 
     def do(self) -> tuple[int, str, str]:
+        if not super().do():
+            return (0, '', '')
+        if self.paths:
+            status, out, _ = self.git.diff(
+                '--cached', '--', *self.paths, _readonly=True
+            )
```
davvid, line 2920: Please change this to use this syntax for `--cached`:

```python
status, out, _ = self.git.diff(
    '--', *self.paths, cached=True, _readonly=True
)
```

We translate `keyword_args=True` into dashed `--keyword-args` for git subcommands inside the `cola.git.Git` command wrapper. Prefer using that feature for options such as `--cached` by passing `cached=True` instead of the `'--cached'` string.

Please make that same change in `class Unstage` below as well.

cola/cmds.py
```diff
@@ -2992,6 +2989,57 @@ def stage_all(self) -> tuple[int, str, str]:
...
+        core.write(tmp_file, self._old_diff)
+        try:
+            status, out, err = gitcmds.apply_diff(self.context, tmp_file)
+        finally:
+            core.unlink(tmp_file)
+        if status != 0:
+            Interaction.log_status(status, out, err)
+        self.model.update_file_status()
+
+
```
davvid: One last request before we merge -- please move `class Unstage` down so that it's immediately before `class UnstageAll`, just like it originally was.

These diffs move that class higher up in the file, which disconnects its diffs completely from the original code. You can see that in the diff below -- that entire section is red (and this entire section is green) when the diffs should be interspersed in the original section below.

That'll make it so that we can actually see what changed. Right now it's presented as a brand new full on additional new class and a full on removal of a class later down in the file due to the movement of the class. Keep it in-place.

davvid (changes requested):
Thanks for the updates. This is very close to merging now so please take a look.

test/cmds_test.py
```diff
@@ -220,3 +220,53 @@ def test_undo_last_commit_confirms_action(prefs, interaction):
...
+
+
+from cola import core
+from cola import gitcmds
+
+from . import helper
+from .helper import app_context
+
+# Prevent unused imports lint errors.
+assert app_context is not None
```
davvid, line 232: Heads-up that imports should be at the top-level (we try to avoid inline imports whenever possible, even in tests).

I'm fixing this up locally post-merge.

AFRADOX:
Hi davvid,
Updated based on your latest feedback:

Changed '--cached' to cached=True in both Stage.do() and Unstage.do()
Moved class Unstage back to its original position (right before class UnstageAll)

test/cmds_test.py
davvid: There was also an unused import. I typically have [garden](https://gitlab.com/garden-rs/garden) installed (`cargo install garden-tools`) and I use these two invocations to run the tests and checks locally (these would have caught these issues):

```
# Setup a virtualenv for testing
garden dev

# Run checks and tests
garden check -vv
```

davvid (approved):
I've fixed these up locally post-merge. Thanks

PR #1620, opened 2026-07-15
Safer Revert: Clearly Confirm Changes Before Proceeding

### LINKED ISSUE #1548: Feature request: Safer revert: clearly confirm changes before proceeding
### Motivation

Revert is a destructive operation.
Even when users understand the risk, confidence depends on being able to clearly see **what will be reverted** before proceeding.

### Problem

When reverting multiple files, Git Cola shows a warning dialog that says the operation cannot be undone.
This warning increases caution, but the dialog does not clearly show **the complete set of files** or **the full scope of changes** that will be reverted.

As a result:

* Users feel uneasy about whether only the intended files are affected
* Users try to re-check selections in the file list, which may be hidden or partially visible
* Especially with multiple selected files, users hesitate to press the Revert button

### Current behavior

Select the two files `.gitignore` and `README.md`, and select Revert.

Git Cola shows the Revert dialog and prompts a warning.

Because the operation cannot be undone, carefully check whether the selected files include anything not wanted.
However, especially when multiple files are selected, not all files or changes can be checked because they are hidden in the GUI (also because the dialog is modal).
Hesitate to press the Revert button.

|Image|Image|
|----|----|
|Current | Proposed |

### Proposed behavior

Select the two files `.gitignore` and `README.md`, and select Revert.

Git Cola shows the Revert dialog and prompts a warning.

Git Cola can show:

* The complete set of files that will be reverted
* The size of the changes (diff stat)
* The specific content (unified diff)

cf. `$ git diff --patch-with-stat   .gitignore README.md` to get them.

Proceed with the operation with confidence.

### Notes

* This proposal focuses on **confirmation before revert**
* Undo-able revert (#1549) is a related but separate topic and can be discussed independently
* Showing the full scope of changes inside the dialog makes the destructive operation easier to trust

raidos23:
# Pull Request: Safer Revert: Clearly Confirm Changes Before Proceeding

Closes #1548

## Overview
Revert is a destructive operation. Previously, git-cola displayed a generic warning message box when reverting files, which did not show:
1. The list of files to be reverted.
2. The diff statistic summary.
3. The actual content changes (unified diff).

This Pull Request implements a detailed confirmation dialog `RevertConfirmDialog` that allows users to review the complete list of files and code changes to be reverted, ensuring safer operations and increasing confidence.

---

### 1. New Revert Confirmation Dialog
- Created `RevertConfirmDialog` in `cola/widgets/revert.py`:
  - Displays a warning message about the destructive nature of the operation.
  - Displays a summary block featuring files, insertions, and deletions counts.
  - Lists the target files in a dedicated read-only pane.
  - Embeds a `DiffTextEdit` showing the full unified diff (`git diff --patch-with-stat`) of the changes to be reverted.
- Implemented `summarize_changes()` to compute insertions/deletions stats from the git diff output.

### 2. Command Integration & Bug Fixes
- Integrated the new dialog into `RevertUnstagedEdits` and `RevertUncommittedEdits` in `cola/cmds.py`.
- Fixed a selection bug: instead of hardcoding selection lists (which caused incorrect file listings under mixed staged/unstaged focus states), the dialog now dynamically retrieves target files from the output of `checkout_args()`. This guarantees the dialog always displays exactly what will be reverted.

### 3. Unit Tests
- Added unit test cases to `test/cmds_test.py`:
  - `test_revert_dialog_summary_contains_file_and_line_counts` to verify parsing of insertions/deletions.
  - `test_revert_unstaged_edits_confirm` to verify correct target files are extracted when staged files are selected.
  - `test_revert_unstaged_edits_confirm_unstaged_only` to verify correct target files are extracted when only unstaged files are selected.

---

## Visual Preview
| New Behavior (Detailed Revert Confirmation) |
|---|
| Capture d’écran du 2026-07-15 11-01-29
|

---

## Verification
Run tests to verify changes:
```bash
pytest test/cmds_test.py
```
Status: **All tests passed**


cola/cmds.py
```diff
@@ -2616,16 +2629,19 @@ def checkout_from_head(self) -> bool:
...
-        ok_text = N_('Revert Uncommitted Changes')
-        return Interaction.confirm(
-            title, text, info, ok_text, default=True, icon=self.icon
-        )
+        checkout_args = self.checkout_args()
+        files = [arg for arg in checkout_args if arg != '--' and arg != self.model.head]
+        diff = self.get_diff_output()
+        dialog = RevertConfirmDialog(self.context, title, text, diff, files)
+
+        return dialog.exec() == RevertConfirmDialog.Accepted
```
davvid, line 2644: I ended up not merging this in favor of a different approach. Please see recent commits related to #1621.

One major issue with the PR is that commands should not be doing anything widget-related, at least not directly.

There's a bit of indirection that was added to make things work differently, with a bit more of reuse of existing dialogs.

I don't doubt that we could make things even nicer looking with a different approach, but it would need a bit of a different design than what's in here.

If we wanted to go hog wild, we'd let the callers of the cmds to provide a confirmation function so that a custom dialog can be fully swapped in by the invokers of the commands in question.

I'm closing this since we already have more or less the functionality discussed in #1548 via #1621 and subsequent commits, but that doesn't mean I'm not open to further improvements in this area.

You'll just have to update to the latest, rebase and redesign things a bit before we can consider merging.

PR #1621, opened 2026-07-15
cmds: show a preview diff when confirming a revert

### LINKED ISSUE #1548: Feature request: Safer revert: clearly confirm changes before proceeding

Land784:
Fixes #1548

The revert confirmation dialogs ask for confirmation without showing which changes are about to be discarded, which makes it easy to destroy uncommitted work by mistake.

Attach a diff with a diffstat for the exact paths that will be checked out to the confirmation dialog. The preview is diffed against the index for unstaged edits and against HEAD for uncommitted edits, matching what `git checkout` is about to discard. The path selection is shared between the preview and the checkout so the two cannot diverge.

Per the discussion on #1548, the diff is collapsed by default behind a "Show Details..." toggle to keep the dialog compact, and scrolls rather than truncates for large diffs. `MessageBox` gains an `expand_details` option so that `critical()` error dialogs keep showing their details expanded, unchanged. In text mode the diff is printed before the confirmation prompt.

Testing: added `test_diff_patch_with_stat` covering the new `gitcmds` helper (empty-path guard, index vs HEAD diff base, path filtering); the full test suite passes; verified by hand in the GUI (both revert commands, toggle expand/collapse, Y/N hotkeys, cancel and confirm paths).

Dialog Collapsed:
dialog_collapsed

Dialog Expanded:
dialog_expanded

### DIFF
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -2556,20 +2556,31 @@ class RevertEditsCommand(ConfirmAction):
     def checkout_from_head(self) -> bool:
         return False
 
+    def path_list(self) -> list[str]:
+        """Return the paths that will be checked out"""
+        s = self.selection.selection()
+        if s.staged:
+            return s.staged
+        else:
+            return s.modified
+
     def checkout_args(self) -> list[str]:
         args = []
-        s = self.selection.selection()
         if self.checkout_from_head():
             args.append(self.model.head)
         args.append('--')
 
-        if s.staged:
-            items = s.staged
-        else:
-            items = s.modified
+        items = self.path_list()
         args.extend(items)
 
         return args
+
+    def preview_diff(self) -> str:
+        """Return a diff of the changes that will be reverted"""
+        paths = self.path_list()
+        return gitcmds.diff_patch_with_stat(
+            self.context, paths, self.checkout_from_head()
+        )
 
     def action(self) -> tuple[int, str, str]:
         checkout_args = self.checkout_args()
--- a/cola/cmds.py
+++ b/cola/cmds.py
@@ -2612,8 +2612,9 @@ class RevertUnstagedEdits(RevertEditsCommand):
         )
         info = N_('Revert the unstaged changes?')
         ok_text = N_('Revert Unstaged Changes')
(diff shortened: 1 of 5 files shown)


davvid (approved):
Nicely done. Will merge shortly. I think there might be another PR related to this topic but this one seems much simpler and nicer. I'll have to close those in favor of this one. Thanks!

davvid:
I made a few tweaks after the fact ~ please take a look. Notably, it uses the familiar diff syntax highlighting just like the main diff editor now.

PR #1622, opened 2026-07-16
Improve MacOS settings display

joshuataylor:
The settings menu is off on MacOS, this fixes it just for MacOS, to not break rendering on other OSs - worth testing Linux KDE/Gnome/xfce/etc.

Before:

image

After:

image

The checkboxes were also slightly off, so I had to tweak the widget. I don't know how to check the type without using isinstance, so if there is an idiomatic way to do this with Qt, please let me know. :-)

### DIFF
--- a/cola/themes.py
+++ b/cola/themes.py
@@ -593,10 +593,16 @@ def style_sheet_default(palette: QPalette, bold_fonts: bool) -> str:
     else:
         font_weight = ''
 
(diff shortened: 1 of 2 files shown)


davvid:
Good idea, thanks for fixing this. Yeah, I've always felt a bit iffy about the checkboxes to be honest.

PR #1624, opened 2026-07-23
test: is garden working?

Ahrirg:
Empty change to verify GitHub/Garden testing workflow is working correctly

### DIFF
--- a/cola/__main__.py
+++ b/cola/__main__.py
@@ -1,13 +1,13 @@
 """Run cola as a Python module.
 
-Usage: python -m cola
+Usage: python -m cola 
 """
 
 from cola import main
 
 
 def run() -> None:
-    """Start the command-line interface."""
+    """Start the command-line interface. test comment for garden checks"""
     main.main()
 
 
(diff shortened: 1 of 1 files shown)


p12tic:
@davvid Is the formatting change expected and wanted or should we find what caused CI to fail?

davvid:
Yes, it's working as expected. Look closely at the original commit -- it adds trailing whitespace and that's why the bot is adding a commit that trims it.

I should probably disable the pre-commit stuff and instead make it so that it causes a check error in the PR.

FTR I don't like to see any commits from the bot, so it's kind of a misleading setup. For me, it's a negative signal when they're present. The intention is that the submitter can squash those changes into their original commit after fetching the changes locally.

p12tic:
@davvid The problem is that CI wants to reformat import style in whole repository.

davvid:
That was fixed today, actually. It looks like ruff 0.16 started flagging a bunch of code.

The following commits resolved it:
* d0054d531a9e23fb63899e5c7447edc957ceaeb8
* d238320418609a8072065f0569d794fb80d24b68

PR #1628, opened 2026-07-31
Dev to main

hermes-agent-ak:


davvid:
Interesting. Didn't realize you had a full-on fork but please continue upstreaming parts of it that shake out as being generally useful. For example, the inline DAG history was a great addition.

Thanks!

Issue #1087, opened 2020-09-16
New startup window design is not so handy as expected


laerreal:
After #1086 startup window became very obfuscated in some cases.
Can the new design be optional?

![before](https://user-images.githubusercontent.com/14236428/93327781-82c95d80-f823-11ea-91bf-e829d91d8d4d.png)
![after](https://user-images.githubusercontent.com/14236428/93327796-85c44e00-f823-11ea-8b77-27ca953107d2.png)

laerreal:
It were difficult to me to revert changes because only one of `src`s is Git-Cola.

davvid:
Thanks for catching this early before release.

Here's a few things we can do.
1) Add tooltips to disambiguate the folders entries.
2) Use `<parent_dir>/<basename>` rather than just `<basename>` as the folder name.  That would result in entries called `git-cola/src`, `cura/src`, `qemu/src`, `filesync/src`, which I think would probably make the folder view more generally useful without getting too cluttered.
3) Add a toggle switch on the left side to swap between the list and folder views.

What do you think?

laerreal:
Looks like 3 is the best way. Let user to choose preferred design.

davvid:
We were already using tooltips so 1) was already done.

For 2) I added logic so that it uses the longest unique path suffix instead of the basenames that it was previously using.

Lastly, there's now a tab bar at the top that allows choosing between the Folder Icon and List modes.

laerreal:
Works fine! Thanks!

laerreal:
Few probles still remain.
1. Some entries are refer to one path (likely, it's because of Windows understands both `/` and `\`)
2. Some pathes are not self describing (ex.: "assets" on picture; it's inside "progdefender/src")

It would be nice to optinally use full pathes in list tab.

![issue](https://user-images.githubusercontent.com/14236428/93718414-ade5e100-fb84-11ea-8460-cd5d1855f11e.png)

davvid:
Thanks again.  I think I have this working correctly now.  I simplified it further so that the list view always displays absolute paths.  I did have to tweak it so that we always use the `/` path separator, but I think the behavior should be better now.  I haven't really tried this on Windows much so I'll dive in a bit and see if there's any other loose ends.

laerreal:
Now it looks good (Windows). Thanks!

PR #1629, opened 2026-08-02
fix(ui): fix recent repository path elision

### LINKED ISSUE #1627: List of recent projects truncates relevant information from path
First of all, thank you for this fantastic tool!

After some recent update the list of recent projects that is shown on launch, does not show the full paths anymore but truncates the path on the right with "...". This strips the most relevant information.

For example in this screenshot, without hovering the mouse on the entries and waiting for a popup, I have no idea what the two `/repos/...` repositories might be.

Image

It would be great if git-cola either showed the full path or did the truncation in the middle of the paths instead of the right end.

raidos23:
Fix the startup recent/bookmarks list so long repository
paths are elided in the middle instead of losing the most useful
part of the path.
Closes #1627

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,0 +1,9 @@
+"""Pytest configuration for Qt-based tests."""
+
+import os
+
+
+# Use an offscreen Qt backend in CI and headless local sessions.
+# This prevents QApplication from aborting when the display server is unavailable
+# or the active platform plugin is unstable.
+os.environ.setdefault('QT_QPA_PLATFORM', 'offscreen')
--- a/cola/models/prefs.py
+++ b/cola/models/prefs.py
@@ -2,6 +2,7 @@ import sys
 from typing import Any
 
 from qtpy import QtCore
+from qtpy.QtCore import Qt
 from qtpy.QtCore import Signal
 
 from .. import core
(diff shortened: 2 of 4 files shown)


kannes:
Thank you!

test/conftest.py
```diff
@@ -0,0 +1,9 @@
+"""Pytest configuration for Qt-based tests."""
+
+import os
+
+
+# Use an offscreen Qt backend in CI and headless local sessions.
+# This prevents QApplication from aborting when the display server is unavailable
+# or the active platform plugin is unstable.
+os.environ.setdefault('QT_QPA_PLATFORM', 'offscreen')
```
davvid, line 9: This is nice and makes our test invocations more consistent. We currently handle this in `garden.yaml` so this would let us drop those settings.

test/startup_test.py
```diff
@@ -87,3 +104,15 @@ def test_get_with_list_mode(app_context):
...
+def test_bookmarks_list_view_uses_middle_elide(app_context, qapp):
+    """The startup bookmarks/recent list should elide long paths in the middle."""
+    view = startup.BookmarksListView(
+        app_context,
+        QtGui.QStandardItemModel(),
+        lambda: None,
+        lambda items: None,
+    )
+
+    assert view.textElideMode() == Qt.ElideMiddle
```
davvid: I question the usefulness of this test (similar to my comment on #1631 

Please take a look at the comments on that PR for the suggested changes.

davvid (changes requested):
As mentioned in #1631 this needs to be a drop-down option with `Left, Middle, Right, None` so that users can choose what to elide.

It's okay if we default to Middle, but we should:
- Allow users to change it
- Remember the setting across invocations

Related-to: #1087

raidos23:
Got it, thanks! I'll work on it.

raidos23:
Implemented configurable text elision mode with Left, Middle, Right, and None options.

davvid:
Thanks!

PR #1631, opened 2026-08-02
startup: elide recent-repo paths in the middle (#1627)

### LINKED ISSUE #1627: List of recent projects truncates relevant information from path

WAHIB-EL-KHADIRI:
## What & why

Closes #1627.

The recent-repositories list on the startup dialog (`BookmarksListView` in `cola/widgets/startup.py`) relied on Qt's default text elide mode, `Qt.ElideRight`. When a path is wider than the item, the **right** end is replaced with `…`, which cuts off the repository name — the most identifying part of the path. As reported, two `/repos/…`-style entries become indistinguishable without hovering for the tooltip.

## Change

Set the view's elide mode to `Qt.ElideMiddle`:

```python
self.setTextElideMode(Qt.ElideMiddle)
```

Now long paths keep both the leading context and the trailing name — `/home/user/…/myproject` instead of `/home/user/very/long/pa…` — which matches the reporter's request and your note that entries should stay unambiguous. `Qt` is already imported in the module, so this is a one-line change.

## Test

Adds `test_recent_list_elides_in_middle`, which constructs a `BookmarksListView` and asserts `textElideMode() == Qt.ElideMiddle`.

Verified locally (PyQt6, offscreen): the constructed view reports `TextElideMode.ElideMiddle`.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cola/widgets/startup.py
+++ b/cola/widgets/startup.py
@@ -338,6 +338,9 @@ class BookmarksListView(QtWidgets.QListView):
         self.setIconSize(make_size(defs.medium_icon))
         self.setDragEnabled(False)
         self.setWordWrap(True)
+        # Elide long paths in the middle so the trailing repository name stays
+        # visible instead of being cut off by the default right-side elision.
+        self.setTextElideMode(Qt.ElideMiddle)
 
         # Context Menu
         self.open_action = qtutils.add_action(
--- a/test/startup_test.py
+++ b/test/startup_test.py
@@ -1,10 +1,28 @@
 """Test Startup Dialog (git cola --prompt) Context Menu and related classes"""
+from qtpy import QtGui
+from qtpy.QtCore import Qt
+
 from cola.widgets import startup
 
 from .helper import app_context
 
 # Prevent unused imports lint errors.
 assert app_context is not None
+
+
+def test_recent_list_elides_in_middle(app_context):
+    """Long repository paths elide in the middle, keeping the trailing name.
+
+    The default right-side elision cut off the repository name, which is the
+    most useful part of a recent-repository path.
+    """
+    view = startup.BookmarksListView(
+        app_context,
+        QtGui.QStandardItemModel(),
+        lambda *args: None,
+        lambda *args: None,
+    )
+    assert view.textElideMode() == Qt.ElideMiddle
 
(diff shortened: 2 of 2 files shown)


davvid:
Sorry claude. Also, this is going to cause a regression for others. Look at the history of changes in this area and you'll see that we had a similar issue in the past, but it was complaining that the elision was happening in the middle.

If we're going to change this then it likely needs to be a {Left, Middle, Right, None} combo-box option so that we can choose what and whether to elide.

cola/widgets/startup.py
```diff
@@ -338,6 +338,9 @@ def __init__(self, context, model, open_selected_repo, set_model, parent=None):
         self.setIconSize(make_size(defs.medium_icon))
         self.setDragEnabled(False)
         self.setWordWrap(True)
+        # Elide long paths in the middle so the trailing repository name stays
+        # visible instead of being cut off by the default right-side elision.
+        self.setTextElideMode(Qt.ElideMiddle)
```
davvid, line 343: CLAUDE: These comments are useless.

test/startup_test.py
```diff
@@ -7,6 +10,21 @@
...
+    """
+    view = startup.BookmarksListView(
+        app_context,
+        QtGui.QStandardItemModel(),
+        lambda *args: None,
+        lambda *args: None,
+    )
+    assert view.textElideMode() == Qt.ElideMiddle
+
+
```
davvid, line 27: CLAUDE: We don't need a test for this. This is borderline tautological.

WAHIB-EL-KHADIRI:
Understood, thanks for catching the regression @davvid — you're right that forcing ElideMiddle changes the behavior others rely on. Appreciate the review, and no worries about closing it.

PR #1635, opened 2026-08-03
Prefer `cola.*` config keys for editor, difftool, mergetool, and history browser (#1632)

### LINKED ISSUE #1632: git-cola specific editor, historybrowser etc
Currently, git-cola reads and writes some tools in other git config namespaces than `cola`, at least: `gui.editor`, `gui.historybrowser`, `diff.tool` and `merge.tool`.

A problem with this is that they affect other tools than git-cola, which may be undesirable. And it can also break them in some cases.

An example of a breaking scenario is when running git-cola from the flatpak: with that one needs the workaround from https://github.com/flathub/com.github.git_cola.git-cola/issues/49#issuecomment-2143381776. But `flatpak-spawn` might not be in `$PATH` so one needs to know to configure a full path to it in git-cola's configs in order to not break other, non-flatpak tools making use of the same config keys. On Ubuntu 26.04 that is `/usr/libexec/flatpak-xdg-utils/flatpak-spawn` (comes in the `flatpak-xdg-utils` package).

While the above is surely a corner case, having git-cola use its specific keys for all these tools would seem nicer. It could fall back to the general ones in case a `cola` namespace one is not set.

raidos23:
close #1632 
## Summary
- Prefer `cola.*` config keys for Git Cola preferences.
- Fall back to the legacy Git config keys for backward compatibility.

## Changes
- Updated preference accessors in `cola/models/prefs.py`.
- Updated the Preferences UI in `cola/widgets/prefs.py`.
- Added unit tests covering priority, fallback behavior, and UI mapping.

## Rationale
- Keep Git Cola preferences separate from Git's global tool configuration while remaining fully backward compatible.

## Testing
- Unit tests pass locally.

## Note
- Existing legacy keys continue to work; users can opt in by configuring the new `cola.*` keys.

### DIFF
--- a/cola/widgets/prefs.py
+++ b/cola/widgets/prefs.py
@@ -105,9 +105,18 @@ class FormWidget(QtWidgets.QWidget):
             getter = self.cfg.get
 
         for config, widget in self.widget_to_config.items():
-            value = getter(config)
-            if value is None:
-                value = self.defaults[config]
+            if config == prefs.COLA_EDITOR:
+                value = prefs.editor(self.context)
+            elif config == prefs.COLA_HISTORY_BROWSER:
+                value = prefs.history_browser(self.context)
+            elif config == prefs.COLA_DIFFTOOL:
+                value = prefs.difftool(self.context)
+            elif config == prefs.COLA_MERGETOOL:
+                value = prefs.mergetool(self.context)
+            else:
+                value = getter(config)
+                if value is None:
+                    value = self.defaults[config]
             set_widget_value(widget, value)
 
 
@@ -405,13 +414,13 @@ class SettingsFormWidget(FormWidget):
                 Defaults.fixup_commit_count,
             ),
             prefs.SORT_BOOKMARKS: (self.sort_bookmarks, Defaults.sort_bookmarks),
-            prefs.DIFFTOOL: (self.difftool, Defaults.difftool),
-            prefs.EDITOR: (self.editor, fallback_editor()),
+            prefs.COLA_DIFFTOOL: (self.difftool, Defaults.difftool),
+            prefs.COLA_EDITOR: (self.editor, fallback_editor()),
             prefs.BACKGROUND_EDITOR: (
                 self.background_editor,
                 Defaults.background_editor,
             ),
-            prefs.HISTORY_BROWSER: (
+            prefs.COLA_HISTORY_BROWSER: (
                 self.historybrowser,
                 prefs.default_history_browser(),
             ),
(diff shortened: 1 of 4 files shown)


cola/widgets/prefs.py
```diff
@@ -105,9 +105,18 @@ def update_from_config(self):
...
-            if value is None:
-                value = self.defaults[config]
+            if config == prefs.COLA_EDITOR:
+                value = prefs.editor(self.context)
+            elif config == prefs.COLA_HISTORY_BROWSER:
+                value = prefs.history_browser(self.context)
+            elif config == prefs.COLA_DIFFTOOL:
+                value = prefs.difftool(self.context)
+            elif config == prefs.COLA_MERGETOOL:
+                value = prefs.mergetool(self.context)
```
davvid, line 115: This is not very maintainable. This code used to be very generic, but now it's becoming inflexible. This needs to be generalized.

cola/widgets/prefs.py
```diff
@@ -425,7 +434,7 @@ def __init__(self, context, model, parent):
                 self.keep_merge_backups,
                 Defaults.merge_keep_backup,
             ),
-            prefs.MERGETOOL: (self.mergetool, Defaults.mergetool),
+            prefs.COLA_MERGETOOL: (self.mergetool, Defaults.mergetool),
```
davvid, line 437: This approach has ramifications that makes it unacceptable. Specifically, many of these configuration values are *intended* to affect the core git configuration values.

In other words, I fully expect to be able to configure a mergetool or difftool and have the command-line `git difftool` follow suite.

This is a niche use case (and I question whether we even need these) so I would not adjust the values set by the preferences.

What I would suggest instead would be to *only* add the override/fallback behavior, and then document it. Users that need that can run a one-off `git config` to set these overrides. This is not the mainstream behavior so I'm not really too keen to accommodate it with such a wide sweeping change.

docs/git-cola.rst
```diff
@@ -1422,22 +1422,35 @@ The default text editor to use is defined in `gui.editor`.
...
+
 merge.tool
 ----------
 
 The default merge tool to use.
 
+Git Cola also honors the `cola.mergetool` key. When set, it takes precedence
+over `merge.tool` for Git Cola-specific mergetool settings.
+
 user.email
```
davvid, line 1454: Do we really need all of these? The difftool and mergetool ones won't even work as expected (that's not how the code works -- we just run `git difftool` and it's core Git that reads the configs).

I would suggest only implementing the override behavior for just the editor, given that we actually invoke it directly.

That said, we already have `cola.backgroundeditor` and it has higher precedence than `gui.editor`, so I'm actually leaning towards thinking that we really don't need this feature.

Alternatively, users could unset these values and then set `EDITOR` in their environment when launching cola and we'll honor it.

davvid (changes requested):

davvid:
Closing this for now because I don't really think we need it. I'll update the issue to mention the alternatives for now. If there's a cleaner implementation and design we can consider, but we need to keep it simple.

davvid:
After more discussion (see #1632) we are still going to pursue this idea, but we'll want to have them separated out so that the Preferences dialog has them as separate fields that default to empty strings.

That way they'll be truly overrides, and the user will have the ability to configure both the core variables and the override variables, separately.

PR #1636, opened 2026-08-07
commitmsg: derive the Commit button shortcut hint from the action

joshuataylor:
The Commit button's tooltip hardcodes `Shortcut: Ctrl+Enter`. Qt maps Ctrl to Command on macOS, so the hint is wrong there -- and because the whole string was translated as one unit, the wrong shortcut was included with every locale too.

Instead, what we can do is render the shortcut from the commit action itself with [`QKeySequence.NativeText`](https://doc.qt.io/qt-6/qkeysequence.html#SequenceFormat-enum), which shows the Command "glyph" (I think that's the term?) on macOS and `Ctrl+Return` elsewhere.

If the shortcut ever changes, the tooltip now follows it for free instead of drifting.

New tests in `test/commitmsg_test.py` assert the tooltip contains the action's actual native-text shortcut rather than a hardcoded string.

---

Before:
image

After:
image

### DIFF
--- a/cola/widgets/commitmsg.py
+++ b/cola/widgets/commitmsg.py
@@ -109,7 +109,16 @@ class CommitMessageEditor(QtWidgets.QFrame):
         )
         self.description.menu_actions.extend(menu_actions)
 
-        commit_button_tooltip = N_('Commit staged changes\nShortcut: Ctrl+Enter')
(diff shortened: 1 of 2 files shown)


davvid:
Thanks!

PR #1638, opened 2026-08-07
dag: speed up commit list and graph label rendering

joshuataylor:
Two rendering costs in the DAG that grow with history size -- both noticeable when loading large histories (I've been doing a lot of git history spelunking lately, so the DAG gets a workout here):

1. `CommitTreeWidget` didn't set uniform row heights, so Qt called the `GraphDelegate`'s `sizeHint()` -- which does per-row font-metrics work -- for EVERY commit just to compute the view's scroll height. Every row is a single line of text in one font, so [`setUniformRowHeights(True)`](https://doc.qt.io/qt-6/qtreeview.html#uniformRowHeights-prop) lets Qt size the view from one row instead of all of them.

2. `Label.boundingRect()` (the ref/tag labels in the graph view) rebuilt a `QPainterPath` and ran `addText()` per tag on every call -- and Qt queries `boundingRect()` frequently (BSP indexing, hit testing, paint culling). The rect only depends on the immutable tags and the fixed label font, so it's now computed once and cached on the item, guarded by font identity.

The cached rect equals the previously computed geometry, so no `prepareGeometryChange()` is needed -- the item's geometry never actually changes, we just stop recomputing it.

No visual change, purely less work per frame / per load.

### DIFF
--- a/cola/widgets/dag.py
+++ b/cola/widgets/dag.py
@@ -1147,6 +1147,11 @@ class CommitTreeWidget(standard.TreeWidget, ViewerMixin):
         ViewerMixin.__init__(self)
 
         self.setSelectionMode(QtWidgets.QAbstractItemView.ExtendedSelection)
+        # Every row is a single line of text in the same font, so all rows are
+        # the same height. Telling Qt this lets it size the view from one row
+        # instead of calling the delegate's sizeHint() for every commit, which
+        # is a large saving on big histories.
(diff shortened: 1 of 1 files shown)


davvid:
Thanks! Good catch on the uniform row heights ~ I thought we already used that here (we definitely use it elsewhere). All these little tweaks add up and make a difference!

PR #1639, opened 2026-08-07
Show an empty-state hint in the diff pane

joshuataylor:
On first launch (or whenever nothing is selected) the diff pane is just a blank grey rectangle -- there's no hint about what it is or how to get something into it.

The diff editor was created with an empty hint string, so this sets a placeholder guiding the user to select a file to view its diff.

Deliberately scoped to only the main editable `DiffEditor` -- the DAG and commit-preview diff panes keep their blank hint, since a blank pane there means "no commit selected" and a staging hint would be misleading.

New tests in `test/diff_hint_test.py` assert the main editor gets the hint and the other diff views stay blank.


davvid:
Thanks!

PR #1640, opened 2026-08-07
Align and pad the Diff Options menu rows

joshuataylor:
The two embedded rows in the Diff Options menu (Intra-line diff mode and Maximum diff size) are `QWidgetAction`s built with `no_margin`, so on macOS they sit flush against the menu edges -- their labels start LEFT of the native checkable items, whose text is indented past the check/icon column, and the combo/spin box touch the menu's right edge.

Before:

image

After:

image

The fix gives both rows a leading gutter so their labels line up with the checkable items, a stretch so the combo and spin box right-align consistently, and small content margins so the controls no longer touch the menu edge.

Rather than hardcoding the gutter width, it's read from the active style via [`PM_SmallIconSize`](https://doc.qt.io/qt-6/qstyle.html#PixelMetric-enum) so it tracks the platform -- the macOS style uses different padding and inset values than Fusion -- falling back to the icon-size constant when the metric reports 0.

This is the shared `Options` widget, so both the main window and the Git DAG diff panes pick it up. Worth a quick look on Linux (KDE/GNOME/etc) to confirm it still looks right there -- it should, since the gutter comes from the active style's own metric.

### DIFF
--- a/cola/widgets/diff.py
+++ b/cola/widgets/diff.py
@@ -1078,6 +1078,15 @@ class Options(QtWidgets.QWidget):
         # Create widgets
         self.widget = parent
         self.filename = filename  # The filename plain text display.
+        # Width of the leading gutter for the embedded menu rows below, sized so
+        # their labels line up with the native checkable items' text. Read from
+        # the active style rather than hardcoded so it tracks the platform --
+        # macOS uses different padding/inset values than Fusion. Falls back to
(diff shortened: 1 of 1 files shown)


davvid:
Yeah, the margins in there have always bugged me but I never had a chance to dive in and fix them. Very nicely done.

PR #1641, opened 2026-08-07
status: keep only the next file selected after staging multiple

joshuataylor:
Staging several untracked files at once leaves the wrong rows highlighted in the Status widget.

This was confusing to figure out, and explain..

Say `Untracked` contains `[a, b, c, d, e]` and you select `a`, `b`, `c` and stage them -- you'd expect the selection to land on `d` (the next file), but both `d` AND `e` end up selected.

before (I'm using the keyboard and staging files, and it's really glitchy):

https://github.com/user-attachments/assets/62436ac9-9c5a-48db-9313-e1f24d61c81c

after:

https://github.com/user-attachments/assets/e2c17243-85b0-4015-93d7-f1eb1802fd8e

This one took a while to root-cause, and the mechanism is a fun Qt ordering quirk:

`StatusTreeWidget._save_selection()` runs from the `about_to_update` signal, which `MainModel` delivers with `Qt.QueuedConnection` -- so it actually fires AFTER the model has already swapped in the post-stage file lists. It then rebuilt `old_selection` with `tree_selection()`, mapping the still-selected old tree rows onto the NEW path list: selected rows 0,1 (`a`, `b`) resolved to the new `d`, `e`, and `_restore_selection()` happily reselected every "old" path still present.

The fix, I think, snapshot `old_contents`/`old_selection` from the `previous_contents` signal instead, which the model emits just before `about_to_update` carrying the pre-update lists -- the same source `_restore_selection()` already trusts for its cross-section transplanting. The live-model path is kept as a fallback for the first refresh, when `previous_contents` is `None`.

### DIFF
--- a/cola/widgets/status.py
+++ b/cola/widgets/status.py
@@ -602,8 +602,25 @@ class StatusTreeWidget(QtWidgets.QTreeWidget):
         return entry
 
     def _save_selection(self):
-        self.old_contents = self.contents()
-        self.old_selection = self.selection()
+        # This runs from the queued about_to_update signal, which is delivered
+        # after the model has already swapped in the new file lists. Reading the
+        # live model here (self.contents()/self.selection()) would map the tree's
+        # still-old selected rows against the new path lists and record the wrong
+        # paths -- e.g. staging a,b,c would report the following d,e as selected.
+        # previous_contents is emitted just before about_to_update and holds the
+        # pre-update lists, so use it to snapshot the old contents and selection.
+        old = self.previous_contents
+        if old is None:
+            self.old_contents = self.contents()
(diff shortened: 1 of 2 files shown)


davvid:
Does it do the right thing for modified vs. staged files as well? (haven't read the code yet)

Good stuff, looking forward to merging something like this after the remote server topic.

davvid:
My other thought was.. does removing the QueuedConnection on any subset of the current signals "fix" the original implementation by reordering the delivery?

davvid:
I tested this out a bit and it works pretty nicely, so I've gone ahead and merged. As always, I'm always down for further improvements if we find 'em.

PR #1642, opened 2026-08-08
Implement `cola.*` configuration overrides

### LINKED ISSUE #1632: git-cola specific editor, historybrowser etc

raidos23:
Implement `cola.*` overrides for the editor, history browser, diff tool, and merge tool settings.

### Changes

- Add separate `cola.*` override settings for:
  - `cola.editor`
  - `cola.historybrowser`
  - `cola.difftool`
  - `cola.mergetool`
- Keep the existing `gui.*` / Git configuration keys as fallbacks.
- Make `cola.*` values take precedence when configured.
- Add a shared `_config_with_override()` helper to keep the fallback logic centralized.
- Pass the configured diff and merge tools explicitly when launching them.
- Keep the normal settings and their overrides as separate UI fields.

### Verification

Tested manually in a dedicated Git repository:

1. Removed both `cola.difftool` and `diff.tool`.
2. Configured only `diff.tool` to `vimdiff`.
3. Verified that `diff.tool` was used when `cola.difftool` was unset.
4. Configured `cola.difftool` with a different tool.
5. Verified that `cola.difftool` took precedence.
6. Removed `cola.difftool` and verified that the configured `diff.tool` was used again.

All tested cases passed.

Closes #1632

### DIFF
--- a/cola/difftool.py
+++ b/cola/difftool.py
@@ -12,6 +12,7 @@ from . import git
 from . import gitcmds
 from . import hotkeys
 from . import icons
+from .models import prefs
 from . import qtutils
 from . import utils
 from .i18n import N_
@@ -48,7 +49,11 @@ class LaunchDifftool(cmds.ContextCommand):
                 shellquote_terms = {'xfce4-terminal'}
                 shellquote_default = terminal in shellquote_terms
 
-                mergetool = ['git', 'mergetool', '--no-prompt', '--']
+                mergetool = ['git', 'mergetool', '--no-prompt']
+                tool = prefs.mergetool(self.context)
+                if tool:
+                    mergetool.append(f'--tool={tool}')
+                mergetool.append('--')
                 mergetool.extend(paths)
                 needs_shellquote = cfg.get(
                     'cola.terminalshellquote', shellquote_default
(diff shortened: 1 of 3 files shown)


davvid:
Thanks! Merging.. the only  missing piece was the documention in `git-cola.rst` for these new variables. That was fixed up in 42559f2d2ce6d36e8f335e414df259a9b6b99e19 with some minor tweaks on top.

PR #1643, opened 2026-08-08
Improve SVG image rendering in the diff viewer

joshuataylor:
I was staging an SVG and noticed the SVGs came up blurry in the image diff pane, while the PNG sitting right next to them looked fine, and the same SVG looks perfectly fine opened anywhere else.

It bugged me enough to go digging, and it turns out to be a rasterisation-size quirk rather than anything wrong with the files. I learnt way more about both qt and SVGs!

Before:

before-svg

After:

after-svg

Note, PNGs DO work:

png-example

## What's happening

The image diff viewer loads every file with `QtGui.QPixmap(path)` (`Viewer.set_images` in `cola/widgets/diff.py`). That reads a raster at its real pixels, so a 512x512 PNG barely rescales to fit the pane and stays sharp. An SVG, though, gets rasterised through Qt's `qsvg` plugin at its intrinsic size: a logo whose only size hint is `viewBox="0 0 118 118"` becomes a 118x118 bitmap, which `QGraphicsView` then upscales ~5x to fill the pane. Nothing in the image path used `QSvgRenderer` or the device pixel ratio, so the vector never got a chance to render at display size -- which is exactly what makes it look fine everywhere else.

## The fix

The idea is to make `ImageView` re-rasterise from a vector source instead of a fixed bitmap. 

The scene rect stays in the image's logical units (118), but the SVG is rendered to a `QImage` of `logical_size * render_scale` device pixels, wrapped in a pixmap with `setDevicePixelRatio(render_scale)`. The `QGraphicsPixmapItem` then occupies the logical size while painting at higher resolution, so all the existing `fitInView`/zoom maths stays untouched. `render_scale` is `abs(transform().m11()) * viewport().devicePixelRatioF()`, recomputed at the tail of `fitInView` (which resize, wheel-zoom, pan and reset all route through) and re-rasterised past a small hysteresis band, capped at 4096px so wheel-zoom doesn't thrash or balloon memory. It falls out nicely for rasters too: they have a device pixel ratio of 1, so `size() / dpr == size()` and their path is unchanged.

IMHO the fixed 4096px cap is fine, but I'm happy to turn it into a `cola.*` setting if you'd rather -- no strong feelings either way.

## What I tweaked

- `cola/widgets/imageview.py` gains `looks_like_svg()` (a cheap extension/content sniff that gates `QSvgRenderer` so raster images do not trigger a Qt parse warning), `render_svg_to_pixmap()`, `ImageView.set_image_source()`, `update_render_resolution()`, and a shared `_composite_checkerboard()` helper. `update_scene_rect`/`image_scene_rect` now size the scene in logical units (`width() / devicePixelRatio()`). The standalone image viewer's `ImageView.load()` is SVG-aware too.
- `cola/widgets/diff.py` gains `load_image_source()` (an SVG renderer or a `QPixmap`), `source_logical_size()`, `comp_logical_size()`, `paint_source()` and a scale-aware `render_comp_image()` that composites the side-by-side / diff / xor modes at a given scale and returns a dpr-tagged pixmap. `Viewer.render()` installs a re-rasterising callback via `set_image_source` when any source is an SVG, and keeps the existing raster path unchanged otherwise.

Happy for this to be split into smaller pieces, or reworked entirely -- take whatever's useful (none, some, all, I really don't mind :-)).

## References

Here are some doc references I used (I have these docs cloned locally to ripgrep through, makes it SOO much easier when working with qt!!).

- [`QPixmap`](https://doc.qt.io/qt-6/qpixmap.html) -- the file-path constructor rasterises at the format's intrinsic size and is the source of the blur; the fix instead tags the rendered pixmap with [`setDevicePixelRatio()`](https://doc.qt.io/qt-6/qpixmap.html#setDevicePixelRatio).
- [`QSvgRenderer`](https://doc.qt.io/qt-6/qsvgrenderer.html) -- vector rendering: [`isValid()`](https://doc.qt.io/qt-6/qsvgrenderer.html#isValid), [`defaultSize`](https://doc.qt.io/qt-6/qsvgrenderer.html#defaultSize), [`viewBox`](https://doc.qt.io/qt-6/qsvgrenderer.html#viewBox-prop) and [`render()`](https://doc.qt.io/qt-6/qsvgrenderer.html#render). PyQt6 binding: [riverbankcomputing.com/.../qsvgrenderer.html](https://www.riverbankcomputing.com/static/Docs/PyQt6/api/qtsvg/qsvgrenderer.html).
- [High DPI Displays](https://doc.qt.io/qt-6/highdpi.html) and [`QPaintDevice::devicePixelRatioF()`](https://doc.qt.io/qt-6/qpaintdevice.html#devicePixelRatioF) -- the device-independent vs device pixel model the render scale is derived from.
- [`QGraphicsView::fitInView()`](https://doc.qt.io/qt-6/qgraphicsview.html#fitInView) and [`QTransform::m11()`](https://doc.qt.io/qt-6/qtransform.html#m11) -- how the view scales the scene, and the horizontal scale factor the re-rasterise resolution reads from.
- [`QGraphicsPixmapItem::setTransformationMode()`](https://doc.qt.io/qt-6/qgraphicspixmapitem.html#setTransformationMode) -- the item defaults to `Qt::FastTransformation` (nearest-neighbour), so an under-sized raster is scaled up without smoothing; rendering the vector at display resolution avoids relying on it.
- [`QPainter::RenderHint`](https://doc.qt.io/qt-6/qpainter.html#RenderHint-enum) -- `Antialiasing` and `SmoothPixmapTransform`, set on the compositing painter in `render_comp_image()`.

### DIFF
--- a/cola/widgets/diff.py
+++ b/cola/widgets/diff.py
@@ -6,6 +6,7 @@ from typing import Optional
 
 from qtpy import QtCore
 from qtpy import QtGui
+from qtpy import QtSvg
 from qtpy import QtWidgets
 from qtpy.QtCore import Qt
 from qtpy.QtCore import Signal
@@ -714,6 +715,7 @@ class Viewer(QtWidgets.QFrame):
         self.model = model = context.model
         self.images = []
         self.pixmaps = []
+        self.sources = []
         italic_font = self.font()
         italic_font.setItalic(True)
 
@@ -954,26 +956,39 @@ class Viewer(QtWidgets.QFrame):
     def set_images(self, images):
         self.images = images
         self.pixmaps = []
+        self.sources = []
         if not images:
             self.reset()
             return False
 
-        # In order to comp, we first have to load all the images
-        all_pixmaps = [QtGui.QPixmap(image[0]) for image in images]
-        pixmaps = [pixmap for pixmap in all_pixmaps if not pixmap.isNull()]
-        if not pixmaps:
+        # In order to comp, we first have to load all the images. SVGs are kept
+        # as renderers so they can be re-rasterised crisply at display size.
+        sources = [load_image_source(image[0]) for image in images]
+        sources = [source for source in sources if source is not None]
+        if not sources:
             self.reset()
             return False
 
-        self.pixmaps = pixmaps
+        self.sources = sources
+        self.pixmaps = [
+            source for source in sources if isinstance(source, QtGui.QPixmap)
+        ]
         self.render()
         self.cleanup()
         return True
 
     def render(self):
         # Update images
-        if self.pixmaps:
-            mode = self.options.image_mode.currentIndex()
+        mode = self.options.image_mode.currentIndex()
+        if self.sources and any(is_svg_source(source) for source in self.sources):
+            # Vector sources: hand the view a re-rasterising callback so the
+            # image stays sharp at any zoom level or device-pixel-ratio.
+            sources = self.sources
+            logical = comp_logical_size(sources, mode)
+            self.image.set_image_source(
+                lambda scale: render_comp_image(sources, mode, scale), logical
+            )
+        elif self.pixmaps:
             if mode == self.options.SIDE_BY_SIDE:
                 image = self.render_side_by_side()
             elif mode == self.options.DIFF:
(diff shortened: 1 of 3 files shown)


davvid:
Nice! It's much clearer now, really cool. Thanks

PR #1645, opened 2026-08-18
Enable filesystem monitoring and archiving in server mode

Ahrirg:
This PR fixes and enables the previously disabled filesystem monitoring and `Save as Tarball/Zip...` support when running in server mode.


cola/operations.py
```diff
@@ -315,9 +348,89 @@ def expanduser(self, path: str) -> str:
...
+        else:
+            return core.run_command(cmd, *args, **kwargs)
+
+    def start_monitor(self, worktree: str | None, git_dir: str) -> None:
+        if self._monitor is not None:
+            return
+
+        from . import fsmonitor
+        from . import git
+        from . import gitcfg
```
davvid: These inline imports are something we should really try to avoid.

Up until now we've been very diligent about keeping dependencies straight and not having circular import loops in the code. This is introducing them.

Please take a look at `cola/interaction.py` and the late binding we do in `def install()` in `cola/widgets/standard.py` to rebind those low-level core methods to GUI-dependent widgets.

We're using dependency injection to keep things modular and free of these import problems. Note that `cola/app.py` is what binds them together and injects stuff by calling `standard.install()`.

Please take a pass at following this approach with the operations module. It's a very critical low-level component so I understand why reaching for other parts of the code is necessary at this level. The functionality is good ~ it just needs a little bit of indirection to make it fit together nicely.

The "operations" module is effectively that level of indirection already for clients of it, so this is more an architectural pattern that we need to follow.


You can already see the code problems in the existing `cola/operations.py` -- it is using an inline import of the server module to avoid an import loop. Note how `cola/server.py` already imports `operations`, so if operations did not do the deferred / inline import then there would be a circular import dependency.

Please follow a similar "inject dependencies using install()" pattern for the server parts and for these new dependencies that are being introduced.

Avoiding inline imports is a hard rule we need to follow.

cola/operations.py
```diff
@@ -315,9 +348,89 @@ def expanduser(self, path: str) -> str:
...
     def run_command(
-        self, cmd: list[core.UStr | str], *args, **kwargs
+        self,
+        cmd: list[core.UStr | str],
+        communicate_data: list[str] | None = None,
+        *args,
+        **kwargs,
     ) -> tuple[int, core.UStr, core.UStr]:
-        return core.run_command(cmd, *args, **kwargs)
+        _stdout = None
```
davvid: I would fix this up myself, but since I left notes for other larger changes (and you're already in here making great progress) I'll leave it as a note since it's better to be on the same page.

I would have just called this variable `stdout` instead of `_stdout`. I don't see any reason why we shouldn't just call it stdout (or another name that doesn't start with `_`) here.

cola/operations.py
```diff
@@ -315,9 +348,89 @@ def expanduser(self, path: str) -> str:
...
+    def _execute_command(
+        self,
+        cmd: list[core.UStr | str],
+        communicate_data: list[str] | None,
+        *args,
+        **kwargs,
+    ):
+        if communicate_data is not None:
+            proc = core.start_command(cmd, *args, **kwargs)
+            out, err = proc.communicate(communicate_data)
```
davvid: I'm slightly confused here -- I didn't know `communicate` accepted a list. My understanding is that it accepts bytes (or str if we specify an encoding).

I would also suggest including the keyword argument name, e.g. `proc.communicate(input=...)` so keep things explicit and obvious.

On first glance it *seems* like passing a list to `communicate(...)` is a mistake. Please double-check.

if it happens to be that communicate *does* accept lists, but only with newer versions of Python, then the short answer is that we can't use that functionality and will need to do things the "old fashioned" way for maximum backwards-compatibility.

davvid (changes requested):
Nicely done overall. Please address the implementation concerns and we'll get this merged. cheers

cola/operations.py
Ahrirg: Yeah indeed, this was a mistake on my end, i fixed it. Thanks for pointing it out!

cola/operations.py
Ahrirg: I implemented it using the injection approach. If I misunderstood what you meant, please let me know.

cola/fsmonitor.py
```diff
@@ -88,7 +90,7 @@ def refresh(self) -> None:
 
 class _BaseThread(QtCore.QThread):
     def __init__(self, context: ApplicationContext, monitor: _Monitor) -> None:
-        QtCore.QThread.__init__(self)
+        QtCore.QObject.__init__(self)
```
davvid: Hmm, I think it'd be best to keep this as `QThread`, even if they're (currently?) interchangeable.

cola/operations.py
```diff
@@ -315,9 +360,85 @@ def expanduser(self, path: str) -> str:
         return core.expanduser(path)
 
     def run_command(
-        self, cmd: list[core.UStr | str], *args, **kwargs
+        self,
+        cmd: list[core.UStr | str],
+        communicate_data: list[str] | None = None,
```
davvid: The `communicate_data` type annotations still mention `list`, so that'll need adjustment. This is very close now.

After that, please rename the function argument to  `input` instead of `communicate_data` so that it mirrors the same arguments as the [core stdlib communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate). One less thing to think about / remember.

cola/operations.py
davvid: Also, make sure to pass the kwarg name at the call site, e.g. `out, err = proc.communicate(input=input)`

cola/operations.py
davvid: Maybe kinda janky, but since we already have `**kwargs` here we can smuggle the `input` param through it instead (of having a dedicated name parameter) so that `run_command` can keep a generic signature that "feels" like a convenient subprocess pass-through that doesn't throw exceptions.

In that case, I'd suggest capturing `input = kwargs.pop('input', None)` early on before calling `start_command(..)` so that `input` will already have been scrubbed from `kwargs` and we can just forward it all along.

This may even let us do without the if/else blocks, since it may allow the logic to unify again.

I just noticed this.. by "unify again".. I really mean that this `input=...` logic around `communicate(..)` should be pushed down into the `core.run_command` function itself. This wrapper shouldn't really be responsible for it, it seems. It'd be better to let `core.run_command` handle a bit more of it so that this wrapper can be made simpler.

cola/operations.py
```diff
@@ -1,22 +1,53 @@
...
 from typing import Any
 
+from qtpy import QtCore
+
 from . import core
 from . import utils
 
 if TYPE_CHECKING:
     from . import server
+    from .fsmonitor import _Monitor
```
davvid: If we need to import this for typechecking then it should not have an underscore name. At that point we should consider it public and rename it accordingly. We never import "private" stuff, even for typehints.

cola/operations.py
```diff
@@ -1,22 +1,53 @@
...
+        pass
+
+    @staticmethod
+    def git_create(ops: IOperations) -> Git:
+        pass
+
+    @staticmethod
+    def gitcfg_create(context) -> GitConfig:
+        pass
+
```
davvid: Thanks for implementing the `install()` functions and injecting the functionality. This is pretty much exactly what I had in mind.

cola/server.py
```diff
@@ -178,6 +181,10 @@ def send_message(self, message: str, seq_number: int):
 
 def run(address, port, verbose) -> None:
     """Start the websocket cola operations server"""
+    fsmonitor.install()
+    git.install()
+    gitcfg.install()
```
davvid: Does the main (non-server) app also need to have these functions installed? Even if it's inert, it wouldn't hurt to have things setup consistently app-wide IMO rather than having these steps run only here.

davvid (changes requested):
Thanks again! I'll take another look again soon once these last details are sorted out.

cola/operations.py
Ahrirg: done

cola/operations.py
Ahrirg: Sorry, I may not have committed the change from list[str] to bytes.  I unified it soo it's fully removed now.

cola/server.py
Ahrirg: I found a way to achieve this without requiring any install, by splitting operations into multiple files.

davvid (approved):

PR #1647, opened 2026-09-08
fix: default to PyQt6 when both PyQt5 and PyQt6 are installed (#1625)

### LINKED ISSUE #1625: Can't select folder (for Clone or Open Git Repository)
Clicking the "Open" button doesn't do anything anymore.

When trying to clone a repository, there's a workaround: by manually entering a non-existing folder name.
But that means that it can't clone in the projects folder root anymore.

It works with `QT_QPA_PLATFORMTHEME=gtk3 git-cola`.

System:
git-cola version: 4.19.0
Qt: 6.11.1
PyQt: 6.11.0
QtPy: 2.4.3
KDE Plasma (Wayland): 6.7.3

Gargeesharmaa:
Summary
Fixes an issue on modern desktop environments (such as KDE Plasma / Qt6 setups) where folder selection dialogs (e.g., for "Open" or "Clone" operations) silently fail when both PyQt5 and PyQt6 are installed on the system (#1625).

Root Cause
QtPy defaults to PyQt5 when both binding packages are present unless the QT_API environment variable is explicitly set. Using legacy PyQt5 bindings in Qt6/KDE Wayland environments leads to broken native file picker dialogs and dark theme rendering issues.

Solution
Added early auto-detection in cola/qtcompat.py (and launcher initializers) to set os.environ["QT_API"] = "pyqt6" if QT_API is not already defined by the user and PyQt6 is available in the environment.

Ensures git-cola automatically leverages PyQt6 out of the box without requiring manual environment exports on updated systems.

Closes #1625

### DIFF
--- a/cola/qtcompat.py
+++ b/cola/qtcompat.py
@@ -1,5 +1,14 @@
 from __future__ import annotations
+import os
 from collections.abc import Callable
+
+# Prefer PyQt6 over PyQt5 when QT_API is not explicitly set by the user
+if "QT_API" not in os.environ:
+    try:
+        from qtpy import PYQT6  # noqa: F401
+        os.environ["QT_API"] = "pyqt6"
+    except ImportError:
+        pass
 
 from qtpy import QtCore
 from qtpy import QtGui
@@ -12,7 +21,6 @@ except ImportError:
     PYQT4 = False
 
 from . import hotkeys
-
 
 def patch(
     obj: type[QtWidgets.QGraphicsItem | QtGui.QKeySequence],
--- a/cola/qtcompat.py
+++ b/cola/qtcompat.py
@@ -3,10 +3,11 @@ import os
 from collections.abc import Callable
 
 # Prefer PyQt6 over PyQt5 when QT_API is not explicitly set by the user
-if "QT_API" not in os.environ:
+if 'QT_API' not in os.environ:
     try:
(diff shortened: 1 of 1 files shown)


cola/qtcompat.py
```diff
@@ -1,6 +1,16 @@
...
 from collections.abc import Callable
 
+# Prefer PyQt6 over PyQt5 when QT_API is not explicitly set by the user
+if 'QT_API' not in os.environ:
+    try:
+        from qtpy import PYQT6  # noqa: F401
+
+        os.environ['QT_API'] = 'pyqt6'
+    except ImportError:
+        pass
```
davvid, line 12: Unfortunately it's not quite that easy. If you `import qtpy` (which happens right above the `os.environ` line here) then the environment variable setup has already been performed. It's done at import-time so setting the environment variable post-import has zero effect.

I really think the best effort would be follow up with https://github.com/spyder-ide/qtpy/issues/519 and get the "prefer PyQt6 over PyQt5" preference fixed upstream in the qtpy project.

Once we've got that then we can sync against the qtpy release where that change is introduced.

Gargeesharmaa:
Thanks for the explanation @davvid! I saw the updates on spyder-ide/qtpy#519. It makes complete sense to wait for QtPy 2.6.0 to switch the default lookup order to PyQt6 upstream. Glad to see the issue moving forward there!

PR #1649, opened 2026-09-12
qt: prefer PyQt6 when QT_API is unset and PyQt6 is installed (#1625)

nevil06:
QtPy defaults to PyQt5 when available. On modern Linux desktop environments such as KDE Plasma 6 (Wayland), running with Qt5 causes native portal folder dialogs (e.g. for 'Open Git Repository' or 'Clone') to silently fail to open or pick folders, as well as severe dark theme contrast issues (#1625).

Use importlib.util.find_spec() during package initialization in cola/__init__.py to detect PyQt6 presence without importing it, and default QT_API to 'pyqt6' if not already explicitly set in os.environ. If QT_API is already specified by the user or script, that setting is preserved.

### DIFF
--- a/cola/__init__.py
+++ b/cola/__init__.py
@@ -1,0 +1,19 @@
+"""Git Cola package initialization."""
+from __future__ import annotations
+import importlib.util
+import os
+
+
+def auto_select_qt_api() -> None:
+    """Default to PyQt6 when QT_API is unset and PyQt6 is installed.
+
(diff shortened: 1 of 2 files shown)


davvid:
Duplicate: #1647 

For now you can just `export QT_API=PyQt6` in your environment. Also, if you have stray pyqt5 packages around you may be able to uninstall them and avoid the issue.

